PR #1, opened 2019-10-18
Initialize Observatorium API

kakkoyun:
cc @brancz @bwplotka @metalmatze


Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+RUN apk add --update --no-cache ca-certificates tzdata git make && update-ca-certificates
+
+ADD . /opt
+WORKDIR /opt
+
+RUN git update-index --refresh; make observatorium
+
+FROM scratch as runner
+
+COPY --from=builder /usr/share/zoneinfo /usr/share/zoneinfo
```
metalmatze: We should be running with UTC, so I don't think we need this.

Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM golang:1.13.1-alpine3.10 as builder
+
+RUN apk add --update --no-cache ca-certificates tzdata git make && update-ca-certificates
+
+ADD . /opt
+WORKDIR /opt
+
+RUN git update-index --refresh; make observatorium
+
+FROM scratch as runner
```
metalmatze: I don't really like running scratch, it's super hard to debug inside the container if something is wrong. Personally I'd prefer alpine, but I can also live with busybox.

internal/prober.go
```diff
@@ -0,0 +1,101 @@
...
+	return func(w http.ResponseWriter, _ *http.Request) {
+		if !probeFunc() {
+			http.Error(w, fmt.Sprintf("observatorium is NOT %v", t), http.StatusServiceUnavailable)
+			return
+		}
+		if _, err := io.WriteString(w, fmt.Sprintf("observatorium is %v", t)); err != nil {
+			level.Error(p.logger).Log("msg", "failed to write probe response", "probe type", t, "err", err)
+		}
+	}
+}
```
metalmatze: Very nice!!!
One small comment, instead of the probeType, what's wrong with readinessHandler and livenessHandler?

main.go
```diff
@@ -0,0 +1,176 @@
...
+)
+
+func main() {
+	profile := os.Getenv("PROFILE") != ""
+
+	debug := os.Getenv("DEBUG") != ""
+	if debug {
+		runtime.SetMutexProfileFraction(10)
+		runtime.SetBlockProfileRate(10)
+	}
```
metalmatze: These could also be flags, no? :thinking:

main.go
```diff
@@ -0,0 +1,176 @@
...
+
+	"github.com/observatorium/observatorium/internal"
+	"github.com/observatorium/observatorium/internal/proxy"
+	"go.uber.org/automaxprocs/maxprocs"
+
+	"github.com/go-kit/kit/log/level"
+	"github.com/oklog/run"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/prometheus/common/route"
```
metalmatze: Personally, I'd prefer go-chi as router here, which is a bit more maintained in terms of features, I believe. Not sure how common it is to use Prometheus' route. To me it seems more like an externally exposed internal package, but I can be wrong here.

internal/prober.go
metalmatze: Thinking about this further while reviewing. I don't think this handler should be concerned with whether it is liveness or readiness. The handler endpoint is determined outside of this and then the real difference should be the funcs passed into this handler.

main.go
kakkoyun: I don't have strong feelings about the router at this stage. We don't use any of the offered features right now, I just thought given the requirements, we would need a more feature-complete router than `http.ServeMux`. That being said, I have chosen Prometheus router because it's a thin wrapper around https://github.com/julienschmidt/httprouter , which is the one I'd prefer to use. Again, I don't have any strong feelings about it.

If you all think, we should go with `go-chi`, I can quickly make the necessary changes.

internal/prober.go
kakkoyun: Could you please elaborate this furthermore? What would you like to see as an API from the perspective of the consumer?

main.go
metalmatze: Makes sense. The problem with httprouter is that it doesn't use the stdlib http handlers. My question mostly was about Prometheus router being used by other projects though. :)

.drone.yml
```diff
@@ -0,0 +1,47 @@
+---
```
bwplotka: I guess drone is what we want to use? Wonder why not GH action if we start from scratch?

internal/prober.go
bwplotka: Short nit: Do you think we can import this from thanos  `pkg/prober`? or is it too much overhead? We could then improve single version of the same probing thing.

bwplotka: Essentially would be nice to have the improvements on Thanos side as well.

main.go
```diff
@@ -0,0 +1,176 @@
...
+		metricsReadEndpoint  string
+		metricsWriteEndpoint string
+	}{}
+
+	flag.StringVar(&opts.listen, "listen", ":8080", "The address on which internal server runs.")
+	flag.StringVar(&opts.gracePeriod, "grace-period", "5s", "The time to wait after an OS interrupt received.")
+	flag.StringVar(&opts.debugName, "debug.name", "observatorium", "The Name to add as prefix to log lines.")
+	flag.StringVar(&opts.logLevel, "log.level", "info", "The log filtering level. Options: 'error', 'warn', 'info', 'debug'.")
+	flag.StringVar(&opts.logFormat, "log.format", internal.LogFormatLogfmt, "The log format to use. Options: 'logfmt', 'json'.")
+	flag.StringVar(&opts.metricsReadEndpoint, "metrics-read-endpoint", "", "The endpoint to which to make write requests for metrics.")
```
bwplotka: This is what API? Query API?

internal/proxy/prometheus.go
```diff
@@ -0,0 +1,34 @@
...
+		r.URL.Host = prometheusURL.Host
+		// Normalize request so Prometheus doesn't know that request was proxied.
+		r.URL.Path = strings.TrimPrefix(r.URL.Path, prefix)
+	}
+
+	return func(w http.ResponseWriter, r *http.Request) {
+		// Put the upstream PrometheusURL into the request's context for the director.
+		r = r.WithContext(context.WithValue(r.Context(), promKey, endpoint))
+
+		proxy := &httputil.ReverseProxy{
```
bwplotka: I think this is anti-patter to allocate new reverse proxy on every request :thinking: At least this is not like I was using it all the time (: See: https://github.com/improbable-eng/kedge/blob/master/pkg/winch/http/proxy.go

internal/proxy/prometheus.go
```diff
@@ -0,0 +1,34 @@
...
+
+		r.URL.Scheme = prometheusURL.Scheme
+		r.URL.Host = prometheusURL.Host
+		// Normalize request so Prometheus doesn't know that request was proxied.
+		r.URL.Path = strings.TrimPrefix(r.URL.Path, prefix)
+	}
+
+	return func(w http.ResponseWriter, r *http.Request) {
+		// Put the upstream PrometheusURL into the request's context for the director.
+		r = r.WithContext(context.WithValue(r.Context(), promKey, endpoint))
```
bwplotka: Why do we  put this into context? Using context for hidden request meta should be avoided unless it's has to be there. This is because it's is easy to forgot to add it or clone it on the way and it is easy to overallocate by accident. What's the use case here?

This endpoint is definitely static, right?

internal/proxy/prometheus.go
```diff
@@ -0,0 +1,34 @@
...
+	"context"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+	"strings"
+)
+
+type promURL string
+
+func NewPrometheus(prefix string, endpoint *url.URL) http.HandlerFunc {
```
bwplotka: TBH I can't see anything Prometheus specific in this code :thinking:

internal/proxy/prometheus.go
```diff
@@ -0,0 +1,34 @@
...
+	var promKey promURL
+
+	director := func(r *http.Request) {
+		prometheusURL := r.Context().Value(promKey).(*url.URL)
```
bwplotka: This is very scary. What if we forget to fill it? See my comment above.

main.go
bwplotka: I think itis quite easy to change overtime - I personally don't have strong opinion other than choosing one (: Something that uses native HTTP handlers would be amazing.

main.go
```diff
@@ -0,0 +1,176 @@
...
+		gracePeriod          string
+		debugName            string
+		logLevel             string
+		logFormat            string
+		metricsReadEndpoint  string
+		metricsWriteEndpoint string
+	}{}
+
+	flag.StringVar(&opts.listen, "listen", ":8080", "The address on which internal server runs.")
+	flag.StringVar(&opts.gracePeriod, "grace-period", "5s", "The time to wait after an OS interrupt received.")
```
bwplotka: Nice idea! I wish we have on Thanos (:

internal/prober.go
kakkoyun: I don't mind, I like the idea actually. Let's iterate on this one and then I can commit the same code to Thanos as well.

main.go
kakkoyun: This will proxy to Query API, yes. I'm open to naming suggestions.

main.go
kakkoyun: Alright, I'll check `go-chi` then, I like the idea to be able to use native HTTP handlers.

main.go
squat: In this case we need to update the help text, which currently says:

> "The endpoint to which to make write requests for metrics.“

internal/proxy/prometheus.go
kakkoyun: This is something I have copied over from one of our old iterations on an equivalent API. I have removed the authentication logic, and then I didn't put much thought on it. I'll clean it up and simplify.

internal/proxy/prometheus.go
kakkoyun: This is something I have copied over from one of our old iterations on an equivalent API.
I'll check the link you provided and improve these bits. 👍

main.go
kakkoyun: Happy to add it there as well.

main.go
```diff
@@ -0,0 +1,176 @@
...
+	"net/url"
+	"os"
+	"os/signal"
+	"runtime"
+	"syscall"
+	"time"
+
+	"github.com/observatorium/observatorium/internal"
+	"github.com/observatorium/observatorium/internal/proxy"
+	"go.uber.org/automaxprocs/maxprocs"
```
squat: Tiny nit: there should be a new linebacker before the Uber dep rather than after

main.go
```diff
@@ -0,0 +1,176 @@
...
+		debugName            string
+		logLevel             string
+		logFormat            string
+		metricsReadEndpoint  string
+		metricsWriteEndpoint string
+	}{}
+
+	flag.StringVar(&opts.listen, "listen", ":8080", "The address on which internal server runs.")
+	flag.StringVar(&opts.gracePeriod, "grace-period", "5s", "The time to wait after an OS interrupt received.")
+	flag.StringVar(&opts.debugName, "debug.name", "observatorium", "The Name to add as prefix to log lines.")
```
squat: Nit: lowercase name

.drone.yml
kakkoyun: I have just picked `drone` because of familiarity. I'll have look at github actions,

metalmatze: Keep in mind that we want to eventually run these components against deployed Kubernetes cluster, I think, and thus something like Circle would be probably best. However I'm fine with any of the CI to start with.

kakkoyun: Yes, maybe I should go with CircleCI, I want to add basic e2e tests as soon as possible.

bwplotka: Cool, can K8s do it

main.go
bwplotka: so this can be mistaken by `remote read` so maybe `metrics-query-endpoint` ?

.drone.yml
kakkoyun: Unfortunately not the public offering of Drone. I don't know about github, I'll check it.

internal/prober.go
metalmatze: The idea here is to have an arbitrary number of check functions being passed to that handler and once that handler is triggered via an HTTP request it will call these functions and report back whether or not those funcs returned successfully.
For that, there doesn't need to be anything specific in those handlers.

As an example, look at 
https://github.com/heptiolabs/healthcheck/blob/master/types.go#L22:6
https://github.com/heptiolabs/healthcheck/blob/master/handler.go#L75-L103

kakkoyun: @metalmatze Although I really like the idea to have multiple checkers for various checks, I think we should keep it simple for now, if we need more sophisticated checks we can just use `healthcheck` directly. That being said, I have refactored `prober` a bit more.

main.go
```diff
@@ -0,0 +1,144 @@
...
+	flag.StringVar(&opts.listen, "listen", ":8080", "The address on which internal server runs.")
+	flag.StringVar(&opts.gracePeriod, "grace-period", "5s", "The time to wait after an OS interrupt received.")
+	flag.StringVar(&opts.debugName, "debug.name", "observatorium", "The name to add as prefix to log lines.")
+	flag.IntVar(&opts.debugMutexProfileFraction, "debug.mutex-profile-fraction", 10,
+		"The parameter which controls the fraction of mutex contention events that are reported in the mutex profile.")
+	flag.IntVar(&opts.debugBlockProfileRate, "debug.block-profile-rate", 10,
+		"The parameter controls the fraction of goroutine blocking events that are reported in the blocking profile.")
+	flag.StringVar(&opts.logLevel, "log.level", "info", "The log filtering level. Options: 'error', 'warn', 'info', 'debug'.")
+	flag.StringVar(&opts.logFormat, "log.format", internal.LogFormatLogfmt, "The log format to use. Options: 'logfmt', 'json'.")
+	flag.StringVar(&opts.metricsQueryEndpoint, "metrics-query-endpoint", "", "The endpoint which to make queries against for metrics.")
```
squat: Grammar nit:
```suggestion
	flag.StringVar(&opts.metricsQueryEndpoint, "metrics-query-endpoint", "", "The endpoint against which to query for metrics.")
```

main.go
```diff
@@ -0,0 +1,144 @@
...
+	flag.StringVar(&opts.metricsWriteEndpoint, "metrics-write-endpoint", "", "The endpoint which to make write requests against for metrics.")
```
squat: Same:
```suggestion
	flag.StringVar(&opts.metricsWriteEndpoint, "metrics-write-endpoint", "", "The endpoint against which to make write requests for metrics.")
```

squat: Thinking we may want to namespace these flags, e.g.:

```
--metrics.query-endpoint
--metrics.write-endpoint
```
or
```
--endpoint.metrics-query
--endpoint.metrics-write
```

.drone.yml
kakkoyun: I couldn't manage to activate github actions for our organization, since it's still an experimental feature. I'm going with CircleCI.

main.go
kakkoyun: I'll go with the first one, maybe we should even go with:
```
--metrics.query.endpoint
--metrics.write.endpoint
```
Just being presumptions here.

What do you think?

squat: that sgtm. that will allow us to eventually add additional flags, e.g. auth/tls flags, to each area separately

internal/server/server.go
```diff
@@ -0,0 +1,108 @@
...
+	r.Use(middleware.RequestID)
+	r.Use(middleware.RealIP)
+	r.Use(middleware.Recoverer)
+
+	registerProber(r, p)
+	r.Get("/metrics", func(w http.ResponseWriter, r *http.Request) {
+		promhttp.InstrumentMetricHandler(reg, promhttp.HandlerFor(reg, promhttp.HandlerOpts{})).ServeHTTP(w, r)
+	})
+
+	r.Get("/prometheus/query",
```
brancz: I think we'd want this to be versioned and technology independent, so this should be "/v1/metrics/query" no? Plus this needs query and query_range.

brancz:
Very nice!

internal/server/server.go
kakkoyun: 👍I have overlooked those paths, I'll make the necessary changes.

README.md
```diff
@@ -25,7 +25,7 @@ Usage of ./observatorium:
     	The log filtering level. Options: 'error', 'warn', 'info', 'debug'. (default "info")
   -metrics.query.endpoint string
     	The endpoint against which to query for metrics.
-  -metrics.write.endpoint string
+  -metrics.receive.endpoint string
```
brancz: My assumption is that eventually we will have this read from a database as we want to have hard tenancy on the ingestion path, but I think we're fine with this until we actually have that need.

internal/server/server.go
```diff
@@ -48,12 +48,17 @@ func New(logger log.Logger, reg *prometheus.Registry, opts ...Option) Server {
...
-			proxy.New(logger, "/prometheus/write", options.metricsWriteEndpoint, options.proxyOptions...)))
+	queryEndpoint := "/api/v1/metrics/query"
+	r.Get(queryEndpoint,
+		ins.newHandler("query", proxy.New(logger, queryEndpoint, options.metricsQueryEndpoint, options.proxyOptions...)))
+
+	queryRangeEndpoint := "/api/v1/metrics/query_range"
+	r.Get(queryEndpoint,
+		ins.newHandler("query_range", proxy.New(logger, queryRangeEndpoint, options.metricsQueryEndpoint, options.proxyOptions...)))
+
+	receivePath := "/api/v1/metrics/receive"
```
brancz: I realize that the receive component used `receive` here, but I think I actually now like `write` better. Thoughts?

kakkoyun: sgtm. I'll revert it. Somehow receive sounds better for me an instant moment :D

kakkoyun:
@metalmatze @bwplotka Do I have your blessing as well?

metalmatze:
Go ahead. Let's do everything else in smaller PRs :blush:

Makefile
```diff
@@ -0,0 +1,64 @@
...
+all: clean lint test observatorium
+
+tmp/help.txt: clean build
+	mkdir -p tmp
+	-./observatorium --help &> tmp/help.txt
+
+README.md: tmp/help.txt
+	embedmd -w README.md
+
+observatorium: vendor main.go $(wildcard *.go) $(wildcard */*.go)
```
bwplotka: I guess `build` might be more straightforward?

internal/proxy/proxy.go
```diff
@@ -0,0 +1,57 @@
...
+	rev := httputil.ReverseProxy{
+		Director:   director,
+		BufferPool: bufferPool,
+		ErrorLog:   stdErrlogger,
+	}
+
+	return &Proxy{logger: logger, reverseProxy: &rev}
+}
+
+func (p *Proxy) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
```
bwplotka: :+1:

Makefile
kakkoyun: We also have `build`. The distinction is, one of them specifies os and arch, and builds a binary to be used in the container. I can remove `observatorium` if it's creating confusion, in upcoming PRs.

bwplotka: It's fine

PR #3, opened 2019-12-13
Add simple local smoke tests for API

kakkoyun:
This PR,

* adds local Observatorium stack run as binaries with a shellscript
* runs simple smoke tests that ensure proxy endpoints work as expected


docker-compose.yaml
```diff
@@ -0,0 +1,111 @@
...
+    command:
+      - --config.file=/config/prometheus.yml
+      - --log.level=debug
+    volumes:
+      - ./data/test/config:/config
+    ports:
+      - 9090
+    links:
+      - observatorium
+      - thanos-receiver
```
squat: we need links to thanos-store and thanos-querier too, since they are scraped for metrics

docker-compose.yaml
```diff
@@ -0,0 +1,111 @@
...
+      - receive
+      - --grpc-address=0.0.0.0:10901
+      - --http-address=0.0.0.0:10902
+      - --remote-write.address=0.0.0.0:19291
+      - --objstore.config-file=/config/objstore.yml
+      - --tsdb.path=/var/thanos/receive
+      - --label=replica="receive-0"
+      - --label=receive="true"
+      - --tsdb.retention=4h
+      - --receive.hashrings-file=/config/hashrings.json
```
squat: if you only have one replica, then there is no need for a hashring file

metalmatze:
This seems to somewhat duplicate the efforts I currently work on in the observatorium/configuration repo to run e2e tests against the entire stack. This component should be part of that.

I'd say let's keep the e2e tests in here simple and terms of running `httptest` servers and testing those handlers. The rest can more easily be done with the rest of the stack.

kakkoyun:
@metalmatze You're right, I'm changing the PR title and so the purpose. I'll create unit tests for API itself by mocking backing http services.

@squat  @metalmatze I have added proxy routes for UI as well (just to test it basically).
PTAL, if it's ok I'll merge this and move on to deployments bits. Then I'll return and put unit tests in place

metalmatze:
Are the `data/test` files now obsolete?

kakkoyun:
@squat @metalmatze  Will be updated after https://github.com/observatorium/up/pull/2/files merged/

tools.go
```diff
@@ -0,0 +1,8 @@
+// +build tools
```
kakkoyun: @squat This was the missing piece https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module

test/integration.sh
```diff
@@ -0,0 +1,60 @@
...
+    --log.level=debug
+) &
+
+(
+  ./tmp/bin/thanos receive \
+    --grpc-address=127.0.0.1:10901 \
+    --http-address=127.0.0.1:10902 \
+    --remote-write.address=127.0.0.1:19291 \
+    --tsdb.path="$(mktemp -d)"
+#    --log.level=debug \
```
squat: Uncomment this? Or remove?

test/integration.sh
```diff
@@ -0,0 +1,60 @@
...
+) &
+
+(
+  ./tmp/bin/thanos query \
+    --grpc-address=127.0.0.1:10911 \
+    --http-address=127.0.0.1:9091 \
+    --store=127.0.0.1:10901 \
+    --web.external-prefix=http://localhost:8080/ui/v1/metrics
+#    --log.level=debug \
```
squat: Same

Makefile
```diff
@@ -62,3 +94,31 @@ container: observatorium Dockerfile
 .PHONY: container-push-push
 container-push: container
 	docker push $(DOCKER_REPO):$(VERSION) $(DOCKER_REPO):latest
+
+test-dependencies: $(PROMETHEUS) $(THANOS) $(UP) $(EMBEDMD) $(GOLANGCILINT) $(SHELLCHECK)
+
+$(BIN_DIR):
+	mkdir -p $(BIN_DIR)
+
+$(PROMETHEUS): $(BIN_DIR)
```
squat: Is this ever actually used?

PR #11, opened 2020-02-05
Add load tests

kakkoyun:
This PR adds a test to generate synthetic load and records resource usage.

Also,
* adds forgotten CLI parameter,
* observes resource usage,
* observes latencies,
* add load test baseline results.


kakkoyun:
I have put this rough PR together, I used https://github.com/open-fresh/avalanche to generate write load @krasi-georgiev @squat @metalmatze 

However, I also want to consider using more generic load testing tools like https://github.com/tsenart/vegeta, to generate load and fetch latencies while observing resource usage by wiring a Prometheus to the gateway, as in this PR.

One other option to add another mode to https://github.com/observatorium/up, so that it can generate both read and write loads (we can also use Avalanche as a lib).

We briefly discussed this in our sync, however, I'd like to hear your further opinions on these options.

cc @brancz @bwplotka

bwplotka:
image

:crying_cat_face:

kakkoyun:
@bwplotka vendoring 🤷‍♂

squat:
can you split the vendoring into a separate commit? that way we can review the code without reviewing all of the 12,000 lines of vendor :)

kakkoyun:
@squat Commit without vendor https://github.com/observatorium/observatorium/pull/11/commits/a0b6902aa7d2e93fdfd9265f7579be0f1b29327e

metalmatze:
I've tried running this locally with `make test-load`.

So far I've only seen:
```
2020-02-06 14:53:10.572812877 +0100 CET m=+90.009270872: refreshing metric values
14:53:10 write.go:121: updating remote write metrics
	no query failures; checking again in 1s...
	no query failures; checking again in 1s...
	no query failures; checking again in 1s...
	no query failures; checking again in 1s...
	no query failures; checking again in 1s...
```

kakkoyun:
@metalmatze That's why it's WIP 😅I haven't adjusted arguments yet.

kakkoyun:
For the preliminary results, check rendered version of the doc:
https://github.com/observatorium/observatorium/blob/fc55dcf1124881940fdc27a54dc9a352d7419cc1/docs/benchmark.md

docs/benchmark.md
```diff
@@ -0,0 +1,108 @@
...
+Most relevant results are the ones on resource consumption.
+CPU usage is pretty much stable.
+Memory usage correlates with number of goroutine, which correlates number of open connections.
+Memory usage increases and request latencies increase as the backend services' load increase, which is expected.
+
+### Resource consumption
+
+#### CPU Usage
+
+> `rate(process_cpu_seconds_total{job="observatorium"}[1m]) * 100`
```
metalmatze: If we want milli cores we need to do `*1000`, I think.

docs/benchmark.md
```diff
@@ -0,0 +1,108 @@
...
+
+##### Query P50
+
+> histogram_quantile(0.50, sum by (job, le) (rate(http_request_duration_seconds_bucket{job="observatorium", handler="query_range"}[1m])))'
+
+![./loadtests/query_range_dur_50.png](./loadtests/query_range_dur_50.png)
+
+##### Query Average
+
+> sum by (job) (rate(http_request_duration_seconds_sum{job="observatorium", handler="query_range"}[1m])) * 100
```
metalmatze: I'd put the `*100` at the front. Compute the percentage on the inner one `100 * (sum(5xx) / sum(total))`

docs/benchmark.md
```diff
@@ -0,0 +1,108 @@
...
+With parameters:
+
+```console
+$ ./test/load.sh -r 300 -c 1000 -m 3 -q 10 -o gnuplot
+```
+
+> It runs tets for 5 minutes, simulating 3000 machines sending metrics and 10 consumers querying for their data every second.
+> Observatorim API GW runs in-front of 1 Thanos Receive and 1 Thanos Querier.
+
+## Results
```
krasi-georgiev: shouldn't you also include some formula how to calculate the overall resource requremenent for a given load.
I mean something that will answer the  question: what resources would I need for  ingesting 1000 metrics and querying .... 

I remember that Clayton also wanted to know the total amout of memory needed when running a given query  so probably a good idea to add something like this as a follow up PR

Makefile
```diff
@@ -61,11 +76,11 @@ format: $(GOLANGCILINT)
 
 .PHONY: go-fmt
 go-fmt:
-	@fmt_res=$$(gofmt -d -s $$(find . -type f -name '*.go' -not -path './vendor/*' -not -path './jsonnet/vendor/*')); if [ -n "$$fmt_res" ]; then printf '\nGofmt found style issues. Please check the reported issues\nand fix them if necessary before submitting the code for review:\n\n%s' "$$fmt_res"; exit 1; fi
+	@fmt_res=$$(gofmt -d -s $$(find . -type f -name '*.go' -not -path './vendor/*' -not -path './jsonnet/vendor/*' -not -path './tmp/*')); if [ -n "$$fmt_res" ]; then printf '\nGofmt found style issues. Please check the reported issues\nand fix them if necessary before submitting the code for review:\n\n%s' "$$fmt_res"; exit 1; fi
```
krasi-georgiev: can you use the TMP_DIR var here?

tools.go
```diff
@@ -5,6 +5,7 @@ package main
 import (
```
krasi-georgiev: why is this file needed?

test/load.sh
```diff
@@ -0,0 +1,203 @@
+#!/bin/bash
```
krasi-georgiev: is n't this too much for bash? I mean wouldn't this be quite simplified if written in golang?

@observatorium/maintainers what do you think?

tools.go
kakkoyun: This file is here to track versions of the tools that we use for the project, with new go modules. in this context, a tool is binary that is used at build time or development time. For further information, check: https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module

docs/benchmark.md
kakkoyun: I will add a section to document to explain this.

kakkoyun:
@squat I'd love to hear your opinions as well.

docs/benchmark.md
```diff
@@ -0,0 +1,108 @@
+# Benchmark baseline results
+
+This document contains baseline benchmark results for Observatorium API Gateway under synthetic load.
```
bwplotka: What are trying to benchmark here? I assume API Gateway right? So why we need Thanos behind it? You essentially measure Thanos latency not API Gateway latency I guess. So mocked thing might be better for this then.

docs/benchmark.md
```diff
@@ -0,0 +1,108 @@
...
+
+#### CPU Usage
+
+> `rate(process_cpu_seconds_total{job="observatorium"}[1m]) * 100`
+
+![./loadtests/cpu.png](./loadtests/cpu.png)
+
+#### Memory Usage
+
+> `process_resident_memory_bytes{job="observatorium"}'`
```
bwplotka: Not a big deal but wonder where those 200MB came from

docs/benchmark.md
```diff
@@ -0,0 +1,108 @@
...
+
+![./loadtests/mem.png](./loadtests/mem.png)
```
bwplotka: It would be 100000x more readable if that would be Prometheus/Grafana graphs, but not a blocker

docs/benchmark.md
```diff
@@ -0,0 +1,108 @@
...
+
+![./loadtests/goroutines.png](./loadtests/goroutines.png)
+
+### Latencies
+
+#### Write Latency Percentiles
+
+##### Write P99
+
+> histogram_quantile(0.99, sum by (job, le) (rate(http_request_duration_seconds_bucket{job="observatorium", handler="write"}[1m])))'
```
bwplotka: Again.. those latency is Thanos + API latency or just API latency? (:

docs/benchmark.md
```diff
@@ -0,0 +1,108 @@
...
+
+> sum by (job) (rate(http_request_duration_seconds_sum{job="observatorium", handler="write"}[1m])) * 100
+> /
+> sum by (job) (rate(http_request_duration_seconds_count{job="observatorium", handler="write"}[1m]))'
+
+![./loadtests/write_dur_avg.png](./loadtests/write_dur_avg.png)
+
+#### Query Range Latency Quartiles
+
+##### Query P99
```
bwplotka: ditto

docs/benchmark.md
krasi-georgiev: As long as the metrics measure gw latencies I prefer to use some real backend instead of mocking

docs/benchmark.md
krasi-georgiev: I prefer it as well but don't think these can be automated right?

docs/benchmark.md
bwplotka: But GW latency includes backend latency XD You are just benchmarking Thanos here...

docs/benchmark.md
bwplotka: Why not? Just allow this tool to be scraped by Prometheus

docs/benchmark.md
kakkoyun: That is true, what I can do it mock the backend _responses_, so that gateway would behave as it'd behave in real-life with minimal latency. This would give more accurate results. I'm gonna work on it. Thanks @bwplotka

docs/benchmark.md
kakkoyun: I don't have in-depth knowledge about reverse proxy in go, however, this is an interesting point to check out. I'm going to investigate it.

docs/benchmark.md
kakkoyun: This is scraped by Prometheus and graphs are generated by github.com/go-pluto/styx
So I don't get screenshots manually, I think that's what @krasi-georgiev meant. 
Maybe be this could be done by spawning a Grafana instance and getting a couple of snapshots using snapshot API. To be honest, it's a bit overkill for the scope of this PR. We can readdress this if this becomes part of the CI/CD pipeline.

krasi-georgiev: yes, this is what I meant, simple and fully automated image graphs.

kakkoyun:
Results after switching mock provider https://github.com/observatorium/observatorium/blob/51775824fdcdf2747dceaff612c7af0ef8c99149/docs/benchmark.md

docs/benchmark.md
kakkoyun: I have checked it again, it's not 200MB rather 20MB.

kakkoyun: A snapshot of the heap: 
profile004

PR #13, opened 2020-03-02
Add latency and error Alerts

krasi-georgiev:
moved the jb jsonnet files in the jsonnet folder as otherwise jb install deletes everything in the root vendor dir.

to generate from within the jsonnet dir
`jsonnet -J vendor alerts.jsonnet | gojsontoyaml > http-request-latency-burnrate.yaml`

I will add  alerts for all endpoint in the same PR after you check this first one.

cc @kakkoyun , @metalmatze 

 - [ ] how to ensure the le bucket exists ? easy to add a selector for a non existing bucket
 - [ ] recording rules have duplicated names


krasi-georgiev:
hm `jb` doesn't have the `-J` flag

metalmatze:
It has the `-J jsonnet/vendor` in the makefile for `make examples/manifests`:
https://github.com/observatorium/observatorium/blob/master/Makefile#L172

Issue #15, opened 2020-03-04
Add needed APIs to be able to configure as Data-Source in Grafana


rollandf:
Observatorium API should be able to be used as a data source in Grafana.

Currently adding e.g. http://observatorium-api-gateway/api/v1/metrics to Grafana as a Prometheus source fails on "Save & Test".

Maybe we need also  [`GET /api/v1/labels`](https://prometheus.io/docs/prometheus/latest/querying/api/#getting-label-names) and others.

rollandf:
From a tcp dump when add in Grafana, I see that a query fails:

`09:49:52.086709 IP (tos 0x0, ttl 52, id 46199, offset 0, flags [DF], proto TCP (6), length 401)
    10.35.206.161.60166 > 192.168.130.11.http: Flags [P.], cksum 0xa65e (correct), seq 1:350, ack 1, win 506, options [nop,nop,TS val 3522164400 ecr 1965862302], length 349:
 HTTP, length: 349
        GET /api/v1/metrics/api/v1/query?query=1%2B1&time=1583333391.706 HTTP/1.1
        Host: observatorium-api-gateway-observatorium.apps-crc.testing
        User-Agent: Grafana/6.6.2
        Accept: application/json, text/plain, */*
        Accept-Encoding: gzip, deflate
        Accept-Language: en-US,en;q=0.5
        X-Forwarded-For: ::1, ::1
        X-Grafana-Nocache: true
        X-Grafana-Org-Id: 1

09:49:52.088083 IP (tos 0x0, ttl 64, id 49150, offset 0, flags [DF], proto TCP (6), length 349)
    192.168.130.11.http > 10.35.206.161.60166: Flags [P.], cksum 0x1cc8 (incorrect -> 0xc97a), seq 1:298, ack 350, win 235, options [nop,nop,TS val 1965862587 ecr 3522164400
], length 297: HTTP, length: 297
        HTTP/1.1 404 Not Found
        Content-Type: text/plain; charset=utf-8
        X-Content-Type-Options: nosniff
        Date: Wed, 04 Mar 2020 14:49:51 GMT
        Content-Length: 19
        Set-Cookie: 0d123c92f98f343573bb08977e7649ff=596a591919f5f9169b57aac242c3fe4f; path=/; HttpOnly
        Cache-control: private
`

rollandf:
Fixed by #17

PR #17, opened 2020-03-13
Change URL Scheme

### LINKED ISSUE #9: Make API endpoints compatible with Prometheus client_go
Using a better URL path would help to integrate with https://github.com/prometheus/client_golang. 
To be more concrete use `/metrics/api/v1/*` scheme instead of `/api/v1/metrics/*`.

Endpoints are defined as constants,
https://github.com/prometheus/client_golang/blob/803ef2a759d7caaaa0de58e3815f1be4c8b5a42a/api/prometheus/v1/api.go#L119-L139

kakkoyun:
This PR changes API URL scheme from `/api/v1/metrics/` to `/api/metrics/v1`, `{interface}/{domain}/{version}`

There are a couple of reasons that justify these changes:
* To be compatible with existing tools such as Prometheus client libraries, e,g `client_go` and Grafana
(These tools expect a base URL and they add paths such as `api/v1/query`)
* To change versions of domains independently, e.g `/api/metrics/v1` and `/api/logs/v3` 

This is a breaking change,  however, Observatorium API GW hasn't been used in production and released a stable version, so this change should be ok.

Resulting APIs that could be use from outside:
```
/api/metrics/v1/api/v1/query
/api/metrics/v1/api/v1/query_range
/api/metrics/v1/write
```

Fixes #9 , #15.

cc @brancz @metalmatze @krasi-georgiev 

# Changes

* Adds a new flags
  -  `--metrics.read.endpoint` fallback base URL for metrics read requests.
    - (Fallback read URL now used to proxy requests that have unrecognized paths directly to backing service.
* Removes `--metrics.query-range.endpoint`, a flag which was a proxy URL to forward query_range requests. It's not needed anymore.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -48,36 +48,39 @@ func New(logger log.Logger, reg *prometheus.Registry, opts ...Option) Server {
 
 	registerProber(r, p)
 
-	r.Get("/-/metrics", func(w http.ResponseWriter, r *http.Request) {
+	r.Get("/metrics", func(w http.ResponseWriter, r *http.Request) {
 		promhttp.InstrumentMetricHandler(reg, promhttp.HandlerFor(reg, promhttp.HandlerOpts{})).ServeHTTP(w, r)
 	})
 
 	if options.metricsUIEndpoint != nil {
-		r.Get("/metrics/ui/v1/*",
-			ins.newHandler("ui", proxy.New(logger, "/metrics/ui/v1/", options.metricsUIEndpoint, options.proxyOptions...)))
+		r.Get("/ui/metrics/v1/*",
+			ins.newHandler("ui", proxy.New(logger, "/ui/metrics/v1", options.metricsUIEndpoint, options.proxyOptions...)))
 	}
 
-	if options.metricsQueryEndpoint != nil {
-		queryPath := "/metrics/api/v1/query"
-		r.Mount(queryPath,
-			ins.newHandler("query", proxy.New(logger, queryPath, options.metricsQueryEndpoint, options.proxyOptions...)))
-	}
+	namespace := "/api/metrics/v1"
+	r.Route(namespace, func(r chi.Router) {
+		if options.metricsQueryEndpoint != nil {
+			queryPath := "/api/v1/query"
+			r.Mount(queryPath,
+				ins.newHandler("query", proxy.New(logger, namespace+queryPath, options.metricsQueryEndpoint, options.proxyOptions...)))
+		}
 
-	if options.metricsQueryRangeEndpoint != nil {
-		queryRangePath := "/metrics/api/v1/query_range"
-		r.Mount(queryRangePath,
-			ins.newHandler("query_range", proxy.New(logger, queryRangePath, options.metricsQueryRangeEndpoint, options.proxyOptions...)))
-	}
+		if options.metricsQueryRangeEndpoint != nil {
+			queryRangePath := "/api/v1/query_range"
+			r.Mount(queryRangePath,
+				ins.newHandler("query_range", proxy.New(logger, namespace+queryRangePath, options.metricsQueryRangeEndpoint, options.proxyOptions...)))
+		}
 
-	if options.metricsReadEndpoint != nil {
-		readPath := "/metrics/api/v1/*"
-		r.Get(readPath,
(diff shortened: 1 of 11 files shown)


nmagnezi:
I cherry-picked this PR and build a [test image](https://quay.io/repository/nmagnezi/observatorium?tab=tags) to test thing locally.

I currently get:
```
$ kubectl -n observatorium logs observatorium-xyz-observatorium-api-gateway-78db66cd7c-qckfp
level=info msg="--metrics.read.endpoint is not specified"
msg="parse flag" err="--metrics.query.endpoint and --metrics.query-range.endpoint have to be specified"
```

Probably because this PR dictates a change to the args we should provide to observatorium-api-gateway.

Current params:
```
    spec:
      containers:
      - args:
        - --web.listen=0.0.0.0:8080
        - --metrics.ui.endpoint=http://observatorium-xyz-observatorium-api-gateway-thanos-query.observatorium.svc.cluster.local:9090
        - --metrics.query.endpoint=http://observatorium-xyz-cortex-query-frontend.observatorium.svc.cluster.local:9090/api/v1/query
        - --metrics.write.endpoint=http://observatorium-xyz-thanos-receive.observatorium.svc.cluster.local:19291/api/v1/receive
        - --log.level=warn
        image: quay.io/nmagnezi/observatorium:pr-17-2020-03-15-7458993
```

kakkoyun:
@nmagnezi You have to specify `query` and `query-range` separately. It's already pointed out in the error message.

Could you try it as follows? Thanks in advance, it's great that you're testing it :)

```diff
  observatorium \
    --web.listen=0.0.0.0:8080 \
    --metrics.ui.endpoint=http://127.0.0.1:9091/ \
- # This is not used/needed when metrics.query.endpoint and metrics.query-range.endpoint specified.
- # But better to specify it for now, to support other endpoints that have exposed by Thanos Querier.
+   --metrics.read.endpoint=http://127.0.0.1:9091/api/v1 \
    --metrics.query.endpoint=http://127.0.0.1:9091/api/v1/query \
+   --metrics.query-range.endpoint=http://127.0.0.1:9091/api/v1/query_range \
    --metrics.write.endpoint=http://127.0.0.1:19291/api/v1/receive \
```

metalmatze:
If you say so, I trust you on this rename :)

One thing I want to point out though is that Grafana expecting `api/v1/` is a different API like the one we're versioning. We're talking about Observatorium and then about Prometheus' API and they could theoretically be versioned independently.

bwplotka:
Agree with @metalmatze here.

To me, it might either mean `/v1/metrics/api/v1` or `/metrics/v1/api/v1` (:

kakkoyun:
@metalmatze @bwplotka Good point. I'm going to do it. Thanks.

internal/server/server.go
```diff
@@ -48,36 +48,39 @@ func New(logger log.Logger, reg *prometheus.Registry, opts ...Option) Server {
...
+				ins.newHandler("query", proxy.New(logger, namespace+queryPath, options.metricsQueryEndpoint, options.proxyOptions...)))
+		}
+
+		if options.metricsQueryRangeEndpoint != nil {
+			queryRangePath := "/api/v1/query_range"
+			r.Mount(queryRangePath,
+				ins.newHandler("query_range", proxy.New(logger, namespace+queryRangePath, options.metricsQueryRangeEndpoint, options.proxyOptions...)))
+		}
+
+		if options.metricsReadEndpoint != nil {
```
brancz: We don't need to revise this here, but I don't feel like we should differentiate here, just always send this against the same endpoint. The proxy cache will proxy through any requests it doesn't handle, so we can minimize the configuration necessity here.

brancz:
Even if it's slightly odd, it's future proof as it is now, and that's mostly what we were aiming for here, so I'm happy with it.

squat:
Ack I'm happy with it too despite the verbosity. Only one of the versions is actually ours so it's less real complexity than it appears

krasi-georgiev:
agreed, can't think of an alternative either. This suffix is really hard coded in the Prom client and in Grafana.

internal/server/server.go
kakkoyun: 👍 I'll do the changes.

PR #19, opened 2020-03-25
Upgraded the slo-lib

krasi-georgiev:
Signed-off-by: Krasi Georgiev <8903888+krasi-georgiev@users.noreply.github.com>


krasi-georgiev:
yep the generation is missing, I will add.

krasi-georgiev:
Actually now that I think about this I think we should put the alerts directly in the config repo as they are quite specific to our SLO

PR #22, opened 2020-04-01
Upgrade go version to 1.14.1

kakkoyun:


kakkoyun:
As far as I know, it's fixed and there's a new minor version https://golang.org/doc/devel/release.html#go1.14

squat:
I think in general we should be pinning to exact go versions specifically to avoid issues like building with the wrong go image because go1.14 is .0 in some places and .1 in others. How about pinning this to go1.14.1?

kakkoyun:
@squat Makes sense. I've pinned it.

squat:
:+1:

PR #23, opened 2020-04-02
Remove the alerts libsonnet

krasi-georgiev:
Will place this directly in the config repo as it is specific to our
SLOs

Signed-off-by: Krasi Georgiev <8903888+krasi-georgiev@users.noreply.github.com>


kakkoyun:
I think we should keep them here as well. 
I would like to see facilities to monitor the individual components in their own repo.
Could we revert this one?

Although we are all owners, it'd be really nice to have at least one review before we merge PRs.

krasi-georgiev:
yeah this just removed something I added and wasn't used anywhere. I would say let's see how will use the budget SLOs in the actual deployment and when/if there is a use case, can add an example here.

If we add one now would it be for a Prometheus deployment, k8s, Prometheus operator?

Here is an example that can be used directly to generate alerts and rules for the gateway as well
https://github.com/metalmatze/slo-libsonnet/blob/master/examples/http-request-error-burnrate.jsonnet

PR #24, opened 2020-04-03
Verify TLS

kakkoyun:
Verify config and serve TLS

Depends (for integration tests):
xref: https://github.com/observatorium/up/pull/14


internal/tls/config.go
```diff
@@ -0,0 +1,87 @@
...
+	return tlsCfg, nil
+}
+
+// Helpers
+
+func tlsVersion(versionName string) (uint16, error) {
+	var versions = map[string]uint16{
+		"VersionTLS10": tls.VersionTLS10,
+		"VersionTLS11": tls.VersionTLS11,
+		"VersionTLS12": tls.VersionTLS12,
```
brancz, line 79: we've got TLS1.3 now :)

brancz:
Regarding your questions, I think no, both of those projects are already well maintained, and I think it's better to rely on and share that code in a single place and update here instead of doing all the same mistakes again.

main.go
```diff
@@ -124,70 +183,91 @@ func exec(logger log.Logger, reg *prometheus.Registry, opts options) error {
...
+	flag.StringVar(&cfg.tls.certFile, "tls-cert-file", "",
+		"File containing the default x509 Certificate for HTTPS. Leave blank to disable TLS.")
+	flag.StringVar(&cfg.tls.keyFile, "tls-private-key-file", "",
+		"File containing the default x509 private key matching --tls-cert-file. Leave blank to disable TLS.")
+	flag.StringVar(&cfg.tls.clientCAFile, "tls-client-ca-file", "",
+		"File containing the TLS CA to verify clients against."+
+			"If no client CA is specified, there won't be any client verification on server side.")
+	flag.StringVar(&cfg.tls.minVersion, "tls-min-version", "VersionTLS13",
+		"Minimum TLS version supported. Value must match version names from https://golang.org/pkg/crypto/tls/#pkg-constants.")
+	flag.StringVar(&rawTLSCipherSuites, "tls-cipher-suites", "",
```
brancz, line 234: I think we've done some previous research on sane defaults for these, but I think it was based around TLS 1.2, maybe we can do this again for 1.3?

kakkoyun: 👍 Let me look into it.

kakkoyun: Apparently, we cannot even set the `CipherSuite` if we use TLS 1.3. I'm attaching the related documentation. So sane default should be `nil`.

https://golang.org/pkg/crypto/tls/#CipherSuite

```diff
CipherSuites is a list of supported cipher suites for TLS versions up to TLS 1.2. 
If CipherSuites is nil, a default list of secure cipher suites is used, 
with a preference order based on hardware performance. 
The default cipher suites might change over Go versions. 
+Note that TLS 1.3 ciphersuites are not configurable.
```

brancz: Amazing. I didn't even know.

kakkoyun:
@squat I'd love hear your opinions as well.

brancz:
Merge at your own discretion.

krasi-georgiev:
Would it be better to have the tls auto-reload when changed on disk?

Here is an almost ready PR with some very good tests for Thanos
https://github.com/thanos-io/thanos/pull/1672

kakkoyun:
Tests fail because of `up` needs to be updated https://github.com/observatorium/up/pull/14.

PR #25, opened 2020-04-03
Add timeout knobs

kakkoyun:
This adds all the possible knobs to be more flexible for the adjustments.
Also changes the default values.

### DIFF
--- a/internal/proxy/proxy.go
+++ b/internal/proxy/proxy.go
@@ -18,24 +18,26 @@ import (
 )
 
 const (
-	// DefaultBufferCount TODO
+	// DefaultBufferCount is the default value for the maximum size of the buffer pool for the reverse proxy.
 	DefaultBufferCount = 2 * 1024
-	// DefaultBufferSizeBytes TODO
+	// DefaultBufferSizeBytes is the default value for the length of the buffers in the buffer pool for the reverse proxy.
 	DefaultBufferSizeBytes = 32 * 1024
-	// DefaultFlushInterval TODO
+	// DefaultFlushInterval is the default value for the flush interval of reverse proxy to flush to the client while copying the response body.
 	DefaultFlushInterval = time.Duration(-1)
-	// DefaultTimeout TODO
-	DefaultTimeout = 30 * time.Second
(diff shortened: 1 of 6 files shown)


metalmatze:
Do we really want to expose all of these via flags? :thinking: 
Sane defaults should be good enough for 90% of these, don't you think?

kakkoyun:
@metalmatze Yes, happy to remove flags 👍 We have too many of them.

internal/proxy/proxy.go
```diff
@@ -23,6 +24,18 @@ const (
 	DefaultBufferSizeBytes = 32 * 1024
 	// DefaultFlushInterval TODO
 	DefaultFlushInterval = time.Duration(-1)
+	// DefaultTimeout TODO
```
squat, line 27: What is left to do here? Simply right the comment?

internal/proxy/proxy.go
```diff
@@ -23,6 +24,18 @@ const (
...
+	DefaultTimeout = 30 * time.Second
```
squat, line 28: It seems that most of these consts exposed as public identifiers are never used outside of this package. Can we make them unexported if they are not used?

internal/proxy/proxy.go
kakkoyun: yes :)

internal/proxy/proxy.go
```diff
@@ -17,12 +18,26 @@ import (
 )
 
 const (
-	// DefaultBufferCount TODO
+	// DefaultBufferCount The default value for the maximum size of the buffer pool for the reverse proxy.
```
squat, line 21: nit: s/The/is the/

internal/server/server.go
```diff
@@ -17,10 +17,19 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
+// DefaultGracePeriod The default value of the duration gracefully shuts down the server without interrupting any active connections.
```
squat, line 20: same for these consts

squat:
let's squash the commit :)

PR #30, opened 2020-04-20
Fix UI routing

kakkoyun:
Fix and refactor UI routes probably got broken while rebasing.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -65,19 +65,32 @@ func New(logger log.Logger, reg *prometheus.Registry, opts ...Option) Server {
 
 	registerProber(r, p)
 
-	uiPath := "/ui/metrics/v1"
-
-	r.Get("/", func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


kakkoyun:
This is needed before re-routing traffic on stage.

cc @krasi-georgiev

internal/server/server.go
```diff
@@ -65,19 +65,32 @@ func New(logger log.Logger, reg *prometheus.Registry, opts ...Option) Server {
...
+		})
+
+		// NOTICE: Following redirects added to be compatible with existing Read UI.
+		// Paths are explicitly specified to prevent unnecessary request to read handler.
+		for _, p := range []string{
+			"graph",
+			"stores",
+			"status",
+		} {
+			p := p
```
squat, line 89: why do we need this extra declaration here? It looks to me like all uses of `p` are synchronous with the loop

kakkoyun: If you see line 91, we are using it in a closure because of that usage we need to re-declare it in the loop scope.

squat: ack yes I see it now

PR #32, opened 2020-04-21
Only trigger release job for SemVer tags

kakkoyun:
### Changes
* Only trigger release job for SemVer tags

The job gets triggered for all master merges, we just want it to get triggered with tagged versions.
https://quay.io/repository/observatorium/observatorium?tag=latest&tab=tags

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,15 +63,6 @@ workflows:
       - lint
       - test
       - generate
-  tagged-master:
-    jobs:
-      - container-release:
-          filters:
-            tags:
-              only: /^v.*/
-            branches:
-              only:
-                - master
   commit-master:
     jobs:
       - build
@@ -85,3 +76,18 @@ workflows:
             branches:
               only:
                 - master
+  tagged-master:
+    jobs:
+      - container-release:
+          filters:
+            tags:
+              # v1.1.0
+              # v1.1.0-rc0
+              # v1.1.0-rc1
+              # v1.1.0-rc.0
+              # v1.1.0-rc.1
+              only: /^v([0-9]+)\.([0-9]+)\.([0-9]+)(?:-rc(\.)?([0-9]+))?$/
+            branches:
+              only:
+                - master
+                - /release-.*/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,7 +84,8 @@ workflows:
               # Suggested SemVer regex:
               # https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
               # https://regex101.com/r/vkijKf/1/
-              only: /^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$/
+              # NOTICE: with an additional "v" as prefix.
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -85,3 +76,13 @@ workflows:
             branches:
               only:
                 - master
+  tagged-master:
+    jobs:
+      - container-release:
+          tags:
+            only: /^v([0-9]+)\.([0-9]+)\.([0-9]+)$/
```
squat, line 83: With this change we won't build any release candidates; is that what we want?

kakkoyun: Still WIP. I will update the regex to include release candidates.

kakkoyun:
@squat Thanks for the pointer, I didn't aware that an official regex is available. I will use it.

kakkoyun:
@squat @metalmatze Do you see any blockers for this?

squat:
This makes it so that containers are only tagged and pushed to the registry whenever a commit is tagged, right? This is counter to our normal practice of tagging a container for every single commit. I think tagging containers more frequently helps us iterate faster, so I would vote not to merge this change

kakkoyun:
@squat This doesn't change anything that we already have. We still tag image by the commit and push them. This is just an addition to it. It would create a cleaner image tag that matches the semantic version tag.

squat:
ok got it :+1: thanks for the clarification.
 in that case I think there is something a bit funny in the config

Makefile
```diff
@@ -128,8 +128,9 @@ container-push: container
 
 .PHONY: container-release
 container-release: container
+	# https://git-scm.com/docs/git-tag#Documentation/git-tag.txt---points-atltobjectgt
 	@docker tag $(DOCKER_REPO):$(VCS_BRANCH)-$(BUILD_DATE)-$(VERSION) $(DOCKER_REPO):$(VERSION)
-	docker push $(DOCKER_REPO):$(VERSION)
+	docker push $(DOCKER_REPO):$(strip $(shell [ -d .git ] && git tag --points-at HEAD))
```
squat, line 133: The tag getting pushed here `$(DOCKER_REPO):$(strip $(shell [ -d .git ] && git tag --points-at HEAD))` does not match the tag created above `$(DOCKER_REPO):$(VERSION)`

Makefile
```diff
@@ -128,8 +128,9 @@ container-push: container
...
 	docker push $(DOCKER_REPO):latest
```
squat, line 134: `$(DOCKER_REPO):latest` is not tagged in this make target. So if you do `make container-release` without first doing `make container-push`, you won't have a `latest` tag to push

kakkoyun: It's tagged a `container` stage which is a dependency of this action.

Dockerfile
```diff
@@ -2,9 +2,12 @@ FROM golang:1.14.1-alpine3.11 as builder
 
 RUN apk add --update --no-cache ca-certificates tzdata git make bash && update-ca-certificates
 
-ADD . /opt
 WORKDIR /opt
 
+COPY go.* Makefile /opt/
+RUN make vendor
```
squat, line 8: This project already vendors it's dependencies, why do we `make vendor` again before building? If we don't trust the deps we committed then we should do a `make vendor && git diff --exit-code` as part of the CI. Otherwise, we should definitely use the checked-in vendor code for our builds. No?

kakkoyun: Yes, we already implicitly call `vendor` before we build. That's why I want to this explicitly. Also, we pass the vendor dir along so this is just like an integrity check. This additional thing won't hurt, will it? It definitely speeds things up.

squat: mmm i'd say let's keep this PR focused; this is about ` Only trigger release job for SemVer tags` so let's put this unrelated change into a new PR.

Also, why do we even do another `make vendor` before building if we know that the deps are checked into the repo? CI does an integrity check with `make vendor && git diff --exit-code`. There should be no reason for another explicit  `make vendor`, instead, we should declare the vendor dir and go.* files deps of the binary build target, so that `go mod tidy && go mod vendor` is only run when actually necessary. WDYT? Otherwise what we should really do IMO is not commit our vendor dir, which is also a valid path

kakkoyun: This is what we already have. Then I'm reverting the changes and we're good.

PR #34, opened 2020-04-22
Fix TLS error message

kakkoyun:


internal/tls/config.go
```diff
@@ -59,7 +59,7 @@ func NewServerConfig(logger log.Logger, certFile, keyFile, clientCAFile, minVers
 
 		certPool := x509.NewCertPool()
 		if !certPool.AppendCertsFromPEM(caPEM) {
-			return nil, fmt.Errorf("building client CA: %w", err)
+			return nil, errors.New("building client CA failed")
```
squat, line 62: nit rather than `building`, since we are not generating a CA, consider `parsing`

PR #35, opened 2020-04-23
Add query and query_range as generic handlers forwarding all verbs

metalmatze:
Next to adding the `/api/v1/query_range` endpoint too, the handlers are now simple handlers accepting every verb and then letting the downstream service decide on accepting or declining requests with specific verbs.

/cc @squat @kakkoyun @brancz


cben:
(Background: clusters-service, which is one of the backends of [OCM](https://cloud.redhat.com/openshift), is doing POST to /api/v1/query to avoid query length limits.  
I don't think we're hitting them yet, but we're planning to send long `_id=~"...|...|..."` lists to query the clusters an OCM user is authorized for, so we're always sending POST.)

metalmatze:
That's good insight to why a POST makes more sense than a GET. I was already wondering why you want POST :)

PR #38, opened 2020-04-27
 rbac: initial RBAC implementation

squat:
This commit adds an initial implementation of RBAC for Observatorium
along with unit tests. RBAC rules can be read from a YAML file and look
a lot like Kubernetes RBAC rules. The only supported permissions are
`read` and `write` at the moment.

Once identification middleware is added to the stack, we can create e2e
tests for authorization.

cc @metalmatze


rbac/rbac.go
```diff
@@ -0,0 +1,121 @@
...
+
+const (
+	// Write gives access to write data to a tenant.
+	Write Permission = "write"
+	// Read gives access to read data from a tenant.
+	Read Permission = "read"
+)
+
+// Role describes a set of permissions to interact with a tenant.
+type Role struct {
```
brancz, line 21: I think we talked about adding `resource` here, to control what exactly about this tenant the role is allowed to do.

brancz, line 21: I think we talked about adding `resource` here, to control what exactly about this tenant the role is allowed to do.

rbac/rbac.go
```diff
@@ -0,0 +1,121 @@
...
+	Name     string   `json:"name" yaml:"name"`
+	Subjects []string `json:"subjects" yaml:"subjects"`
+	Roles    []string `json:"roles" yaml:"roles"`
+}
+
+// Authorizer can auth a subject against a tenant for a permission.
+type Authorizer interface {
+	// Authorize answers the question: can subject S perform permission P on Tenant T?
+	Authorize(subject string, permission Permission, tenant string) bool
+}
```
metalmatze, line 38: We don't have the consumer yet, so I'm fine having it here. Going foward we should move it to the place it's going to be consumed. :) Do you want to add a `// TODO: ` or good like this?

rbac/rbac.go
squat: Added a `resources` field so a role can control access to multiple permissions on multiple resources for multiple tenants :+1:

metalmatze:
slick :ok_hand:

PR #39, opened 2020-04-28
Create internal server

metalmatze:
Next to creating the internal server we also move to metalmatze/signal healtcheck and internalserver to improve reusability across projects.


main.go
```diff
@@ -165,16 +162,27 @@ func exec(logger log.Logger, reg *prometheus.Registry, cfg config) error {
 		}
 	}
 	{
+		{
+			// checks if server is up
+			healthchecks.AddLivenessCheck("http",
+				// TODO: Make use of cfg.server.listen to dynamically configure this too
+				healthcheck.HTTPCheck("http://localhost:8080/", http.MethodGet, http.StatusMovedPermanently, time.Second),
```
metalmatze, line 169: Do you have an idea on how to extract this address from `cfg.server.listen`? I guess it can always use the localhost as host. We need to switch between https and the port depending on how they are configured. I guess for HTTP we could actually accept any kind of certificate, or do you think we should load the same and verify that it works too?

internal/server/server.go
```diff
@@ -32,18 +30,14 @@ const DefaultWriteTimeout = 2 * time.Minute
...
-	prober *prober.Prober
 	srv    *http.Server
 
 	opts options
 }
 
 // New creates a new Server.
 func New(logger log.Logger, reg *prometheus.Registry, opts ...Option) Server {
-	options := options{
-		gracePeriod: DefaultGracePeriod,
```
kakkoyun, line 44: It's ok to remove the flag, but I think we should consider specifying a default value. At least `120s` it's the default for our queries to timeout.

main.go
metalmatze: The problem is, that we need to parse something like `:8080` and `url.Parse` actually fails with `parse ":8081": missing protocol scheme`.

squat: hmm, if we explicitly create listeners via `l, err := net.Listen("tcp", cfg.server.listen)`, then we can use the `Addr` method of the listener to get the address: https://golang.org/pkg/net/#Listener

This gets us most of the way there, I think. Then we need to look for two special cases `[::]` (IPv6 all addresses) and `0.0.0.0` (IPv4 all addresses) and we can likely sub those both for localhost + the scheme prefix depending on whether certs were given or not

main.go
```diff
@@ -244,6 +253,8 @@ func parseFlags(logger log.Logger) (config, error) {
 		"The address on which public server runs.")
 	flag.StringVar(&cfg.server.listenInternal, "web.internal.listen", ":8081",
 		"The address on which internal server runs.")
+	flag.StringVar(&cfg.server.healthcheckURL, "web.healthchecks.url", "http://localhost:8080",
+		"The URL like http://localhost:8080 on which public server runs and to run healthcheck tests against.")
```
squat, line 257: I think we can leave out `like http://localhost:8080`, since URL says it all already.
Also, nit: the other flag descriptions don't end in a `.`. I prefer periods but let's choose one.

```suggestion
		"The URL at which the public server runs and against which to run healthcheck tests ")
```

metalmatze: All usages end with a period. :thinking: So does mine?!

squat: gah yes you were right, i must have mis-read :see_no_evil:

internal/server/server.go
```diff
@@ -169,8 +139,3 @@ func (s *Server) Shutdown(err error) {
 		level.Error(s.logger).Log("msg", "shutting down failed", "err", err)
 	}
 }
-
-func registerProber(r *chi.Mux, p *prober.Prober) {
-	r.Get("/-/healthy", p.HealthyHandler())
```
clyang82, line 174: I think we need to update here - https://github.com/observatorium/observatorium/blob/7470f0a697f459ec7db7ec7e73de0dbd638cfff2/jsonnet/lib/observatorium-api.libsonnet as well since we removed the `prober`. right now, I am using the `quay.io/observatorium/observatorium:latest`, the pod cannot be running.
```
  Warning  Unhealthy  52m (x3 over 53m)      kubelet, ip-10-0-137-85.us-east-2.compute.internal  Liveness probe failed: HTTP probe failed with statuscode: 404
  Warning  BackOff    8m16s (x28 over 21m)   kubelet, ip-10-0-137-85.us-east-2.compute.internal  Back-off restarting failed container
  Warning  Unhealthy  3m12s (x475 over 53m)  kubelet, ip-10-0-137-85.us-east-2.compute.internal  Readiness probe failed: HTTP probe failed with statuscode: 404
```

metalmatze: Yes, that needs updating. I'm surprised people use `:latest` though. Best example why not to use `:latest`. :wink:

PR #40, opened 2020-05-04
Split HTTP server into multiple metrics packages - make it composable

metalmatze:
Essentially, rather than having a big Server instance that's used across the project, I tend to think of creating smaller bits and pieces, in this case handlers per package, and then assemble them in the main.go again.

That means that `internal/api/metrics/v1/` is responsible for `/api/metrics/v1` on the HTTP endpoints and `internal/api/metrics/legacy` is actually mounted to `/api/v1/query`. The package itself however, is totally unaware of that path prefix and could be moved independently or even mounted somewhere else without modifying the package itself.
The proxy also became more like a middleware collection where we use bits and pieces to create proxies per handler (and even multiple ones) as we can have more fine-grained control over timeouts and other values.

/cc @squat @kakkoyun @brancz


metalmatze:
This depends on #39

brancz:
We can move ahead here now that #39 is merged :)

internal/proxy/proxy.go
```diff
@@ -45,7 +47,7 @@ type Proxy struct {
 	reverseProxy *httputil.ReverseProxy
 }
 
-func New(logger log.Logger, prefix string, endpoint *url.URL, opts ...Option) *Proxy {
+func NewBig(logger log.Logger, prefix string, endpoint *url.URL, opts ...Option) *Proxy {
```
metalmatze, line 50: I think this can go already within this PR.

internal/proxy/proxy.go
```diff
@@ -109,3 +111,46 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			m(r)
+		}
+	}
+}
+
+func MiddlewareSetUpstream(upstream *url.URL) Middleware {
+	return func(r *http.Request) {
+
+		r.URL.Scheme = upstream.Scheme
+		r.URL.Host = upstream.Host
```
metalmatze, line 129: We still need to do something like:
`
```suggestion
		r.URL.Scheme = upstream.Scheme
		r.URL.Host = upstream.Host
		r.URL.Path = path.Join(upstream.Path, r.URL.Path)
```

internal/api/metrics/v1/http.go
```diff
@@ -0,0 +1,149 @@
...
+					}).DialContext,
+				},
+			}
+		}
+		r.Handle("/write", c.instrument.NewHandler(
+			prometheus.Labels{"group": "metricsv1", "handler": "write"},
+			http.StripPrefix("/write",
+				proxyWrite,
+			),
+		))
```
metalmatze, line 145: I'm, wondering if we shouldn't just go with `/receive` here too. That's what the Thanos Receiver does. I'm pretty sure we wouldn't need that StripPrefix thing then.
/cc @brancz @squat

brancz: Yeah let’s rather be consistent

metalmatze:
Updated the write path to be `/api/v1/receive`. It's pretty nice now too, as it looks like Thanos Querier and Thanos Receiver are one component from the outside (which is the ultimate goal of this).

We now have:
```
/api/metrics/v1/api/v1/query
/api/metrics/v1/api/v1/query_range
/api/metrics/v1/api/v1/receive
```

PR #42, opened 2020-05-11
Update configuration to make use of internal server

metalmatze:
We recently added an internal server that exposes healthchecks and metrics. We need to adjust the configuration to appropriately make use if it.
Additionally, I set the livenessProbe to fail after 10*30s = 300s = 5min and readinessProbe to fail after 12*5s = 60s = 1min.

/cc @squat @clyang82


Makefile
```diff
@@ -219,24 +216,22 @@ CONTAINER_CMD:=docker run --rm \
...
-${MANIFESTS}: jsonnet/main.jsonnet jsonnet/lib/* | $(JSONNET) $(GOJSONTOYAML)
-	@rm -rf ${MANIFESTS}
-	@mkdir -p ${MANIFESTS}
-	$(JSONNET) -J jsonnet/vendor -m ${MANIFESTS} jsonnet/main.jsonnet | xargs -I{} sh -c 'cat {} | $(GOJSONTOYAML) > {}.yaml && rm -f {}' -- {}
+examples/manifests: examples/main.jsonnet examples/vendor jsonnet/lib/* | $(JSONNET) $(GOJSONTOYAML)
+	@rm -rf examples/manifests
+	@mkdir -p examples/manifests
+	$(JSONNET) -J examples/vendor -m examples/manifests examples/main.jsonnet | xargs -I{} sh -c 'cat {} | $(GOJSONTOYAML) > {}.yaml && rm -f {}' -- {}
 
-.PHONY: jsonnet-vendor
```
squat, line 235: CI complains with:
```
make: *** No rule to make target 'jsonnet-vendor', needed by 'generate'.  Stop.
```
If we remove this, then we need to update the circle CI config too

metalmatze: Yeah, looking into this already. Sorry, I should have commented, I guess. :innocent:

PR #44, opened 2020-05-18
Redirect to UI including query params

metalmatze:
This is mostly so that in case people have bookmarked queries or there are links to graphs saved elsewhere still work and they get forwarded accordingly.

If someone has a bookmark / link to `graph?g0.range_input=1h&g0.max_source_resolution=0s&g0.expr=up&g0.tab=1` then going forward that will be redirected including the query parameters, so that those links don't break.

/cc @squat

### DIFF
--- a/main.go
+++ b/main.go
@@ -211,7 +211,12 @@ func main() {
 		)
 
 		r.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
-			http.Redirect(w, r, "/api/metrics/v1/graph", http.StatusMovedPermanently)
+			r.URL.Path = "/api/metrics/v1/graph"
+			http.Redirect(w, r, r.URL.String(), http.StatusMovedPermanently)
+		})
+		r.HandleFunc("/graph", func(w http.ResponseWriter, r *http.Request) {
+			r.URL.Path = "/api/metrics/v1/graph"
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -211,7 +211,12 @@ func main() {
 		)
 
 		r.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
-			http.Redirect(w, r, "/api/metrics/v1/graph", http.StatusMovedPermanently)
+			r.URL.Path = "/api/metrics/v1/graph"
+			http.Redirect(w, r, r.URL.String(), http.StatusMovedPermanently)
+		})
+		r.HandleFunc("/graph", func(w http.ResponseWriter, r *http.Request) {
```
squat, line 217: this one should probably be part of the `legacy` API package, right? Since it's supporting an old route?

metalmatze: While I generally agree, the legacy package is actually mounted on `/api/v1/`. So adding it there won't work unless we mount that router to `/`. I'd say it's fine, there's basically nothing that handler is doing. :man_shrugging:

squat: I'm just worried about code sprawl, ie how long to we want to maintain this redirect. putting it here in the main may cement it as part of the API

PR #46, opened 2020-05-20
internal/authorization: create authz middleware

squat:
This commit implements and utilizes an authorization middleware to limit
API usage to subjects who have access to resources belonging to a
tenant, as specified in an RBAC configuration file.

At the heart of it, this commit simply ties the existing rbac.Authorizer
interface to a middleware. One implementation detail is that in order to
identify a subject, we need to find a username from the OIDC token. We
now allow the tenant to specify the field that should be used to
uniquely identify subjects, just as the Kubernetes apiserver does
with its `--oidc-username-claim` flag. The subject's name wil default to
the value of the `sub` field, but can be changed to, e.g., `email` in
order to have nicer usernames in RBAC role bindings.

cc @metalmatze @brancz

### DIFF
--- a/main.go
+++ b/main.go
@@ -33,14 +33,17 @@ import (
 	metricslegacy "github.com/observatorium/observatorium/internal/api/metrics/legacy"
 	metricsv1 "github.com/observatorium/observatorium/internal/api/metrics/v1"
 	"github.com/observatorium/observatorium/internal/authentication"
+	"github.com/observatorium/observatorium/internal/authorization"
 	"github.com/observatorium/observatorium/internal/server"
 	"github.com/observatorium/observatorium/internal/tls"
+	"github.com/observatorium/observatorium/rbac"
 )
 
 type config struct {
 	logLevel  string
 	logFormat string
 
+	rbacConfigPath    string
 	tenantsConfigPath string
 
 	debug   debugConfig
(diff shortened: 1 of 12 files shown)


squat:
ping @metalmatze

main.go
```diff
@@ -270,6 +288,8 @@ func main() {
 							metricsv1.Logger(logger),
 							metricsv1.Registry(reg),
 							metricsv1.HandlerInstrumenter(ins),
+							metricsv1.ReadMiddleware(authorization.WithAuthorizer(authorizer, rbac.Read, "metrics")),
+							metricsv1.WriteMiddleware(authorization.WithAuthorizer(authorizer, rbac.Write, "metrics")),
```
metalmatze, line 292: While this feels a bit weird, and I would prefer to be explicit, it allows us to inject different middlewares from the outside to the package, without the need to change anything in the package, when changing middlewares. 
I guess that's the reasoning?

squat: yes exactly that's the idea. I want to keep the same package API, where the `New...` func simply returns a `http.Handler` and the package has no concept of the authorization/authentication middlewares. I think the only nice way to do that is to inject middlewares into the creator func. The other option is to make the API packages accept an `rbac.Authorizer` option and build and use the middlewares themselves. WDYT?

metalmatze: I think it's fine for now. We might need to come back to the version you just said, but that might be ahead of time right now.

squat: :+1:

PR #47, opened 2020-05-25
Add loki query and push proxy support

periklis:
This PR adds proxy support to Loki `/loki/api/v1/{query, query_range, push}` and `/ready` endpoints. In addition it extends the api gateway CLI parameters to declare a read and write endpoint for the Loki.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,9 @@ THANOS_VERSION ?= 0.12.2
 PROMETHEUS ?= $(BIN_DIR)/prometheus
 PROMETHEUS_VERSION ?= 2.15.2
 
+LOKI ?= $(BIN_DIR)/loki
+LOKI_VERSION ?= 1.5.0
+
 UP ?= $(BIN_DIR)/up
 DEX ?= $(BIN_DIR)/dex
 MOCKPROVIDER ?= $(BIN_DIR)/mockprovider
@@ -135,7 +138,8 @@ container-release: container
 	docker push $(DOCKER_REPO):latest
 
 .PHONY: integration-test-dependencies
-integration-test-dependencies: $(THANOS) $(UP) $(DEX)
+
+integration-test-dependencies: $(THANOS) $(UP) $(DEX) $(LOKI)
 
 .PHONY: load-test-dependencies
 load-test-dependencies: $(PROMREMOTEBENCH) $(PROMETHEUS) $(STYX) $(MOCKPROVIDER)
@@ -162,6 +166,14 @@ $(THANOS): | $(BIN_DIR)
 $(PROMETHEUS): | $(BIN_DIR)
 	@echo "Downloading Prometheus"
 	curl -L "https://github.com/prometheus/prometheus/releases/download/v$(PROMETHEUS_VERSION)/prometheus-$(PROMETHEUS_VERSION).$$(go env GOOS)-$$(go env GOARCH).tar.gz" | tar --strip-components=1 -xzf - -C $(BIN_DIR)
+
+$(LOKI): | $(BIN_DIR)
+	@echo "Downloading Loki"
+	(loki_pkg="loki-$$(go env GOOS)-$$(go env GOARCH)" && \
+	cd $(BIN_DIR) && curl -O -L "https://github.com/grafana/loki/releases/download/v$(LOKI_VERSION)/$$loki_pkg.zip" && \
+	unzip $$loki_pkg.zip && \
+	mv $$loki_pkg loki && \
+	rm $$loki_pkg.zip)
 
 $(UP): | vendor $(BIN_DIR)
 	go build -mod=vendor -o $@ github.com/observatorium/up
--- a/internal/api/logs/v1/http.go
+++ b/internal/api/logs/v1/http.go
@@ -1,0 +1,154 @@
+package http
+
+import (
+	"net"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+	"path"
+	"time"
+
+	"github.com/go-chi/chi"
+	"github.com/go-kit/kit/log"
+	"github.com/observatorium/observatorium/internal/proxy"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	ReadTimeout        = 15 * time.Minute
+	WriteTimeout       = time.Minute
+	lokiUpstreamPrefix = "/loki"
+)
+
+type handlerConfiguration struct {
+	logger           log.Logger
+	registry         *prometheus.Registry
+	instrument       handlerInstrumenter
+	readMiddlewares  []func(http.Handler) http.Handler
+	writeMiddlewares []func(http.Handler) http.Handler
+}
+
+// HandlerOption modifies the handler's configuration
+type HandlerOption func(h *handlerConfiguration)
+
+// Logger add a custom logger for the handler to use.
+func Logger(logger log.Logger) HandlerOption {
+	return func(h *handlerConfiguration) {
+		h.logger = logger
+	}
+}
+
+// Registry adds a custom Prometheus registry for the handler to use.
+func Registry(r *prometheus.Registry) HandlerOption {
+	return func(h *handlerConfiguration) {
+		h.registry = r
+	}
+}
+
+// HandlerInstrumenter adds a custom HTTP handler instrument middleware for the handler to use.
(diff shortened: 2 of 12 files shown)


Makefile
```diff
@@ -160,6 +163,13 @@ $(PROMETHEUS): | $(BIN_DIR)
 	@echo "Downloading Prometheus"
 	curl -L "https://github.com/prometheus/prometheus/releases/download/v$(PROMETHEUS_VERSION)/prometheus-$(PROMETHEUS_VERSION).$$(go env GOOS)-$$(go env GOARCH).tar.gz" | tar --strip-components=1 -xzf - -C $(BIN_DIR)
 
+$(LOKI): | $(BIN_DIR)
+	@echo "Downloading Loki"
+	(cd $(BIN_DIR) && curl -O -L "https://github.com/grafana/loki/releases/download/v$(LOKI_VERSION)/loki-linux-$$(go env GOARCH).zip" && \
```
squat, line 168: we like to track our binary dependencies using the tools.go pattern. See the examples below in the Makefile and the tools.go at the root of the repo to add a new dependency

squat:
Please let us know if you have any questions about fixes so we can get this merged :)))))

Makefile
squat: this will also fix the CI test, which is complaining that the `unzip` binary isn't available

internal/api/logs/v1/http.go
```diff
@@ -0,0 +1,145 @@
+package http
```
squat, line 1: It looks like we need a gofmt on this file

internal/api/logs/v1/http.go
```diff
@@ -0,0 +1,145 @@
...
+                Director: middlewares,
+                ErrorLog: proxy.Logger(c.logger),
+                Transport: &http.Transport{
+                    DialContext: (&net.Dialer{
+                        Timeout: ReadTimeout,
+                    }).DialContext,
+                },
+            }
+        }
+        r.Handle("/loki/api/v1/query", c.instrument.NewHandler(
```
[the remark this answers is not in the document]
squat: from discussion with @metalmatze looks like this is in fact the case. Question: do we want to remove the `/loki` prefix from our API and add it instead proxy config so that paths get rewritten?

internal/api/logs/v1/http.go
periklis: Ahh my goimports hit this file.

Makefile
periklis: Need to figure out the appropriate replace between loki/prometheus/azure-autorest first.

test/integration.sh
```diff
@@ -56,7 +66,34 @@ if ./tmp/bin/up \
...
+     --cert ./tmp/certs/client.pem \
+     --key ./tmp/certs/client.key \
+     -s "https://127.0.0.1:8443/api/logs/v1/loki/api/v1/query" \
+     --data-urlencode 'query=sum(rate({job="varlogs"}[10m])) by (level)'; then
+  result=0
+  printf "\t## logs read tests: ok"
+  exit 0
+fi
+
+if curl \
```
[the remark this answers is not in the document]
squat: this looks better. can we modify the read test to verify that the written logs are correctly returned?

internal/api/logs/v1/http.go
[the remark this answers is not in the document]
periklis: Do you think of `/logs/api/v1/query` rewritten to `/loki/api/v1/query` internally?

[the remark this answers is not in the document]
squat: maybe simply rewriting `/api/v1/query` to `/loki/api/v1/query` so we match the metrics API. Also, that way we avoid a double `/logs/` in the complete API, i.e. `/api/v1/logs` + `/logs/api/v1/query`

Makefile
squat: grrr we always hit this. In thanos we use this: https://github.com/thanos-io/thanos/blob/master/go.mod#L74

periklis: Unfortunately! let's keep it for now as it is and add unzip to the CI settings. WDYT?

squat: If you don't mind, I'll add it to your commit directly so we can have it here rather than mess with the CI just to merge this PR

internal/api/logs/v1/http.go
```diff
@@ -0,0 +1,149 @@
...
+	for _, o := range opts {
+		o(c)
+	}
+
+	r := chi.NewRouter()
+
+	if read != nil {
+		var proxyRead http.Handler
+		{
+			read.Path = fmt.Sprintf("%s%s", read.Path, lokiUpstreamPrefix)
```
squat, line 78: let's use `path.Join` here rather than `fmt.Sprintf` just for safety in case anything unexpected is passed

Makefile
```diff
@@ -163,6 +167,13 @@ $(PROMETHEUS): | $(BIN_DIR)
...
+	unzip loki-linux-$$(go env GOARCH).zip && \
+	mv loki-linux-$$(go env GOARCH) loki && \
+	rm loki-linux-$$(go env GOARCH).zip)
```
metalmatze, line 175: While we set GOARCH we don't set GOOS, this is going to break for people who would like to build on a Mac, I think.

internal/api/logs/v1/http.go
```diff
@@ -0,0 +1,149 @@
...
+			)
+
+			readyProxy = &httputil.ReverseProxy{
+				Director: middlewares,
+			}
+		}
+		r.Mount("/ready", c.instrument.NewHandler(
+			prometheus.Labels{"group": "logsv1", "handler": "ready"},
+			readyProxy,
+		))
```
metalmatze, line 119: Can you explain why we would need to expose the ready endpoint to a user of Observatorium? Isn't this an internal implementation detail?
Like, we might want to use that `/ready` endpoint to check with Kubernetes if the endpoint of a Pod is to be added to a Service, but not expose it to the user direct, am I missing something about Loki here? :)

test/integration.sh
```diff
@@ -77,10 +91,45 @@ if ./tmp/bin/up \
...
+     -v \
+     -H "Content-Type: application/json" \
+     --cacert ./tmp/certs/ca.pem \
+     --cert ./tmp/certs/client.pem \
+     --key ./tmp/certs/client.key \
+     -XPOST \
+     -s "https://127.0.0.1:8443/api/logs/v1/api/v1/push" \
+     --data-raw \
+     '{"streams": [{ "stream": { "foo": "bar2" }, "values": [ [ "1570818238000000000", "fizzbuzz" ] ] }]}'; then
+  result=0
```
squat, line 112: this new logic currently introduces some potential for danger. If the metrics test fails (result is never set to 0) but the logs test succeeds (result is set to 0 later), then this entire script exits 0. We need to either exit after each failed test, or find another way to combine errors after all tests have run

internal/api/logs/v1/http.go
periklis: I introduced this initially to add [here](https://github.com/observatorium/observatorium/pull/47/files#diff-7ddfb3e035b42cd70649cc33393fe32cR217) an extra healthcheck ReadinessCheck agains loki for observatorium.

test/integration.sh
periklis: I suggest we go for a fail fast approach.

squat: cool that simplifies things  :)

test/integration.sh
```diff
@@ -77,10 +91,54 @@ if ./tmp/bin/up \
...
 
-printf "\t## tests: failed" 1>&2
-result=1
-exit 1
+echo "-------------------------------------------"
+echo "- Logs Write tests                        -"
+echo "-------------------------------------------"
+
+if curl \
+     -v \
```
squat, line 106: do we need to add `--fail` to these curl requests? otherwise if there is a 404 or 500 or something, this will still exit 0. Also, do we need to somehow test the content of the response? how do we know that we got back what we were expecting?

periklis: `--fail` is a good hint. However, checking response content is way to much a simple integration test that assert proxying requests/responses. I am not aware if `observatorium/up` has the capability to assert such things. If yes, I would be glad to contribute it there instead of solving this in bash. WDYT?

squat: Agreed. Up doesn't check the content against an expected response but it DOES ensure the status code is 200

PR #48, opened 2020-05-25
.circleci: fix tagged master job

squat:
Currently, merging to master triggers the tagged-master job, because it
seems that CircleCI filters are ORed. This causes the Makefile to try to
tag a docker container with a trailing `:` since the Git tag is emtpy.
This commit changes the logic so that the tagged-master job is only
triggered on a valid SemVer tag.

cc @kakkoyun


squat:
-______- this isn't right

kakkoyun:
Hmm, I've remembered why we had that branch filter 🙈 I started to think, this wasn't a good at all. My bad bad.

PR #51, opened 2020-05-26
jsonnet,examples: add RBAC + tenants to manifests

squat:
This PR adds RBAC and tenants configuration files to the generated
manifests by extending the jsonnet library for the API.

cc @metalmatze @periklis

### DIFF
--- a/examples/main.jsonnet
+++ b/examples/main.jsonnet
@@ -6,13 +6,56 @@ local api = (import '../jsonnet/lib/observatorium-api.libsonnet') {
     version: 'master-2020-05-04-v0.1.1-21-gabb9864',
     image: 'quay.io/observatorium/observatorium:' + cfg.version,
     replicas: 3,
-    metrics:{
+    metrics: {
       readEndpoint: 'http://127.0.0.1:9091',
       writeEndpoint: 'http://127.0.0.1:19291',
     },
-    logs:{
+    logs: {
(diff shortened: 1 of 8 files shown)


squat:
That's true, I actually started with `withTenants` and `withRBAC`, but I saw that we use this other pattern, e.g. : https://github.com/observatorium/configuration/blob/master/components/up.libsonnet#L93

I'm very happy to move it to `with...` if you two like it more

periklis:
I am not the jsonnet pro here. I'll pass to @metalmatze . The `with...` approach gives me the impression that it works best when amending with new or overwrite named fields. It maybe be akward replacing in arrays though.

metalmatze:
It's a tradeoff for sure right now and we're still experimenting.
The problem with all the `withFoo` functions is that it scatters the configuration across multiple different functions / objects in the file. It's super hard to grasp at a certain point.
We've seen that happening in our downstream / internal repository that uses https://github.com/observatorium/configuration/. 
For the API I'm actually fine with the approach in this PR right now.

squat:
cool, then I'm happy with going this way for now too

periklis:
@metalmatze you have a point here, scattering configuration is hard for maintenance. I kept the `withFoo` approach in my loki configuration code for the sort like service monitors that introduction optional resources.

Issue #54, opened 2020-05-31
generate-tls-cert does not parse cmd line arguments


nmagnezi:
The generate-tls-cert tool gets used in integration tests for this repository, and since those don't run on top of kubernetes, [defaults](https://github.com/observatorium/observatorium/blob/3ae9b88c4fc9c58a1e04da564d1fc90741854c11/test/tls/generate.go#L43-L44) work fine.

However, for other use cases such as testing TLS based deployments as done in [PR-273](https://github.com/observatorium/configuration/pull/273), the aforementioned defaults won't work:

`
caller=level.go:63 ts=2020-05-31T08:04:29.825399028Z name=up level=error caller=main.go:184 component=reader msg="failed to query" err="query request failed: Post https://observatorium-xyz-observatorium-api.observatorium.svc.cluster.local:8080/api/metrics/v1/api/v1/query: x509: certificate is valid for localhost, not observatorium-xyz-observatorium-api.observatorium.svc.cluster.local"
`

`generate-tls-cert` should process command line arguments to allow custom certificates.

squat:
Closed by #55

nmagnezi:
thanks for the prompt response @squat!
could you please issue an image tag as well? so this addition will get used in `observatorium/configuration` repo CI?

squat:
No need to manually build the images. CI automatically builds and tags images for every merge to master. Check out the quay repo to find the most recent tag :)

PR #55, opened 2020-05-31
Process cmd line arguments in generate-tls-cert

nmagnezi:
This patch adds a missing call to Parse()[1],
so cmd line arguments will get processed, making
the created certificates valid for additional use cases.

Issue: #54

[1] https://gobyexample.com/command-line-flags


squat:
Would be nice to add an integration test that actually uses this functionality in order to catch regressions and to test that the took actually works

PR #56, opened 2020-06-02
Change default TLS version to VersionTLS12

rollandf:
When configuring Prometheus Remote Write with TLS to send to
Observatorium API, we get the following error:

 `http: TLS handshake error from 10.128.0.248:54642: tls: client offered only unsupported versions: [303 302 301]`

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Usage of ./observatorium:
   -tls-client-ca-file string
     	File containing the TLS CA against which to verify clients.If no client CA is specified, there won't be any client verification on server side.
   -tls-min-version string
-    	Minimum TLS version supported. Value must match version names from https://golang.org/pkg/crypto/tls/#pkg-constants. (default "VersionTLS13")
+    	Minimum TLS version supported. Value must match version names from https://golang.org/pkg/crypto/tls/#pkg-constants. (default "VersionTLS12")
   -tls-private-key-file string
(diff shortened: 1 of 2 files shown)


squat:
This is a default flag, why can't the default simply be overridden in your deployment?

rollandf:
Then it should be the default in `configuration`, no?
Won't  be Prometheus one of the main client working against Observatorium API?

squat:
Lint is broken right now due to a change in the distribution of shellcheck binaries. This should be fixed by: https://github.com/observatorium/observatorium/pull/58

squat:
hmm prometheus supports TLS 1.3, so I'm not sure about this changeset

rollandf:
This is the error we got when configuring RemoteWrite with TLS

squat:
Did you find a fix for the error? Was it a configuration problem?

rollandf:
No, I did not find a fix. Only changing the default version worked.
We encountered it on OCP, with the mTLS configuration.

I will try to find a simple reproduction scenario.

squat:
ok sounds good. ideally we would want both Prom and the Observatorium API to use TLS 1.3, so if there is something wrong at any end of the stack, let's try to fix that rather than downgrade the ciphers

rollandf:
Did you test the API with prom and Remote Write with mTLS?

PR #61, opened 2020-06-07
Generate both TLS and mTLS manifests

nmagnezi:


examples/main.jsonnet
```diff
@@ -64,8 +64,15 @@ local apiWithTLS = api + api.withTLS {
...
     tls+: {
       certFile: './tmp/certs/server.pem',
-      privateKeyFile: './tmp/certs/ca.pem',
-      clientCAFile: './tmp/certs/server.key',
+      privateKeyFile: './tmp/certs/server.key',
+    },
+  },
+};
+
+local withClientTLS = apiWithTLS + api.withClientTLS {
```
squat, line 72: Since this is both TLS and client TLS, maybe consider naming this local `withMTLS`

jsonnet/lib/observatorium-api.libsonnet
```diff
@@ -251,4 +249,31 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
       },
     },
   },
+
+  withClientTLS:: {
+    local api = self,
+
+    config+:: {
+      ClientTLS: {
```
squat, line 257: I think it's a little awkward to have a separate `tls` and `clientTLS` field (note that the field name should be lowercased rather than `ClientTLS`), since it leads to stuttering: `config.clientTLS.clientCAFile`. It also adds mental overhead to remembering all of the fields names, ie if we have `clientTLS` then why is the other field `tls` rather than `serverTLS`? I think it would be better to nest this key directly in the `tls` field.

nmagnezi: Done

examples/main.jsonnet
nmagnezi: Done

squat:
looks good @nmagnezi thanks for your patience :)

PR #62, opened 2020-06-07
Enhance integration tests to use all TLS config options

nmagnezi:


brancz:
I'm a bit worried about extending the integration tests shell script even further, I was already a bit uneasy about ~100LOC, and now doubling this seems to me like we should start to convert this to go based tests.

Wdyt @metalmatze @squat ?

squat:
I agree with @brancz. I also think it is unnecessarily thorough for the integration test. I would be satisfied with running the test just once for the manifests and once for the operator where and to always use all TLS options. In order words, I think we can simply test the most complex configuration and if we want to move to a more powerful go-based test-suite, then we can test more configurations.

nmagnezi:
Ack. Closing.

PR #63, opened 2020-06-07
Adapt GOOS both for linux and darwin

nmagnezi:


Makefile
```diff
@@ -67,7 +68,7 @@ benchmark.md: $(EMBEDMD) tmp/load_help.txt
 	$(EMBEDMD) -w docs/benchmark.md
 
 observatorium: vendor main.go $(wildcard *.go) $(wildcard */*.go)
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on GOPROXY=https://proxy.golang.org go build -mod vendor -a -ldflags '-s -w' -o $@ .
+	CGO_ENABLED=0 GOOS=$(GOOS_NAME) GOARCH=amd64 GO111MODULE=on GOPROXY=https://proxy.golang.org go build -mod vendor -a -ldflags '-s -w' -o $@ .
```
squat, line 71: Why are we setting this (non-)overridable variable? If we simply do `go build` then the binary will get built for the architecture and OS of the machine this is running on. This seems like a roundabout way of getting the same behavior but in a less portable way IMO

Did you mean to use `?=` to allow overriding the OS?

nmagnezi: I'm fine with removing GOOS altogether if that will allow it to adapt to the OS.

squat: Yes, that is the default behavior of the go toolchain :) Doing `go build` without specifying the GOOS and GOARCH variables makes the toolchain compile for the local platform so that it can be run locally :)

squat: However, note that forcing the toolchain to compile to the binary for the OS and architecture of the local platform, which this PR does and would also be done by removing the variables, is not a good idea for CI. What if CI wasn't on amd64, then we'd be building and pushing docker images for e.g. 32bit windows.

nmagnezi: I see. So falling back to an overridable variable.
I just noticed an existing (and unused) OS variable that I can make use of.

PR #67, opened 2020-06-16
api-deployment manifest handling for volumes and volumeMounts

nmagnezi:
The reason for this change is that the operator cannot modify or append settings to existing manifests. It merely builds the config by squashing CR provided config and default config and invoke manifest generation.

The meaning of this is that the `api-deployment` manifest generator to be “smarter” and be able to append anything related to TLS settings when those are provided by the user (either manually in `config` or CR when driven by an operator).

### DIFF
--- a/examples/main.jsonnet
+++ b/examples/main.jsonnet
@@ -60,19 +60,24 @@ local api = (import '../jsonnet/lib/observatorium-api.libsonnet') {
   },
 };
 
-local apiWithTLS = api + api.withTLS {
+local apiWithTLS = api + {
   config+:: {
     tls+: {
-      certFile: './tmp/certs/server.pem',
-      privateKeyFile: './tmp/certs/server.key',
+      secret: {
+        certFile: '/mnt/certs/server.pem',
+        privateKeyFile: '/mnt/certs/server.key',
+        reloadInterval: '1m',
+      },
     },
   },
 };
 
-local withMTLS = apiWithTLS + api.withMTLS {
+local withMTLS = apiWithTLS + {
   config+:: {
-    tls+: {
-      clientCAFile: './tmp/certs/ca.pem',
+    mtls+: {
+      configMap: {
+        clientCAFile: '/mnt/clientca/ca.pem',
+      },
     },
   },
 };
--- a/jsonnet/lib/observatorium-api.libsonnet
+++ b/jsonnet/lib/observatorium-api.libsonnet
@@ -27,6 +27,8 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
 
     rbac: {},
     tenants: {},
+    tls: {},
+    mtls: {},
 
     commonLabels:: {
       'app.kubernetes.io/name': 'observatorium-api',
@@ -84,6 +86,16 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
         if api.config.tenants != {} then
           ['--tenants.config=/etc/observatorium/tenants.yaml']
         else []
+      ) + (
+        if api.config.tls != {} then
+          ['--tls-cert-file=' + api.config.tls.secret.certFile,
+           '--tls-private-key-file=' + api.config.tls.secret.privateKeyFile,
+           '--tls-reload-interval=' + api.config.tls.secret.reloadInterval,]
+        else []
+      ) + (
+        if api.config.mtls != {} then
+          ['--tls-client-ca-file=' + api.config.mtls.configMap.clientCAFile]
+        else []
       )) +
       container.withPorts([
         containerPort.newNamed(api.config.ports[name], name)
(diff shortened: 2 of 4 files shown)


examples/main.jsonnet
```diff
@@ -60,19 +60,24 @@ local api = (import '../jsonnet/lib/observatorium-api.libsonnet') {
...
 };
 
-local apiWithTLS = api + api.withTLS {
+local apiWithTLS = api + {
   config+:: {
     tls+: {
-      certFile: './tmp/certs/server.pem',
-      privateKeyFile: './tmp/certs/server.key',
+      secret: {
+        certFile: './mnt/certs/server.pem',
```
squat, line 67: This should be an absolute path

jsonnet/lib/observatorium-api.libsonnet
```diff
@@ -101,51 +113,79 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
...
-             },
              name: 'tenants',
+             mountPath: '/etc/observatorium/tenants.yaml',
+             subPath: 'tenants.yaml',
+             readOnly: true,
+           },
+         ] else []) +
+        (if api.config.tls != {} then [
+           {
+             name: 'observatorium-api-tls-secret',
```
squat, line 135: Nit: let's try to keep resource names stutter-free, i.e. we don't need to append `secret` to the madness of our secrets, just like we don't add `deployment` to the madness of our deployments

jsonnet/lib/observatorium-api.libsonnet
```diff
@@ -101,51 +113,79 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
...
+             mountPath: '/mnt/certs',
+             readOnly: true,
+           },
+         ] else []) +
+        (if api.config.mtls != {} then [
+           {
+             name: 'observatorium-api-tls-configmap',
+             mountPath: '/mnt/clientca',
```
squat, line 143: Same here

jsonnet/lib/observatorium-api.libsonnet
```diff
@@ -101,51 +113,79 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
...
-    if api.config.rbac != {} || api.config.tenants != {} then
-      deployment.mixin.spec.template.spec.withVolumes(
+      container.withVolumeMounts(
         (if api.config.rbac != {} then [
            {
-             configMap: {
-               name: api.config.name,
-             },
              name: 'rbac',
+             mountPath: '/etc/observatorium/rbac.yaml',
```
squat, line 120: Is this necessary? I don't think we should be exposing also of this in configuration as it makes the API surface extra large and harder to change later IMO

jsonnet/lib/observatorium-api.libsonnet
nmagnezi: I still need to separate the configMap and secret mounts. how should I keep the names seperate?

examples/main.jsonnet
nmagnezi: done

jsonnet/lib/observatorium-api.libsonnet
nmagnezi: so I cannot name  both secret and configmap `observatorium-api-tls, (will override each other in volume mounts. how should those be named?

squat: hmm sorry I really mean this in the volumes section, not the volume mounts. The secret's own name should not contain the word `secret`. It is clear it is a secret already :)

jsonnet/lib/observatorium-api.libsonnet
```diff
@@ -101,51 +113,79 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
...
+           secret: {
+             secretName: api.config.name,
+           },
+           name: 'tenants',
+         },
+       ] else []) +
+      (if api.config.tls != {} then [
+         {
+           secret: {
+             secretName: 'observatorium-api-tls-secret',
```
squat, line 175: here

jsonnet/lib/observatorium-api.libsonnet
```diff
@@ -101,51 +113,79 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
...
+           },
+           name: 'observatorium-api-tls-secret',
+         },
+       ] else []) +
+      (if api.config.mtls != {} then [
+         {
+           configMap: {
+             name: 'observatorium-api-tls-configmap',
+           },
+           name: 'observatorium-api-tls-configmap',
```
squat, line 185: and we could rename this mount to something more useful, like `client-ca`

nmagnezi:
> Looks fine overall but I have some comments. What is motivating the need for this PR? In general, please consider adding some explanation in the PR description body; it's is much friendlier for reviewers to interact with when some context is given

all done.

PR #70, opened 2020-06-17
Add tail endpoint for logs

periklis:
This PR addresses a Loki specific distinction between read and tail requests. Read requests target the query frontend that is backed by the queriers and handles read concurrency and caching. Howeve, tail requests target the queriers directly.

**Reference:** [Grafana Gateway Configuration](https://github.com/grafana/loki/blob/master/production/ksonnet/loki/gateway.libsonnet#L43)

/cc @squat

### DIFF
--- a/internal/api/logs/v1/http.go
+++ b/internal/api/logs/v1/http.go
@@ -76,7 +76,7 @@ func (n nopInstrumentHandler) NewHandler(labels prometheus.Labels, handler http.
 	return handler.ServeHTTP
 }
 
-func NewHandler(read, write *url.URL, opts ...HandlerOption) http.Handler {
+func NewHandler(read, tail, write *url.URL, opts ...HandlerOption) http.Handler {
 	c := &handlerConfiguration{
 		logger:     log.NewNopLogger(),
 		registry:   prometheus.NewRegistry(),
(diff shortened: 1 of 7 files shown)


internal/api/logs/v1/http.go
```diff
@@ -121,6 +121,35 @@ func NewHandler(read, write *url.URL, opts ...HandlerOption) http.Handler {
...
+				ErrorLog: proxy.Logger(c.logger),
+				Transport: &http.Transport{
+					DialContext: (&net.Dialer{
+						Timeout: ReadTimeout,
+					}).DialContext,
+				},
+			}
+		}
+		r.Group(func(r chi.Router) {
+			r.Use(c.readMiddlewares...)
```
squat, line 145: Do you think this is sufficient to use the authorization middleware enforcing the `read` permission? Or do we need a new `tail` or `watch` permission? In other words, is tailing any different from a security perspective than reading? e.g. more load on server?

periklis: @squat IMHO there is no need to have explicit permissions for tailing the logs. The read endpoint can do the same thing without the capability to stream newer logs.

Besides permissions, do you think it is better to call this endpoint watch instead of tail? Tail is closer from a terminology standpoint to Loki and logs, but watch is closer to k8s. WDYT?

squat: I don't have a strong opinion about the endpoint name. I would have said `watch` so that it could be generic, but we likely won't have analogous endpoints for other signals, i.e. there won't be a tail/watch for metrics/traces/profiles etc so we can make a one-off `tail` endpoint if we want

periklis: Ok! Let's keep it tail then.

periklis:
@squat PTAL please :+1:

PR #71, opened 2020-06-19
internal/authentication: enable mTLS auth

squat:
Currently, the Observatorium API has overlapping authentication
concepts:
* the API has a global mTLS setting to validate client certs; and
* each tenant can authenticate clients via OIDC.

It is confusing and messy that these overlap. Identify should be
mediated entirely by the mechanism chosen by the tenant and the API's
TLS configuration should provide a secure transport.

This commit removes the global mTLS settings of the API and enables
mTLS-based authentication on a per-tenant basis.
This means that each tenant can specify a separate CA to authenticate
all requests for resources they own.

cc @observatorium/maintainers

### DIFF
--- a/internal/authentication/oidc.go
+++ b/internal/authentication/oidc.go
@@ -1,0 +1,209 @@
+package authentication
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/coreos/go-oidc"
+	"github.com/go-chi/chi"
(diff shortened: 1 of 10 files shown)


internal/authentication/oidc.go
```diff
@@ -0,0 +1,209 @@
+package authentication
```
squat, line 1: this has simply been split from the existing http.go file

PR #75, opened 2020-06-24
test/integration.sh: verify liveness check works

squat:
This commit ensures that the integration test exercises the liveness
check.

cc @nmagnezi @observatorium/maintainers


squat:
amazing, no need then

PR #76, opened 2020-06-25
internal,rbac: enable groups in rbac

squat:
This commit enables specifying groups in the RBAC file so that roles can
be bound to a set of users rather than just individually per user. Groups
are discovered in OIDC using the claim specified in each tenant's OIDC
configuration. In mTLS, groups are populated from the list of the client
certificate's organizational units.

cc @observatorium/maintainers


brancz:
merge when you're ready

PR #81, opened 2020-07-09
test: Use latest stable Thanos for integration tests

kakkoyun:
Signed-off-by: Kemal Akkoyun <kakkoyun@gmail.com>

### DIFF
--- a/test/integration.sh
+++ b/test/integration.sh
@@ -17,16 +17,16 @@ echo "-------------------------------------------"
 sleep 2
 
 token=$(curl --request POST \
-    --silent \
-    --cacert ./tmp/certs/ca.pem \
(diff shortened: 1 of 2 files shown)


test/integration.sh
```diff
@@ -50,6 +50,7 @@ token=$(curl --request POST \
   ./tmp/bin/thanos receive \
     --receive.hashrings-file=./test/config/hashrings.json \
     --receive.local-endpoint=127.0.0.1:10901 \
+    --receive.default-tenant-id="1610b0c3-c509-4592-a256-a1871353dbfa" \
```
kakkoyun, line 53: This is needed because otherwise, the first request will be failing while we test and the results will be below %100.

PR #84, opened 2020-08-07
Make loki as an optional

clyang82:


main.go
```diff
@@ -100,6 +100,8 @@ const (
 	middlewareTimeout
 )
 
+var enableLogs = true
```
periklis, line 103: Consider removing this global variable and group it under `logsConfig`. You can further enable this flag in `logsConfig` if at least one `{read,write,tail}Endpoint}` is provided.

clyang82: Done

main.go
```diff
@@ -542,26 +546,40 @@ func parseFlags() (config, error) {
...
+		if err != nil {
+			return cfg, fmt.Errorf("--logs.tail.endpoint is invalid, raw %s: %w", rawLogsTailEndpoint, err)
+		}
+
+		cfg.logs.tailEndpoint = logsTailEndpoint
 	}
 
-	cfg.logs.writeEndpoint = logsWriteEndpoint
+	if rawLogsWriteEndpoint != "" {
+
```
squat, line 573: nit: extra newline here

clyang82: done

clyang82:
@squat Do I need to commit `modules.txt` file? I remembered it has some difference in the past. but right now, in my local environment, there is no difference after running `make generate --always-make`

squat:
spoke offline, the commit needed to remove the modules.txt from the PR, which was generated using go 1.13, causing an issue.

squat:
lgtm, this simply needs a rebase now :)))

PR #85, opened 2020-08-07
internal/authentication: don't fail broken oidc

squat:
Currently, if a single tenant has a broken OIDC configuration, then
entire Observatorium API will fail to start. This commit ensures that
broken configs are treated as warnings, allowing the rest of the API to
continue running. Tenants with broken configs will not be able to
authenticate and will get `bad request` responses when accessing their
URLs, since for all intents and purposes, that tenant does not actually
exist.

This code has been tested by appending a bad OIDC configuration to the
test/config/tenants.yaml file and running the integration tests.

cc @metalmatze


squat:
cc @observatorium/maintainers

PR #87, opened 2020-08-28
jsonnet: make TLS settings work with K8s Secret

squat:
Previously, the TLS settings required configuring Jsonnet code AND
manually creating and deploying a magically-named Secret containing
the TLS data. This commit refactors this code so that the Secret is
automatically generated and mounted from provided TLS data.

The deployments PR (https://github.com/observatorium/deployments/pull/273) needs a big refactor in order to work and these changes will support and simplify that code.

cc @metalmatze

### DIFF
--- a/examples/main.jsonnet
+++ b/examples/main.jsonnet
@@ -67,12 +67,10 @@ local api = (import '../jsonnet/lib/observatorium-api.libsonnet') {
 local apiWithTLS = api {
   config+:: {
     tls+: {
-      secret: {
-        serverCertFile: '/mnt/certs/server.pem',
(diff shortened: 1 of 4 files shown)


periklis (approved):
LGTM

metalmatze:
cool! :+1:

marcolan018:
@squat 
does the tls settings supports existing k8s secret, aka mounts an existing secret directly?

squat:
That's right, this PR got a bit ahead of itself and eliminated that functionality. But we need it for the operator and to be more decoupled in general. I've prepared some commits to address this while still cleaning up some of the work from before

PR #88, opened 2020-09-03
allow setting server name in healthcheck TLS configuration

squat:
Currently, the TLS configuration used by the healthchecker HTTP client
always defaults to verifying the server name of the certificate using
the hostname of the healthcheck URL. This does not work in most
production configurations, where the certificate issued by a real CA,
e.g. Let's Encrypt, only includes the domain name in the subject. This
means that verifying the subject, e.g. example.com against the hostname,
i.e. `localhost`, will always fail.

This commit enables configuring the server name to use for certificate
verification.

cc @observatorium/maintainers @clyang82


kakkoyun (approved):
🥇

clyang82:
Thanks @kakkoyun

PR #89, opened 2020-09-04
jsonnet: decouple deployment from secret

squat:
PR #87 was intended to decouple the deployment from a secret with a
magic, hard-coded name, but it obligated all consumers of this library
to render a secret and configmap using this jsonnet, rather than truly
decouple the deployment from the configuration files. This meant that
e.g. a custom resource for the Observatorium operator would need to have
the certificates in-line in order to render this manifest, which caused
two issues:
1. it put secrets in otherwise plain-text files; and
2. it prevented (re-)using secrets that may be created by another
service, e.g. cert-manager.

This PR fixes this by truly decoupling the deployment and the
configuration files. Now, the jsonnet accepts a secret name and keys to
specify where the API can find the certificates it needs.

cc @clyang82 @observatorium/maintainers @marcolan018

### DIFF
--- a/examples/manifests/deployment-with-tls.yaml
+++ b/examples/manifests/deployment-with-tls.yaml
@@ -5,7 +5,7 @@ metadata:
     app.kubernetes.io/component: api
     app.kubernetes.io/instance: observatorium-api
     app.kubernetes.io/name: observatorium-api
-    app.kubernetes.io/version: master-2020-06-26-v0.1.1-105-gc784d77
+    app.kubernetes.io/version: master-2020-09-04-v0.1.1-131-ga4c5a9c
   name: observatorium-api
   namespace: observatorium
 spec:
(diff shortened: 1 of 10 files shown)


examples/manifests/deployment-with-tls.yaml
```diff
@@ -74,16 +75,28 @@ spec:
           name: tenants
           readOnly: true
           subPath: tenants.yaml
-        - mountPath: /mnt/tls
-          name: tls
+        - mountPath: /mnt/tls/cert
```
clyang82, line 78: Can we mount to `/var/run/secrets/tls` instead of `/mnt/tls`?

squat: How about we do `/var/run/tls`? that was we can leave the CA in the same directory even though it's not a secret

clyang82: why CA is from configmap?

squat: because it's not a secret and is potentially shared with many other Pods that should not be mounting the TLS certs

periklis (approved):
LGTM

PR #91, opened 2020-09-07
allow serverName empty

clyang82:


periklis:
@clyang82 Do the healthcheck work under TLS w/o server name defined?

clyang82:
> @clyang82 Does the healthcheck work under TLS w/o server name defined?

This does not work in most production environment. but it works in test environment.

squat (approved):

squat:
> Does the healthcheck work under TLS w/o server name defined?

Yes it does but then you need to have `localhost` as one of the SANs in the cert

PR #92, opened 2020-09-08
jsonnet: enable TLS without ConfigMap

squat:
Some certificate providers do not put the CA used to sign a
certificate in a ConfigMap but rather in the same Secret as the
certificate itself, e.g. cert-manager. This commit enables configuring
the CA for the Observatorium API using the same Secret that contains
the certificate and thus enables operation with cert-manager.

The logic is as follows if a caKey is given:
1. if no configMapName is provided, then mount the CA from the same
Secret as the certificate and key; else
2. if a configMapName is provided, then mount the CA from the ConfigMap.

In other words, when no configMapName is specified, the Jsonnet will
default to mounting the CA from the Secret.

cc @observatorium/maintainers @clyang82


kakkoyun (approved):
💯

PR #93, opened 2020-09-08
Mount mtls ca

clyang82:
Allow mount the mTLS CA and follow this logic:
The logic is as follows if a caKey is given:
- if a SecretName is provided, then mount the CA from the SecretName.
- if a configMapName is provided, then mount the CA from the ConfigMap

And reconfigure the tenants.yaml file to have `caPath` if there is `mTLS`

/assign @squat

### DIFF
--- a/jsonnet/lib/observatorium-api.libsonnet
+++ b/jsonnet/lib/observatorium-api.libsonnet
@@ -146,6 +146,17 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
              readOnly: true,
            },
          ] else []) +
+        (if api.config.tenants.tenants != {} then [
+            {
+              name: tenant.name + '-mtls-%s' % (if std.objectHas(tenant.mTLS, 'configMapName') then 'configmap' else 'secret'),
+              mountPath: '/var/run/mtls/' + tenant.name + '/' + tenant.mTLS.caKey,
+              subPath: tenant.mTLS.caKey,
+              readOnly: true,
+            },
+            for tenant in api.config.tenants.tenants
+            if std.objectHas(tenant, 'mTLS')
+            if std.objectHas(tenant.mTLS, 'caKey')
+         ] else []) +
         (if api.config.tls != {} then [
            {
              name: 'tls-secret',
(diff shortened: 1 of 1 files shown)


jsonnet/lib/observatorium-api.libsonnet
```diff
@@ -146,6 +146,17 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
              readOnly: true,
            },
          ] else []) +
+        (if api.config.tenants.tenants != {} then [
```
squat, line 149: `api.config.tenants.tenants` will allways be an array if it exists. So I think it's better to check `if std.objectHas(api.config.tenants, 'tenants')`

jsonnet/lib/observatorium-api.libsonnet
```diff
@@ -194,6 +205,28 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
            name: 'tenants',
          },
        ] else []) +
+      (if api.config.tenants.tenants != {} then [
```
squat, line 208: same as above

jsonnet/lib/observatorium-api.libsonnet
```diff
@@ -194,6 +205,28 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
...
+          {
+           configMap: {
+             name: tenant.mTLS.configMapName,
+           },
+           name: tenant.name + '-mtls-configmap',
+          },
+          for tenant in api.config.tenants.tenants
+          if std.objectHas(tenant, 'mTLS')
+          if std.objectHas(tenant.mTLS, 'configMapName')
```
squat, line 228: I think it would be clearer if we avoided this duplication and changed this to and `else if` statement above.

jsonnet/lib/observatorium-api.libsonnet
clyang82: done

jsonnet/lib/observatorium-api.libsonnet
clyang82: Done

squat (approved):
lgtm

PR #94, opened 2020-09-16
Create sa instead of use default

clyang82:
Pod analysis shows most observability pods use the default service account with cluster-admin access. A separate service account should be used by the pods so they can each be bound to only the privileges they need. None of them should be using the default service account or the wild card privileges `[map[apiGroups:[*] resources:[*] verbs:[*]] map[nonResourceURLs:[*] verbs:[*]]]` that the default service account currently uses.

### DIFF
--- a/examples/manifests/deployment-with-tls.yaml
+++ b/examples/manifests/deployment-with-tls.yaml
@@ -87,6 +87,7 @@ spec:
           name: tls-configmap
           readOnly: true
           subPath: ca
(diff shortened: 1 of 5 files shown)


periklis:
@clyang82 Could you please provide a meaningful description? The referenced repository is not accessible for everyone.

clyang82:
@periklis Sorry for inconvenience. Just update the description.

squat (approved):
Lgtm

PR #95, opened 2020-09-18
Add a test/run-local.sh script run via make run for development

metalmatze:
I thought that may be helpful. :)

/cc @squat @kakkoyun @periklis

### DIFF
--- a/test/run-local.sh
+++ b/test/run-local.sh
@@ -1,0 +1,75 @@
+#!/bin/bash
+
+# Runs a minimal setup with local binaries locally.
+# This should be most helpful when working on Observatorium API specific features.
+
+set -euo pipefail
+
+result=1
+trap 'kill $(jobs -p); exit $result' EXIT
+
+(./tmp/bin/dex serve ./test/config/dex.yaml) &
+
+(
+  ./observatorium \
+    --web.listen=0.0.0.0:8443 \
+    --web.internal.listen=0.0.0.0:8448 \
+    --web.healthchecks.url=https://localhost:8443 \
+    --tls.server.cert-file=./tmp/certs/server.pem \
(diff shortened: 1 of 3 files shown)


periklis:
I don't have a special opinion on this. I am more of fan using kind, but I pass to @blockloop who used to have woes to test the promtail manifests.

kakkoyun (dismissed):
✅

metalmatze:
@periklis, this is mostly about developing the API itself, like literally changing its code and for that running the binary without even building containers is still the fastest, and maybe even simplest. :)

test/run-local.sh
```diff
@@ -0,0 +1,75 @@
...
+    --tsdb.path="$(mktemp -d)"
+) &
+
+(
+  ./tmp/bin/thanos query \
+    --grpc-address=127.0.0.1:10911 \
+    --http-address=127.0.0.1:9091 \
+    --store=127.0.0.1:10901 \
+    --log.level=error \
+    --web.external-prefix=/ui/metrics/v1
```
squat, line 49: I think this external prefix is not quite correct. I am finding that the static assets aren't loading correctly when I run this

metalmatze:
UI now works properly too, when running locally. :chart_with_upwards_trend:

squat:
there seems to be something funny with the commits in this branch as they include stuff from master but rebased. Can you take a look?

squat:
Otherwise LGTM!

metalmatze:
I wasn't able to push and then rebased. Something went wrong.

metalmatze:
Rebased on master after merging #96 and now the PR seems correct.

squat (approved):
nice! LGTM

PR #96, opened 2020-09-18
Remove ksonnet from Kubernetes objects

metalmatze:
We're slowly moving away from using ksonnet.
This is the beginning of that.
We need add a [kubeval](https://www.kubeval.com/) to validate all generated YAMLs. :)


Makefile
```diff
@@ -49,7 +49,6 @@ GOLANGCILINT ?= $(FIRST_GOPATH)/bin/golangci-lint
 GOLANGCILINT_VERSION ?= v1.21.0
 EMBEDMD ?= $(BIN_DIR)/embedmd
 JSONNET ?= $(BIN_DIR)/jsonnet
-JSONNET_BUNDLER ?= $(BIN_DIR)/jb
```
squat, line 52: Can we also remove the jsonnet bundler installation target on line 222?

squat (approved):
:))) nice

PR #98, opened 2020-09-21
Redirect to OIDC login if no cookie

metalmatze:
If the user didn't provide an Auth Bearer token nor there's a cookie available - we assume that the call was most likely a user via browser and thus redirect to the OIDC login page.

### DIFF
--- a/internal/authentication/oidc.go
+++ b/internal/authentication/oidc.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"path"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


periklis (approved):

squat:
This means that if a machine client also incorrectly does not include a bearer token, the response it Receivers will not be a 401, but a FOUND. Will this be confusing to people configuring clients?

metalmatze:
It is true that even for API requests that do not include a bearer token the response going forward will be 302 rather than a 401.
That's the only change. Additionally, the Location header is set now, but that shouldn't hurt.
In the future, if it does become a problem we could only return Location header with 302 code, if we detected a browser by its UserAgent, but that seems a bit extreme for now.

PR #102, opened 2020-10-09
Add rate limiting per tenant

kakkoyun:
Signed-off-by: Kemal Akkoyun <kakkoyun@gmail.com>

This PR adds an ability to specify rate limits per tenant for each 
endpoint individually. Limits can be specified using tenant configuration.
The rate-limiting middleware can be initialized with global mode
to coordinate state between instances (using [`gubernator`](https://github.com/mailgun/gubernator)`*`)
 or with local mode.

`*` Gubernator is a distributed, high performance, cloud-native 
and stateless rate limiting service.

##### Used Algorithms
Local mode: [`Sliding Window`](https://en.wikipedia.org/wiki/Sliding_window_protocol)
Global mode: [`Leaky Bucket`](https://en.wikipedia.org/wiki/Leaky_bucket)

### Changes

* Adds new rate limiter middleware.
* Add grpc tooling to generate `gubernator` client code.


Makefile
```diff
@@ -233,8 +235,7 @@ $(JSONNET_FMT): vendor |  $(BIN_DIR)
 	go build -mod=vendor -o $@ github.com/google/go-jsonnet/cmd/jsonnetfmt
 
 $(KUBEVAL): $(BIN_DIR)
-	go get -d github.com/instrumenta/kubeval
```
kakkoyun, line 236: This was causing inconsistencies between CI jobs.

test/integration.sh
```diff
@@ -37,6 +37,18 @@ token=$(curl --request POST \
...
   --data scope="openid email" | sed 's/^{.*"id_token":[^"]*"\([^"]*\)".*}/\1/')
 
+(GUBER_MEMBERLIST_KNOWN_NODES=127.0.0.1:7946 gubernator-v0.8.4-0.20200617200142-07e238f8cd86) &
+
+
+echo "-------------------------------------------"
+echo "- Waiting for Gubernator to come up...  -"
+echo "-------------------------------------------"
+
+until curl --output /dev/null --silent --fail --insecure http://127.0.0.1:80/v1/HealthCheck; do
```
kakkoyun, line 47: There is bug in memberlist SD implementation of gubenator that restraints us to change the default port. Memberlist SD won't be used for production.

go.mod
```diff
@@ -11,28 +11,39 @@ require (
...
-	golang.org/x/net v0.0.0-20200324143707-d3edc9973b7e // indirect
 	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d
-	golang.org/x/sys v0.0.0-20200413165638-669c56c373c4 // indirect
 	google.golang.org/genproto v0.0.0-20200413115906-b5235f65be36 // indirect
 	k8s.io/component-base v0.18.0
 )
 
-replace go.etcd.io/etcd => go.etcd.io/etcd v0.5.0-alpha.5.0.20200329194405-dd816f0735f8
+replace (
+	// https://github.com/etcd-io/etcd/issues/11563
```
kakkoyun, line 42: welcome to etcd dependency hell

test/config/tenants.yaml
```diff
@@ -27,10 +27,9 @@ tenants:
   opa:
     url: http://127.0.0.1:8181/v1/data/observatorium/allow
   rateLimits:
-    "/api/metrics/v1/":
+    "/api/metrics/v1/.+/api/v1/receive":
```
squat, line 30: In general, it makes things messy for configuration files to have maps where the keys are special values that users can define. I think it will be cleaner and easier to understand if we use a list:
```yaml
rateLimits:
- path: /api/metrics/......
  limit: 1
  window: 1s
- path: /api/logs/.......
  limit: 100
  window: 1s
```

kakkoyun:
Marking it WIP again to fix dependency problems.

bwplotka (commented):
The size of this PR, makes it impossible to review and even comment anything on this page :cry:

metalmatze:
Two broader things to change before I go into more detail with this PR.

1. Can we leave the flags being parsed in the `main.go` file? We kept it there on purpose. It should be as easy to find as possible as this is the entry point most people look for when beginning to read through the code. :)
2. Maybe we can move these proto files into a more meaningful package. `protoc/google` and `protoc/gubernator` don't mean that much. Personally, I'd say we should just have those (at least the gubernator ones) in the same package as the rest of the middleware files.

kakkoyun:
@metalmatze I have no strong opinions on both of them so I'll handle them

internal/ratelimit/http.go
```diff
@@ -0,0 +1,127 @@
...
+
+type rateLimiter struct {
+	client gubernator.V1Client
+
+	req *gubernator.RateLimitReq
+}
+
+func (l rateLimiter) Handler(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx, cancel := context.WithTimeout(context.Background(), requestTimeout)
```
metalmatze, line 104:
```suggestion
		ctx, cancel := context.WithTimeout(r.Context(), requestTimeout)
```

internal/ratelimit/http.go
```diff
@@ -0,0 +1,127 @@
...
+		defer cancel()
+
+		resp, err := l.client.GetRateLimits(ctx, &gubernator.GetRateLimitsReq{
+			Requests: []*gubernator.RateLimitReq{l.req},
+		})
+		if err != nil {
+			// TODO(kakkoyun): Should we add metric and fail gracefully gracefully?
```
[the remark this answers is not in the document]
metalmatze: I guess an error is only returned if the underlying service has an actual problem or there was some network issue, right? An expected error from the services should be part of the resp payload?!
In that case, I'd say failing hard is the preferred way.

internal/ratelimit/http.go
kakkoyun: Good catch

kakkoyun:
Friendly ping @squat @metalmatze

internal/authentication/oidc.go
```diff
@@ -16,6 +16,8 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"golang.org/x/oauth2"
+
+	"github.com/observatorium/observatorium/internal"
```
bwplotka, line 20: Just thought, not a blocker - it's best to specify some package inside internal. Just internal package sounds like `utils` to me (:

internal/ratelimit/gubernator/gubernator.proto
```diff
@@ -0,0 +1,189 @@
...
+  rpc GetRateLimits (GetRateLimitsReq) returns (GetRateLimitsResp) {
+    option (google.api.http) = {
+      post: "/v1/GetRateLimits"
+      body: "*"
+    };
+  }
+
+
+  // This method is for round trip benchmarking and can be used by
+  // the client to determine connectivity to the server
```
bwplotka, line 39:
```suggestion
  // the client to determine connectivity to the server.
```

internal/ratelimit/gubernator/gubernator.proto
```diff
@@ -0,0 +1,189 @@
...
+  rpc HealthCheck (HealthCheckReq) returns (HealthCheckResp) {
+    option (google.api.http) = {
+      get: "/v1/HealthCheck"
+    };
+  }
+}
+
+// Must specify at least one Request
```
bwplotka, line 47:
```suggestion
// Must specify at least one Request.
```

And below.

internal/ratelimit/gubernator/gubernator.proto
```diff
@@ -0,0 +1,189 @@
...
+message GetRateLimitsReq {
```
bwplotka, line 48:
```suggestion
message GetRateLimitsRequest {
```
I would make it full word and below. Code is generated anyway (See https://developers.google.com/protocol-buffers/docs/style)

bwplotka: Unless you copied this code (if so, let's comment about this)

internal/ratelimit/http.go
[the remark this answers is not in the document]
bwplotka: This is an important decision to make. We are making rate-limiting a hard dependency on request life time. I think similar to auth it should be blocking, but it will be painful if this will become flaky. I hope metric is there with grpc client metric middleware? (:

main.go
```diff
@@ -464,8 +506,8 @@ func main() {
 			Addr:         cfg.server.listen,
 			Handler:      r,
 			TLSConfig:    tlsConfig,
-			ReadTimeout:  readTimeout,  // best set per handler
-			WriteTimeout: writeTimeout, // best set per handler
+			ReadTimeout:  readTimeout,  // best set per handler.
```
bwplotka, line 509: What does this mean?

main.go
```diff
@@ -579,6 +650,8 @@ func parseFlags() (config, error) {
 			" Note that TLS 1.3 ciphersuites are not configurable.")
 	flag.DurationVar(&cfg.tls.reloadInterval, "tls.reload-interval", time.Minute,
 		"The interval at which to watch for TLS certificate changes.")
+	flag.StringVar(&cfg.rateLimiterAddress, "middleware.rate-limiter.grpc-address", "",
+		"The gRPC Server Address against which to run rate limit checks. If it is not specified local non-shared rate limiting will be used.")
```
bwplotka, line 654:
```suggestion
		"The gRPC Server Address against which to run rate limit checks. when rate limits are specified for any tenant. If it is not specified local non-shared rate limiting will be used.")
```

bwplotka: Wonder.. shouldn't this be option for config above? 

```
RateLimits []*struct {
			Endpoint string   `json:"endpoint"`
			Limit    int      `json:"limit"`
			Window   duration `json:"window"`
		} `json:"rateLimits"`
```

test/integration.sh
bwplotka: Is there a way to block it, add some useful information to not forget about this PR comment?

bwplotka (dismissed):
Nice, some comments, otherwise LGTM!

internal/ratelimit/gubernator/gubernator.proto
kakkoyun: This code piece is copied directly from gubernator repo. The script that copies it in Makefile. I will add a README to the directory.

test/integration.sh
metalmatze: That comment is probably also really good to put into the code at this exact place. :blush:

metalmatze:
If only `internal/ratelimit/gubernator/gubernator.proto` is depending on `protoc/google` can't we move those files into the that directory for now to scope it pretty closely? We tried avoiding globals and broader scope in this project so far. :blush:

internal/middleware.go
```diff
@@ -0,0 +1,6 @@
+package internal
+
+import "net/http"
+
+// Middleware is a convenience type for functions that wrap http.Handlers.
+type Middleware func(http.Handler) http.Handler
```
metalmatze, line 6: Rather than moving this around, wh not copy the code into the package you need it? 
This can easily stay in the `internal/authentication` package unmodified. 
Simply create a new one in `internal/ratelimit/gubernator`.

metalmatze (changes requested):

main.go
kakkoyun: What exactly do you mean? Do you want to specify rate limiter endpoint per tenant?

internal/middleware.go
kakkoyun: Our middleware definition is the same for all so I want to share it. What is the value copying the same definition for each package?

main.go
metalmatze: I guess that's a question for the original author and not Kemal? It means that the timeouts are best set specifically per handler, as the handlers probably have different requirements. /query should for example take only fewer seconds compared to /query_range.

metalmatze (approved):
Happy to merge from my side. 
Very nice work! :tada:

PR #107, opened 2020-10-20
internal: Move packages to top level

brancz:
These packages are useful in themselves, so they should be importable.

@observatorium/maintainers


squat:
I really like this idea. In general, I am not a fan of internal packages. If packages are generally useful for the community, then we should allow others to use them and if necessary add a disclaimer saying there is no guaranteed API stability. otherwise I think it becomes an excuse to have untested packages etc

squat (approved):

PR #109, opened 2020-10-20
authentication: Refactor OIDC provider middleware and handler creation

brancz:
@observatorium/maintainers

### DIFF
--- a/authentication/oidc.go
+++ b/authentication/oidc.go
@@ -52,14 +52,18 @@ func NewOIDC(logger log.Logger, configs []TenantOIDCConfig) (http.Handler, map[s
 	warnings := make([]error, 0, len(configs))
 
 	for _, c := range configs {
-		h, m, err := NewProvider(logger, getCookieForTenant(c.Tenant), "/"+c.Tenant, c.OIDCConfig)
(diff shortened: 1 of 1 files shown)


authentication/oidc.go
```diff
@@ -83,10 +87,24 @@ func NewOIDC(logger log.Logger, configs []TenantOIDCConfig) (http.Handler, map[s
...
+	return fmt.Sprintf("observatorium_%s", tenant)
+}
+
+type OIDCProvider struct {
+	logger           log.Logger
+	provider         *oidc.Provider
+	client           *http.Client
+	config           OIDCConfig
+	cookieName       string
+	callbackRedirect string
```
squat, line 100: callback and redirect are usually synonyms in OAuth terminology. We should rather call this either "callbackURL" or "redirectURL"

squat (commented):
overall looks great :))
If we can add godoc comments for the exported methods and types that would be enough for me :)))

brancz:
Done :)

authentication/oidc.go
brancz: Agreed. Renamed to `redirectURL`.

squat (approved):
lgtm

PR #110, opened 2020-10-23
Add serviceCA to jsonnet lib

jpkrohling:
Includes a new configuration property for the API, allowing consumers to specify the name of the config map containing the service-ca.crt, typically created via OpenShift's certificate service.

The first iteration of this PR is mainly my naive suggestion to solve the problem that I'm facing with observatorium/deployments#347.

﻿Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>

### DIFF
--- a/jsonnet/lib/observatorium-api.libsonnet
+++ b/jsonnet/lib/observatorium-api.libsonnet
@@ -212,7 +212,15 @@
                        readOnly: true,
                      },
                    ] else []
-                 ) else []),
+                 ) else [])+
+                (if std.objectHas(api.config, 'serviceCA') then [
+                    {
+                      name: 'service-ca',
+                      mountPath: '/var/run/tls/service-ca.crt',
+                      subPath: 'service-ca.crt',
+                      readOnly: true,
+                    }
+                 ] else []),
             },
           ],
           volumes:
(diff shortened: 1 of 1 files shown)


squat:
Generally we want to decouple configuring a deployment to accept a CA and actually deploying the CA in a configmap. This is because most of the time, CAs (and the like) are generated and provided to us by controllers in configmaps rather than in some binary form that we have to put into a configmap ourselves.
I think a better solution would be expect the configmap with the CA to exist; in the case of the tests in observatoriun/deployments, this would mean we generate it into the tests directory and provision it explicitly.
The Jsonnet here can be modified so that if any tenant in the tenants config (https://github.com/observatorium/observatorium/blob/master/jsonnet/lib/observatorium-api.libsonnet#L23) has hidden fields named `issuerCAConfigMapName` and `issuerCAKey` then we:
* append a volume for `issuerCAConfigMapName` to the pod; and
* append a subpath volume mount for `issuerCAKey` to the API container mounted at the path `issuerCAPath`.

squat:
We do something similar in: https://github.com/observatorium/observatorium/blob/master/jsonnet/lib/observatorium-api.libsonnet#L183-L191

jpkrohling:
I actually saw that, but `mTLS` scared me off, as I thought the server would also validate the client, which doesn't seem to be the case :-) Closing this, as that sounds like the real solution.

PR #114, opened 2020-10-26
Support nodeSelector

clyang82:
Support `nodeSelector` for `observatorium-api` component.

/assign @squat @metalmatze

### DIFF
--- a/jsonnet/lib/observatorium-api.libsonnet
+++ b/jsonnet/lib/observatorium-api.libsonnet
@@ -383,4 +383,21 @@
       },
     },
   },
+
+  withNodeSelector:: {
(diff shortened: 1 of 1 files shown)


kakkoyun (changes requested):
Looking good so far. Thanks a lot.

To actually validate these changes, could you please provide some usage examples? We already have examples. that'd be super awesome to generate another manifest with these specified. So that we can make sure they are validated and examples are provided.

metalmatze:
For now, I want to see where https://github.com/observatorium/operator/pull/21 is going. If there are other reasons to have this in the future, then I'm happy to discuss it in the future again.

PR #118, opened 2020-10-28
Add dex parameter to make run target

blockloop:
make run was failing to find `dex`. This fixes the problem by providing the bingo path to `dex`

```
$ make run
go mod tidy
go mod download
go mod verify
all modules verified
CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on GOPROXY=https://proxy.golang.org go build -a -ldflags '-s -w' -o observatorium .
PATH=$PATH:/home/brett/projects/redhat/observatorium/tmp/bin:/home/brett/.go/bin ./test/run-local.sh
-------------------------------------------
- Waiting for Dex to come up...  -
-------------------------------------------
./test/run-local.sh: line 16: dex: command not found
..................................^C./test/run-local.sh: line 1: kill: (1207933) - No such process
make: *** [Makefile:77: run] Interrupt
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ build: observatorium
 
 .PHONY: run
 run: build $(THANOS) $(DEX) $(LOKI) generate-cert
-	PATH=$$PATH:$(BIN_DIR):$(FIRST_GOPATH)/bin ./test/run-local.sh
+	PATH=$$PATH:$(BIN_DIR):$(FIRST_GOPATH)/bin DEX=$(DEX) ./test/run-local.sh
 
 .PHONY: deps
 deps: go.mod go.sum


periklis (approved):
LGTM

squat:
I don't think this is right. We should be adding the bingo path to the PATH variable so that the binary can be found

blockloop:
> I don't think this is right. We should be adding the bingo path to the PATH variable so that the binary can be found

@squat, I don't think that will work, because the binaries in the bingo path
aren't simply named which is why we use their make target names in the 
Makefile (i.e. `$(DEX)` instead of just `dex`). They include version
information. See [here](https://github.com/observatorium/observatorium/blob/6593067b9c0040abd04782831833d20fbffd157a/.bingo/Variables.mk#L19). Unless I've misunderstood how bingo works which
is quite possible.

squat:
ack, you're totally right. in fact, we do the same thing when running the test-integration target.

PR #120, opened 2020-12-04
Update Gubernator and run unpriviledged

metalmatze:
Mostly to run Gubernator on ports > 1024.


kakkoyun (approved):
Perfect 💯

kakkoyun:
The logs read/write tests are broken because of OPA port clash AFAIT.

metalmatze:
```{"err":"listen tcp :8181: bind: address already in use","level":"fatal","msg":"Listener failed.","time":"2020-12-04T11:39:40Z"}
```
I need to find out what uses that port already and change Gubernator to a different one, I think.

squat:
that's right, the OPA http server is running on 8181: https://github.com/observatorium/observatorium/blob/master/test/config/tenants.yaml#L28

kakkoyun (approved):

PR #126, opened 2021-02-18
Add forwarding upstream authorizer response status codes

periklis:
This PR adds support to forward the upstream authorizers' status codes to observatorium clients (e.g. prometheus remote-write, promtail). In-process authorizers (i.e. static RBAC, in-process OPA) will return as before `StatusForbidden` on any error when processing the authorization request. Upstream authorizers (i.e. OPA-AMS) will forward the status code of the upstream response if non 2xx or else `StatusForbidden` for errors processing the authorization request.

Forwarding status codes enables backpressure signaling to clients (e.g. StatusTooManyRequests) or non-retries on 400s.

/cc @squat

### DIFF
--- a/opa/opa.go
+++ b/opa/opa.go
@@ -58,7 +58,7 @@ type restAuthorizer struct {
 }
 
 // Authorize implements the rbac.Authorizer interface.
-func (a *restAuthorizer) Authorize(subject string, groups []string, permission rbac.Permission, resource, tenant string) bool {
+func (a *restAuthorizer) Authorize(subject string, groups []string, permission rbac.Permission, resource, tenant string) (int, bool) {
 	var i interface{} = Input{
 		Groups:     groups,
 		Permission: permission,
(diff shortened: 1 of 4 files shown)


periklis:
@squat PTAL

periklis:
@kakkoyun Could you take a look please?

opa/opa.go
```diff
@@ -93,30 +95,34 @@ func (a *restAuthorizer) Authorize(subject string, groups []string, permission r
 			"status", res.Status,
 		)
 
-		return false
+		return res.StatusCode, false
```
kakkoyun, line 98: I just wonder if we should take specific actions depending on the class of the errors. For example, would a logline also helpful for debugging purposes when we receive a 500 error? This would surface the configuration errors earlier.

kakkoyun (approved):
LGTM.

Just a nit comment.

opa/opa.go
periklis: @kakkoyun I don't quite get your point here. There is a log line above in L91.

kakkoyun: Sorry I haven't seen that, it was folded 🙈

PR #130, opened 2021-03-16
Return non-200 HTTP response code if concurrent API requests cross the specified concurrent-request-limit #

spaparaju:
With Zero backlog and backlog getting timing out immediately with 1ms, all concurrent requests beyond 'middleware.concurrent-request-limit' result in immediate non-200 HTTP response

### DIFF
--- a/main.go
+++ b/main.go
@@ -119,7 +119,7 @@ type middlewareConfig struct {
 	rateLimiterAddress                string
 	concurrentRequestLimit            int
 	backLogLimitConcurrentRequests    int
-	backLogDurationConcurrentRequests int
+	backLogDurationConcurrentRequests time.Duration
 }
 
 type internalTracingConfig struct {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -369,7 +369,9 @@ func main() {
 		r.Use(middleware.Recoverer)
 		r.Use(middleware.StripSlashes)
 		r.Use(middleware.Timeout(middlewareTimeout)) // best set per handler.
-		r.Use(middleware.Throttle(cfg.middleware.concurrentRequestLimit))
+		// With Zero backlog and backlog getting timing out immediately with 1ms, all concurrent requests beyond cfg.middleware.concurrentRequestLimit result in immediate non-200 HTTP response.
+		backlogDuration, _ := time.ParseDuration("1ms")
```
squat, line 373: Rather than parsing a string, which can return an error, let's use `time.Millisecond` directly

main.go
```diff
@@ -369,7 +369,9 @@ func main() {
 		r.Use(middleware.Recoverer)
 		r.Use(middleware.StripSlashes)
 		r.Use(middleware.Timeout(middlewareTimeout)) // best set per handler.
-		r.Use(middleware.Throttle(cfg.middleware.concurrentRequestLimit))
+		// With zero backlog concurrent requests crossing a rate-limit result in non-200 HTTP response.
+		backlogDuration := 1 * time.Millisecond
```
kakkoyun, line 373: Let's make this configurable. We might need to increase this duration without building a new binary.

kakkoyun (changes requested):
This change makes sense. We should be aware of the possible implications of having a large number of the backlog. As long as we have headroom resource-wise all would be good.

main.go
```diff
@@ -707,6 +711,11 @@ func parseFlags() (config, error) {
 			" If not specified, local, non-shared rate limiting will be used.")
 	flag.IntVar(&cfg.middleware.concurrentRequestLimit, "middleware.concurrent-request-limit", 10_000,
 		"The limit that controls the number of concurrently processed requests across all tenants.")
+	flag.IntVar(&cfg.middleware.backLogLimitConcurrentRequests, "middleware.backlog-limit-concurrent-requests", 0,
+		"The number of concurrent requests that can buffered.")
+	flag.IntVar(&cfg.middleware.backLogDurationConcurrentRequests, "middleware.backlog-duration-concurrent-requests", 1,
```
kakkoyun, line 716:
```suggestion
	flag.DurationVar(&cfg.middleware.backLogDurationConcurrentRequests, "middleware.backlog-duration-concurrent-requests", time.Millisecond,
```

We have a duration type let's use it

kakkoyun (changes requested):
Looking good. Only one thing.

kakkoyun (approved):

Issue #136, opened 2021-04-29
Only one CA certificate is supported in mtls ca file


marcolan018:
Currently when loading certificate from mtls ca file, only the first certificate will be loaded to verify the incoming client certificates. If there are more than one certificates located in the ca file, others will be ignored.(https://github.com/observatorium/api/blob/cbff1da9ba9371516a6254db359ae33545aefdf7/main.go#L239-L251)

It's a very common scenario  that more than one certificates are required. e.g. assume thousands of clients are talking to api gateway now. The CA certificates renewed, there will be a period that some clients will use the certificates signed by new CA cert, but some still using the certificates signed by old CA cert.

squat:
This sounds totally reasonable to me and should be quite straightforward to implement 👍

PR #137, opened 2021-04-30
support multiple certificates in mtls ca file

marcolan018:
This  PR is to fix the issue: https://github.com/observatorium/api/issues/136

It updates CA of MTLSConfig from *x509.Certificate to []*x509.Certificate. When reading ca file, all certificates located in the ca file will be loaded and stored in MTLSConfig.CA. Later, all certificates will be added into certpool to verify the incoming request.

### DIFF
--- a/authentication/mtls.go
+++ b/authentication/mtls.go
@@ -10,7 +10,7 @@ import (
 // MTLSConfig represents the mTLS configuration for a single tenant.
 type MTLSConfig struct {
 	Tenant string
-	CA     []*x509.Certificate
+	CAs    []*x509.Certificate
 }
 
 // NewMTLS creates a set of Middlewares for all specified tenants.
@@ -22,7 +22,7 @@ func NewMTLS(configs []MTLSConfig) map[string]Middleware {
 		middlewares[c.Tenant] = func(next http.Handler) http.Handler {
 			caPool := x509.NewCertPool()
 
-			for _, ca := range c.CA {
+			for _, ca := range c.CAs {
 				caPool.AddCert(ca)
 			}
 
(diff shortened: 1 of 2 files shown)


authentication/mtls.go
```diff
@@ -10,7 +10,7 @@ import (
 // MTLSConfig represents the mTLS configuration for a single tenant.
 type MTLSConfig struct {
 	Tenant string
-	CA     *x509.Certificate
+	CA     []*x509.Certificate
 }
```
squat, line 14: Small nit: can we call this `CAs` to show that it is many certificates?

main.go
```diff
@@ -170,7 +170,7 @@ func main() {
 		MTLS *struct {
 			RawCA  []byte `json:"ca"`
 			CAPath string `json:"caPath"`
-			ca     *x509.Certificate
+			ca     []*x509.Certificate
```
squat, line 173: Same here

squat: Also, i think the CAPath field should be an array to match

squat (commented):
Looks great so far, just don't small details and we can merge it. Good work!

main.go
squat: Once we change the CAPath field to and array called CAPaths, we will need to adjust the integration test configuration here as well: https://github.com/observatorium/api/blob/master/test/config/tenants.yaml#L26

marcolan018: @squat 
I think CAPath should still a single file path, but not multiple?
To support multiple ca certfiicates, the common usage pattern is to put multiple certificates into one ca file, but not provide multiple ca files.

squat: Aha ok that makes a lot of sense. Sgtm

squat (approved):
LGTM

PR #138, opened 2021-05-07
Updated dynamic rules API with RulesRepositoryFile and RulesRepositoryNop

allenmqcymp:
@squat PTAL at rules_file.go and rules_nop.go and changes in main.go


squat:
cc @observatorium/maintainers

api/metrics/v1/rules.go
```diff
@@ -0,0 +1,239 @@
...
+			msg := "failed to list rules"
+			level.Debug(logger).Log("msg", msg, "err", err)
+			http.Error(w, msg, http.StatusInternalServerError)
+			return
+		}
+
+		for i := range rules.Groups {
+			for j := range rules.Groups[i].Rules {
+				if rules.Groups[i].Rules[j].Labels == nil {
+					rules.Groups[i].Rules[j].Labels = make(map[string]string)
```
ianbillett, line 70: Why do we need to do this? To prevent a panic when accessing presumably?

ianbillett (commented):
I'm fear my review here is not going to be super helpful as I'm missing a huge amount of context. The implementation and testing of the proposal looks reasonable ok :man_shrugging: 

Can you include motivations / design docs in the PR description? I'm super curious as to why we need this, and why we are supporting different sources of rules.

This PR makes us a stateful service provider of rules (rules-as-a-service?) to users, and I don't think this PR alone is enough to justify the extra burden that places on us? I'm probably being overly pessimistic :shrug:

api/metrics/v1/rules.go
squat: yes exactly, we're not guaranteed to have a map allocated yet

api/metrics/v1/rules.go
```diff
@@ -0,0 +1,239 @@
...
+	DeleteRule(ctx context.Context, tenant, name string) error
+}
+
+func writeRuleHandler(logger log.Logger, repository RulesWriter, label string) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		tenant, ok := authentication.GetTenant(r.Context())
+		if !ok {
+			const msg = "failed to get tenant"
+			level.Warn(logger).Log("msg", msg)
+			http.Error(w, msg, http.StatusInternalServerError)
```
ianbillett, line 155: is an incorrect tenant really an `http.StatusInternalServerError`? I would have thought `http.StatusBadRequest`?

ianbillett: wait ignore me - that's it failing to parse from the context :facepalm: yeah deffo 5xx

api/metrics/v1/rules_postgres.go
```diff
@@ -0,0 +1,157 @@
...
+	"context"
+	"database/sql"
+
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/pkg/rulefmt"
+	"gopkg.in/yaml.v3"
+)
+
+// Statically check that the Postgres rules repository implements the RulesRepository interface.
+var _ RulesRepository = (*RulesRepositoryPostgres)(nil)
```
ianbillett, line 13: that is a neat trick i did not know about :)

api/metrics/v1/rules_postgres.go
```diff
@@ -0,0 +1,157 @@
...
+
+// RulesRepositoryPostgres the actual implementation of RulesRepository for Postgres.
+type RulesRepositoryPostgres struct {
+	db *sql.DB
+}
+
+func (r *RulesRepositoryPostgres) ListRuleGroups(ctx context.Context, tenant string) (RuleGroups, error) {
+	var groups RuleGroups
+
+	query := `SELECT name, interval, rules FROM metrics_rules WHERE tenant = $1 ORDER BY name ASC`
```
ianbillett, line 29: I'm assuming then that the database has been spun up? The design doc still says "some sort of database - and thus we can for now only persist to disk" so make me think these code paths are unused?

squat: We are totally not in the business of spinning up DBs. If you want to run Observatorium with a rules repository that is rea/write then you need to provide a Postgres DB. All operations will then create/read/update/delete rows in this DB table. This code path _is_ used. The e2e tests that are running in this PR do the following:
1. spin up CockroachDB
2. run a migration to create the necessary table in the DB
3. create/update/read rules from the DB using the API

squat: This PR also enables another read-only rules repository: every tenant can specify their own file that holds the rules for their users. When running like this, users can only read from the file and cannot create, update, or delete rules.

ianbillett: I think we should chat in person about this :) Apologies, but I certainty don't feel like I have enough context to approve this PR, as I do not fully comprehend the use-cases or implications of merging this :raised_hands:

ianbillett:
Chatted with @squat and @bwplotka about this PR earlier today - we will first update & commit the design doc for this PR into the observatorium repo which will capture the long list of things we talked about.

kakkoyun (changes requested):
Looks good! 

One thing I think this feature is missing is to ensure the migrations run against the database. What is our strategy here? Maybe we should provide necessary manifests if we want to achieve this using externally. Or we should provide something in the dockerfile?

PR #140, opened 2021-06-03
authorization: isolate tenants querying metrics

squat:
This commit enforces tenancy on all metrics queries to ensure that
tenant A cannot query for metrics from tenant B. This is done by
repurposing code from prom-label-proxy.


authorization/http.go
```diff
@@ -44,3 +52,90 @@ func WithAuthorizers(authorizers map[string]rbac.Authorizer, permission rbac.Per
...
+			next.ServeHTTP(w, r)
+		})
+	}
+}
+
+const queryParam = "query"
+
+// Adapted from
+// https://github.com/prometheus-community/prom-label-proxy/blob/952266db4e0b8ab66b690501e532eaef33300596/injectproxy/routes.go.
+func enforceQueryValues(e *injectproxy.Enforcer, v url.Values) (values string, noQuery bool, err error) {
```
onprem, line 121:
```suggestion
func enforceQueryValues(e *injectproxy.Enforcer, v url.Values) (values string, foundQuery (or queryPresent) bool, err error) {
```

This confused me at first that we are setting `noQuery` to `false` when there is "no query".

onprem (commented):
A few comments otherwise LGTM. Cool work!

authorization/http.go
squat: this is taken verbatim from the prom-label-proxy, but this is really backwards. Let's fix that :+1:

squat:
Thanks for the review @onprem! I made the suggested changes :+1:

onprem (approved):
LGTM!

PR #141, opened 2021-06-03
Forward the tenant prefix with all requests to upstream

onprem:
With this PR, Obs. API will forward the tenant prefix we strip from the incoming requests (for ex. `/api/metrics/v1/{tenant}/` for metrics API) to the upstream using a HTTP header. The header key is configurable and the default is `X-Forwarded-Prefix`, which is not a standard, but is widely used for the same purpose (e.g. in Traefik).

This forwarding is useful if the upstream applications wants to do some redirection, if they serve a UI, etc. The immediate use case is for serving the Thanos UI properly.

### DIFF
--- a/api/metrics/v1/http.go
+++ b/api/metrics/v1/http.go
@@ -85,7 +85,7 @@ func (n nopInstrumentHandler) NewHandler(labels prometheus.Labels, handler http.
 }
 
 // NewHandler creates the new metrics v1 handler.
-func NewHandler(read, write *url.URL, prefixHeader string, opts ...HandlerOption) http.Handler {
+func NewHandler(read, write *url.URL, opts ...HandlerOption) http.Handler {
 	c := &handlerConfiguration{
 		logger:     log.NewNopLogger(),
 		registry:   prometheus.NewRegistry(),
@@ -103,7 +103,7 @@ func NewHandler(read, write *url.URL, prefixHeader string, opts ...HandlerOption
 		{
 			middlewares := proxy.Middlewares(
 				proxy.MiddlewareSetUpstream(read),
-				proxy.MiddlewareSetPrefixHeader(prefixHeader),
+				proxy.MiddlewareSetPrefixHeader(),
 				proxy.MiddlewareLogger(c.logger),
 				proxy.MiddlewareMetrics(c.registry, prometheus.Labels{"proxy": "metricsv1-read"}),
 			)
@@ -141,7 +141,7 @@ func NewHandler(read, write *url.URL, prefixHeader string, opts ...HandlerOption
 			{
 				middlewares := proxy.Middlewares(
 					proxy.MiddlewareSetUpstream(read),
-					proxy.MiddlewareSetPrefixHeader(prefixHeader),
+					proxy.MiddlewareSetPrefixHeader(),
 					proxy.MiddlewareLogger(c.logger),
 					proxy.MiddlewareMetrics(c.registry, prometheus.Labels{"proxy": "metricsv1-ui"}),
 				)
@@ -163,7 +163,7 @@ func NewHandler(read, write *url.URL, prefixHeader string, opts ...HandlerOption
 		{
 			middlewares := proxy.Middlewares(
 				proxy.MiddlewareSetUpstream(write),
-				proxy.MiddlewareSetPrefixHeader(prefixHeader),
+				proxy.MiddlewareSetPrefixHeader(),
 				proxy.MiddlewareLogger(c.logger),
 				proxy.MiddlewareMetrics(c.registry, prometheus.Labels{"proxy": "metricsv1-write"}),
 			)
--- a/api/metrics/v1/http.go
+++ b/api/metrics/v1/http.go
@@ -85,7 +85,7 @@ func (n nopInstrumentHandler) NewHandler(labels prometheus.Labels, handler http.
 }
 
 // NewHandler creates the new metrics v1 handler.
-func NewHandler(read, write *url.URL, opts ...HandlerOption) http.Handler {
+func NewHandler(read, write *url.URL, prefixHeader string, opts ...HandlerOption) http.Handler {
 	c := &handlerConfiguration{
 		logger:     log.NewNopLogger(),
 		registry:   prometheus.NewRegistry(),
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -791,7 +797,9 @@ func stripTenantPrefix(prefix string, next http.Handler) http.Handler {
 			http.Error(w, "tenant not found", http.StatusInternalServerError)
 			return
 		}
-		http.StripPrefix(path.Join("/", prefix, tenant), next).ServeHTTP(w, r)
+
+		tenantPrefix := path.Join("/", prefix, tenant)
+		http.StripPrefix(tenantPrefix, next).ServeHTTP(w, r.Clone(context.WithValue(r.Context(), proxy.TenantPrefixKey, tenantPrefix)))
```
squat, line 802: Maybe we can reduce the complexity of this one (intense) line by using a `WithPrefix` middleware to set the tenant prefix as the prefix in the context

proxy/proxy.go
```diff
@@ -12,6 +13,10 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+const (
+	TenantPrefixKey = "tenant-prefix"
```
squat, line 17: normally we try to keep the context keys unexported to prevent this from being abused. If we write a `WithPrefix` context setter func in this package then we can make this unexported

onprem: Cool, I was thinking maybe that would be an overkill but combined with the above comment it makes a lot of sense.

proxy/proxy.go
```diff
@@ -30,6 +35,17 @@ func MiddlewareSetUpstream(upstream *url.URL) Middleware {
 	}
 }
 
+func MiddlewareSetPrefixHeader(headerKey string) Middleware {
+	return func(r *http.Request) {
+		prefix, ok := GetTenantPrefix(r.Context())
```
squat, line 40: This really has nothing to do with the tenant, right? It's just a prefix used by the proxy. In fact, the middleware is called `SetPrefixHeader`. Maybe we can make this consistent and remove `Tenant`

main.go
kakkoyun, line 802: Rather than cloning the whole request which is a deep copy AFAIK, why don't we build a new context and use https://golang.org/pkg/net/http/#Request.WithContext to propogate the header?

proxy/proxy.go
```diff
@@ -54,3 +70,10 @@ func MiddlewareMetrics(registry *prometheus.Registry, constLabels prometheus.Lab
 func Logger(logger log.Logger) *stdlog.Logger {
 	return stdlog.New(log.NewStdlibAdapter(level.Warn(logger)), "", stdlog.Lshortfile)
 }
+
+func GetTenantPrefix(ctx context.Context) (string, bool) {
```
kakkoyun, line 74: This method only used here, we don't need to export it.

api/metrics/v1/http.go
```diff
@@ -85,7 +85,7 @@ func (n nopInstrumentHandler) NewHandler(labels prometheus.Labels, handler http.
 }
 
 // NewHandler creates the new metrics v1 handler.
-func NewHandler(read, write *url.URL, opts ...HandlerOption) http.Handler {
+func NewHandler(read, write *url.URL, prefixHeader string, opts ...HandlerOption) http.Handler {
```
squat, line 88: Can we either make this a handler option like the rest? Or pass it directly as a middleware using a `ReadMiddleware` option? This would help us avoid complicating the signature, especially if using the prefix middleware is optional. If it is absolutely required for normal operation of this API then it shouldn't be an option but rather an argument as you made it.

main.go
onprem: Ah, there is one gotcha here though, since the prefix we are stripping is dynamic, we have to extract the `tenant` from the request again, and set the header (which is exactly what we are doing here, hence I kept it the same function).

proxy/proxy.go
```diff
@@ -30,6 +35,17 @@ func MiddlewareSetUpstream(upstream *url.URL) Middleware {
 	}
 }
 
+func MiddlewareSetPrefixHeader(headerKey string) Middleware {
```
squat, line 38: `headerKey` is a bit confusing as it seems to mix `header` and "context key". Let's be consistent with usage elsewhere and call it `header`
```suggestion
func MiddlewareSetPrefixHeader(header string) Middleware {
```

api/metrics/v1/http.go
onprem: The current plan is to pass this prefix for all requests (not just for UI ones) because this can be used for redirection etc in future. do we still want this as optional param?

main.go
```diff
@@ -683,6 +687,8 @@ func parseFlags() (config, error) {
 		"The address on which the internal server listens.")
 	flag.StringVar(&cfg.server.healthcheckURL, "web.healthchecks.url", "http://localhost:8080",
 		"The URL against which to run healthchecks.")
+	flag.StringVar(&cfg.server.prefixHeader, "web.prefix-header", "X-Forwarded-Prefix",
```
kakkoyun, line 690: It's a good idea to make this configurable, however maybe it's a bit of an overkill. We can simplify implementation by sticking to `X-Forwarded-Prefix` and leave the responsibility of modifying it to an external proxy if needed.

main.go
squat: We do this all over the place though by writing the middleware instead as a middleware generator:

```go
func StripPrefixMiddleware(prefix string) func(next http.Handler) http.Handler {
        return func(w http.ResponseWriter, r *http.Request) {
            //something with prefix
        }
    }
}
```

kakkoyun (changes requested):
Looking great :+1: 
I have added a couple of comments and maybe simplification ideas.

main.go
onprem: > To create a new request with a context, use NewRequestWithContext. To change the context of a request, such as an incoming request you want to modify before sending back out, use Request.Clone. Between those two uses, it's rare to need WithContext.

I saw this in godoc and decided to use Clone instead. I am not sure if deep copy is necessary here though?

kakkoyun: Exactly. That's why I'm suggesting to use `Request.WithContext`. `Request.Clone` does the deep clone.

kakkoyun: https://golang.org/pkg/net/http/#Request.Clone
> Clone returns a deep copy of r with its context changed to ctx. The provided ctx must be non-nil.

api/metrics/v1/http.go
onprem: This param is now not required at all, guess that settles it :P

kakkoyun (approved):
:heart:

PR #143, opened 2021-06-16
Do not attempt to enforce tenant label on UI endpoints

onprem:
The `uiProxy` and `proxyRead` in metrics API were using the same set of `readMiddlerwares`. Due to this the Obs. API was trying to enforce tenant label on UI/assets requests as well. This resulted in all UI requests failing with error "query not found".

- This PR introduced new option `UIMiddleware` and uses this instead of the `ReadMiddleware` for UI.
- This PR also fixes the missing label enforcing middleware for the legacy API as well.


squat (approved):
nice work

PR #146, opened 2021-06-25
retrying tenant registration for failed attempts

spaparaju:
This PR covers these changes:

1. refactors tenant registration code to include retries for failed registrations
2. retry attempts modeled with exponential backoff


authentication/http.go
```diff
@@ -39,23 +37,25 @@ func WithTenant(next http.Handler) http.Handler {
 
 // WithTenantID returns a middleware that finds the tenantID using the tenant
 // from the URL parameters and adds it to the request context.
-func WithTenantID(tenantIDs map[string]string) Middleware {
+func WithTenantID(tenantName, tenantID string) Middleware {
```
squat, line 40: We need the tenant map because this maps from the tenant name to the tenant ID. The URL will only have the name as part of the HTTP path

authentication/http.go
```diff
@@ -39,23 +37,25 @@ func WithTenant(next http.Handler) http.Handler {
...
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			tenant := chi.URLParam(r, "tenant")
+			queryValues := r.URL.Query()
```
squat, line 43: The tenant name is part of the path, not the query

main.go
```diff
@@ -616,6 +398,388 @@ func main() {
...
+				Window:  time.Duration(rl.Window),
+			})
+		}
+	}
+
+	return rateLimits
+}
+
+// Create OIDC middleware for a tenant.
+func createOIDCMiddlware(cfg *config, tCfg *tenantsConfig,
```
squat, line 760: This function seems like it belongs in the authentication package

main.go
```diff
@@ -616,6 +398,388 @@ func main() {
...
+	tenantName string, oidcConf authentication.TenantOIDCConfig) (http.Handler, authentication.Middleware, error) {
+	var (
+		oidcHandler          http.Handler
+		oidcTenantMiddleware authentication.Middleware
+		err                  error
+	)
+
+	for i := 1; i < cfg.tenantRegistrationRetryCount; i++ {
```
squat, line 768: I don't think this is the right level of abstraction for this variable. IMO, we should be able to retry the entire tenant registration/addition process. So that would mean wrap the entire `newTenant` func in something that can return an error and be retried. Otherwise this seems like it's too restrictive and will quickly need to be refactored.

main.go
```diff
@@ -667,6 +831,8 @@ func parseFlags() (config, error) {
 		"Path to the RBAC configuration file.")
 	flag.StringVar(&cfg.tenantsConfigPath, "tenants.config", "tenants.yaml",
 		"Path to the tenants file.")
+	flag.IntVar(&cfg.tenantRegistrationRetryCount, "tenants.registration-retry-count", 3,
```
squat, line 834: I don't think we should have a limit here. Instead we should continuously retry registering tenants and use exponential backoff.

authentication/http.go
spaparaju: Now `withTenantID()` is called for each tenant == separate handler() for each tenant

authentication/http.go
spaparaju: Ack. This is tricky. 
Current implementation : {} notation around {tenant} is being used for Chi router to pickup the tenant name from the URL and 'automatically' include as a Query parameter for downstream functions to figure out the tenant name. 
Proposed change: {} notation of Chi is not compatible for separate Mounts() for each tenant. So explicitly adding tenant name as a query parameter to avoid bigger changes in the downstream functions.

main.go
spaparaju: 👍 removed this function from the main package as its adding a layer without

main.go
spaparaju: 👍 I was not sure as well (hence I left the draft in draft mode). Now I have implemented basic exponential backoff implementation.  thanks for the suggestion!

main.go
spaparaju: 👍

authentication/http.go
squat: Got it👍

authentication/http.go
squat: Hmm this sounds funny, so this is changing the API to no longer use path parameters and instead move the tenant parameter to the query?

spaparaju: this change explicitly adds query param ('tenant') which previously automatically added by Chi router with {tenant} at the query path.

authentication/http.go
```diff
@@ -30,7 +28,7 @@ const (
 // WithTenant finds the tenant from the URL parameters and adds it to the request context.
 func WithTenant(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		tenant := chi.URLParam(r, "tenant")
+		tenant := r.URL.Query().Get(tenantKey.String())
```
bwplotka, line 31: Should we fail request if no tenant was provided?

authentication/http.go
bwplotka: > which previously automatically added by Chi router with {tenant} at the query path.

I need to agree with Lucas here, I think it previously was added as `path` not as `query`, no?

See it being used in test: https://github.com/observatorium/api/blob/master/test/integration.sh#L142 - it was used in path. With this new implementation I assume this test would fail?

bwplotka (changes requested):
Thanks for fixing this issue 🤟🏽 

> This PR covers these changes:
> refactors tenant registration code to include retries for failed registrations
> retry attempts modeled with exponential backoff

I am afraid it changes 400 lines more than that - we refactor here the whole main, and auth handlers including removal of chi. For next one I would suggest splitting fixes with refactor for ease of review 🙈 

Can we add integration test (ideally in Go) to check retry logic?

bwplotka:
Also can we fix those log lines? 

image

authentication/http.go
bwplotka: Also this key is weird ...`HTTP context key` as URL param 🤔 

image

authentication/http.go
squat: In any case, this is certainly an API change: we used path parameters before and now use path parameters AND query parameters. I think this is not a good interface as it makes things difficult for clients. Not to mention it's confusing. What happens if you specify the url `/API/V1/foo/graph?tenant=bar`? What tenant is actually used?

spaparaju: The query-param is not expected as part of the call (automatically added by Chi) for picking up the tenant-name (gets added to the context). 
@squat I did not test the possible scenario you have outlined. 
With adding tenant name to the request's context, removed code (prior, changes with this PR) related to adding / reading query-param for tenant-name.

authentication/http.go
spaparaju: removed all the references to add/retrieve tenant names to/from query-params

spaparaju:
> Thanks for fixing this issue 🤟🏽
> 
> > This PR covers these changes:
> > refactors tenant registration code to include retries for failed registrations
> > retry attempts modeled with exponential backoff
> 
> I am afraid it changes 400 lines more than that - we refactor here the whole main, and auth handlers including removal of chi. For next one I would suggest splitting fixes with refactor for ease of review 🙈
> 
> Can we add integration test (ideally in Go) to check retry logic?

Added unit test for retry logic which is based on metrics to track retry attempts.

authentication/oidc.go
```diff
@@ -44,54 +44,45 @@ type OIDCConfig struct {
...
+		if r == nil {
 			level.Debug(logger).Log("msg", msg)
 			http.Error(w, msg, http.StatusUnauthorized)
 			return
 		}
 		h.ServeHTTP(w, r)
 	}))
 
-	return r, middlewares, warnings
+	return h, middleware, nil
```
squat, line 85: The router named `r` seems completely unused. We create it, mount a handler and then never return it. This router was composing the handlers for all of the tenants. Maybe it's not needed anymore given your changes?

authentication/oidc.go
```diff
@@ -44,54 +44,45 @@ type OIDCConfig struct {
...
-		middlewares[c.Tenant] = p.Middleware()
-	}
+	middleware := p.Middleware()
 
 	r := chi.NewRouter()
 	r.Mount("/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		tenant, ok := GetTenant(r.Context())
-		const msg = "error finding tenant"
+		_, ok := GetTenant(r.Context())
+		const msg = "P error finding tenant"
```
squat, line 71:
```suggestion
		const msg = "error finding tenant"
```

authentication/http.go
```diff
@@ -93,31 +81,20 @@ func GetGroups(ctx context.Context) ([]string, bool) {
...
-	middlewares := map[string]Middleware{}
-
-	for _, ms := range middlewareSets {
-		for t, m := range ms {
-			middlewares[t] = m
-		}
-	}
-
+// Creates a single Middleware.
+func WithTenantMiddleware(middleware Middleware) Middleware {
```
squat, line 85: This function does not actually use the Middleware passed to it, so I'm worried that we are missing a chunk of logic that is actually important. Note that previously we would find the right Middleware for the tenant and then call `m(next).ServeHTTP`. To work accordingly, this function should be calling `middleware(next).ServeHTTP`

authentication/http.go
```diff
@@ -93,31 +81,20 @@ func GetGroups(ctx context.Context) ([]string, bool) {
...
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			tenant, ok := GetTenant(r.Context())
+			_, ok := GetTenant(r.Context())
 			if !ok {
 				http.Error(w, "error finding tenant", http.StatusBadRequest)
 				return
 			}
-			m, ok := middlewares[tenant]
-			if !ok {
+			if middleware == nil {
```
squat, line 93: The check and the error message do not really align here. I don't think we really need to check that we were given a non-nil function pointer

authorization/http.go
```diff
@@ -36,14 +36,13 @@ func WithAuthorizers(authorizers map[string]rbac.Authorizer, permission rbac.Per
 			if !ok {
 				groups = []string{}
 			}
-			a, ok := authorizers[tenant]
-			if !ok {
+			if authorizer == nil {
```
squat, line 39: Again, the check and the error message do not align here. I don't think we need to check that the function pointer is not nil. If we did, this world certainly be a programmer error rather than a runtime error and thus we should panic or return an error to the caller rather than an HTTP error to the user. Note that the reason for this check in the first place is that the user could supply a nonexistent tenant name, which would result in a failed lookup. You can't have a failed lookup with your new model.

main.go
```diff
@@ -31,6 +32,13 @@ import (
 	"github.com/metalmatze/signal/healthcheck"
 	"github.com/metalmatze/signal/internalserver"
 	"github.com/metalmatze/signal/server/signalhttp"
+	logsv1 "github.com/observatorium/api/api/logs/v1"
```
squat, line 35: Convention is to keep all of the imports grouped so that all Observatorium/API packages are together at the bottom

authentication/oidc.go
spaparaju: router 'r' is used to direct the traffic to the right tenant (one level of indirection). Yes, 'r' is not used, removed the code to create this extra layer of redirection which is not required if we can deal with handler for the tenant that the traffic is destined to.

authentication/http.go
spaparaju: this code belongs to dead code / a function never used. removed the underlying  function.

authorization/http.go
spaparaju: Makes sense. Moved the check for 'nil' auth. closer to its creation ( newTenant() / main)  and fail the tenant registration if auth turns out to be 'nil'

authentication/http.go
spaparaju: This is part of the codebase is no more required as the specified authN Middleware per tenant is now implemented under newTenant() of main package.

squat: If we are not using it internally any longer then we should definitely delete this function. It's been modified not to use the given middleware so it's not useful for the community either.

authentication/mtls.go
```diff
@@ -13,65 +13,65 @@ type MTLSConfig struct {
...
+// NewMTLS create Middleware for a tenant.
+func NewMTLS(config MTLSConfig) Middleware {
+	middleware := func(next http.Handler) http.Handler {
+		caPool := x509.NewCertPool()
 
-	for _, c := range configs {
-		c := c
-		middlewares[c.Tenant] = func(next http.Handler) http.Handler {
-			caPool := x509.NewCertPool()
+		if len(config.CAs) == 0 {
```
bwplotka, line 21: This is surprising - it will trigger nil panics. I wonder if we should not - either panic `NewMTLS` function since it's a programmatic error potentially or return error if we want to handle that on user side. WDYT?

bwplotka: Also, I wonder - we don't want to take local host trusted CA in default case? 🤔

main.go
```diff
@@ -132,15 +133,64 @@ type internalTracingConfig struct {
...
-	defer level.Info(logger).Log("msg", "exiting")
+	// Load all the command line configuration into tenantsConfig struct.
+	tCfg := tenantsConfig{}
+	loadTenantConfigs(&cfg, &tCfg)
+
+	// Onboard tenants.
+	onboardTenants(cfg, tCfg)
+}
+
+func onboardTenants(cfg config, tCfg tenantsConfig) {
```
bwplotka, line 192: Let's return error and use `Fatalf` only in `main` please - for consistency

main.go
```diff
@@ -569,42 +276,27 @@ func main() {
...
-			internalserver.WithPrometheusRegistry(reg),
-			internalserver.WithPProf(),
-		)
-
-		s := http.Server{
-			Addr:    cfg.server.listenInternal,
-			Handler: h,
-		}
-
+		healthchecks := commonHealthChecks(cfg, tCfg)
```
bwplotka, line 296: possibly worth to inline (no need for single time used variable)

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
 	}
 }
 
+// Return handler for common healthchecks across all the tenants.
```
bwplotka, line 311:
```suggestion
// commonHealthChecks returns handler for common healthchecks across all the tenants.
```

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+			caCert, err := ioutil.ReadFile(cfg.tls.healthchecksServerCAFile)
+			if err != nil {
+				stdlog.Fatalf("failed to initialize healthcheck server TLS CA: %v", err)
+			}
+
+			t.TLSClientConfig.RootCAs = x509.NewCertPool()
+			t.TLSClientConfig.RootCAs.AppendCertsFromPEM(caCert)
+		}
+
+		// checks if server is up
```
bwplotka, line 331:
```suggestion
		// Checks if server is up.
```

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+				http.StatusNotFound,
+				time.Second,
+			),
+		)
+	}
+
+	return &healthchecks
+}
+
+// Apply common middlewares across all the tenants.
```
bwplotka, line 346:
```suggestion
// commonMiddlewares applies common middlewares across all the tenants.
```

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+	r.Use(middleware.Recoverer)
+	r.Use(middleware.StripSlashes)
+	r.Use(middleware.Timeout(middlewareTimeout)) // best set per handler.
+	// With default value of zero backlog concurrent requests crossing a rate-limit result in non-200 HTTP response.
+	r.Use(middleware.ThrottleBacklog(cfg.middleware.concurrentRequestLimit,
+		cfg.middleware.backLogLimitConcurrentRequests, cfg.middleware.backLogDurationConcurrentRequests))
+	r.Use(server.Logger(tCfg.logger))
+}
+
+// Create internal server.
```
bwplotka, line 359:
```suggestion
```
I thinks this is self-explanatory, what about removing this?

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+
+	s := http.Server{
+		Addr:    cfg.server.listenInternal,
+		Handler: h,
+	}
+
+	return &s
+}
+
+// Load RBAC config information from rbac.yaml.
```
bwplotka, line 376: ditto

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+
+	authorizer, err := rbac.Parse(f)
+	if err != nil {
+		return nil, err
+	}
+
+	return authorizer, nil
+}
+
+// Load tenant config information from tenants.yaml.
```
bwplotka, line 393: ditto

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+		tCfg.rateLimitClient = ratelimit.NewClient(tCfg.reg)
+		if err := tCfg.rateLimitClient.Dial(ctx, cfg.middleware.rateLimiterAddress); err != nil {
+			stdlog.Fatal(err)
+		}
+	}
+
+	tCfg.ins = signalhttp.NewHandlerInstrumenter(tCfg.reg, []string{"group", "handler"})
+}
+
+// Create autorizer for a tenant.
```
bwplotka, line 453: to remove?

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+func newAuthorizer(tCfg *tenantsConfig, t *tenant) error {
+	skip := level.Warn(log.With(tCfg.logger, "msg", "skipping invalid tenant"))
+
+	if t.OPA != nil {
```
bwplotka, line 457: Prefer lower indent flow, so we could do `if t.OPA == nil { return nil}`

Also that looks surprising. Are we sure caller does not expect us to fail if we cannot create authorizer?

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+	return nil
+}
+
+// Load MTLS auth information for a tenant.
+func loadMTLSConf(tCfg *tenantsConfig, t *tenant) error {
+	var err error
+
+	skip := level.Warn(log.With(tCfg.logger, "msg", "skipping invalid tenant"))
+
+	if t.MTLS != nil {
```
bwplotka, line 487: ditto

bwplotka: Same as above, I wonder if we can check that oncaller side - much simpler to read + non surprising

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+	return nil
+}
+
+// Load OIDC auth information for a tenant.
+func loadOIDCConf(tCfg *tenantsConfig, t *tenant) error {
+	skip := level.Warn(log.With(tCfg.logger, "msg", "skipping invalid tenant"))
+
+	var err error
+
+	if t.OIDC != nil {
```
bwplotka, line 538: ditto

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+
+			t.OIDC.issuerCA = cert
+		}
+	}
+
+	return nil
+}
+
+// Onboard a tenant.
+func newTenant(cfg *config, tCfg *tenantsConfig, r *chi.Mux, t *tenant) {
```
bwplotka, line 572: Nice! Let's maybe rename to:

```suggestion
func registerTenant(cfg *config, tCfg *tenantsConfig, r *chi.Mux, t *tenant) {
```

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+		// Calculate wait time before rerying a failed tenant registration.
+		waitTime := expBackOffWaitTime(retryCount)
+		time.Sleep(time.Duration(waitTime) * time.Millisecond)
+
+		rateLimits := tenantRateLimits(tCfg, t)
+
+		oidcConf, mTLSConf, err := tennatAuthN(t)
+		if err != nil {
+			level.Error(tCfg.logger).Log("msg", "tenant must specify either an OIDC or an mTLS configuration", t.Name, err)
+			retryCounterMetric.With(prometheus.Labels{"tenant": t.Name}).Inc()
```
bwplotka, line 587: Can we have metric for overall tenants registered?

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+				level.Error(tCfg.logger).Log("msg", "tenant failed to register. retrying ..", t.Name, err)
+				retryCounterMetric.With(prometheus.Labels{"tenant": t.Name}).Inc()
+				retryCount++
+
+				continue
+			}
+		}
+
+		if oidcHandler != nil {
+			r.Mount("/oidc/"+t.Name, oidcHandler)
```
bwplotka, line 612: What if someone registers tenant with the previously used name?

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+		}
+
+		if t.OPA != nil {
+			authZ = t.OPA.authorizer
+		} else {
+			authZ = tCfg.authorizer
+		}
+
+		if authZ == nil {
+			level.Error(tCfg.logger).Log("msg", "invalid authorization for tenant", t.Name)
```
bwplotka, line 627: If that's true, why we added handler already?

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+			tenant, ok := authentication.GetTenant(r.Context())
+			if !ok {
+				w.WriteHeader(http.StatusNotFound)
+
+				return
+			}
+			http.Redirect(w, r, path.Join("/api/metrics/v1/", tenant, "graph"), http.StatusMovedPermanently)
+		})
+
+		// Metrics
```
bwplotka, line 644:
```suggestion
		// Register new metric endpoints.
```

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+			if tCfg.rateLimitClient != nil {
+				r.Use(ratelimit.WithSharedRateLimiter(tCfg.logger, tCfg.rateLimitClient, rateLimits...))
+			} else {
+				r.Use(ratelimit.WithLocalRateLimiter(rateLimits...))
+			}
+
+			r.Mount("/api/v1/"+t.Name, metricsLegacyHandler(cfg, tCfg, authZ))
+			r.Mount("/api/metrics/v1/"+t.Name, stripTenantPrefix("/api/metrics/v1", metricsHandler(cfg, tCfg, authZ)))
+		})
+		// Logs
```
bwplotka, line 659:
```suggestion
		// Register new logging endpoints.
```

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+			r.Group(func(r chi.Router) {
+				r.Use(authentication.WithTenant(t.Name, t.ID))
+				r.Use(authN)
+				r.Use(authentication.WithTenantHeader(cfg.logs.tenantHeader, t.ID))
+				r.Mount("/api/logs/v1/"+t.Name, stripTenantPrefix("/api/logs/v1", logsHandler(cfg, tCfg, authZ)))
+			})
+		}
+
+		level.Info(tCfg.logger).Log("msg", "tenant registration is successful :", t.Name)
+
```
bwplotka, line 670: some counter increment? (: Let's add more observability here since it's now concurrent.

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+			} else {
+				r.Use(ratelimit.WithLocalRateLimiter(rateLimits...))
+			}
+
+			r.Mount("/api/v1/"+t.Name, metricsLegacyHandler(cfg, tCfg, authZ))
+			r.Mount("/api/metrics/v1/"+t.Name, stripTenantPrefix("/api/metrics/v1", metricsHandler(cfg, tCfg, authZ)))
+		})
+		// Logs
+		if cfg.logs.enabled {
+			r.Group(func(r chi.Router) {
```
bwplotka, line 661: I hope chi handler registrations are go-routine safe, can we double check?

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+func registerTenantRetryMetric(tCfg *tenantsConfig) (*prometheus.CounterVec, error) {
+	retryCounterMetric := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "tenant_onboarding_attempts_total",
+			Help: "Number of tenant onboarding attempts.",
+		},
+		[]string{"tenant"},
+	)
+
+	err := tCfg.reg.Register(retryCounterMetric)
```
bwplotka, line 684: I think we overall are not able to handle registration of tenant with the same name so let's maybe remove this check and add general check first think when we do `newTenant` `registerTenant`?

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+	if err != nil {
+		level.Info(tCfg.logger).Log("msg", "duplicate registration of metric", "error", err)
+		return retryCounterMetric, err
+	}
+
+	return retryCounterMetric, nil
+}
+
+// Handler for Metrics legacy API.
```
bwplotka, line 693:
```suggestion
// metricsLegacyHandler creates handler for Metrics legacy API.
```

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+		metricslegacy.Logger(tCfg.logger),
+		metricslegacy.Registry(tCfg.reg),
+		metricslegacy.HandlerInstrumenter(tCfg.ins),
+		metricslegacy.SpanRoutePrefix("/api/v1/{tenant}"),
+		metricslegacy.ReadMiddleware(authorization.WithAuthorizers(authZ, rbac.Read, "metrics")),
+		metricslegacy.ReadMiddleware(authorization.WithEnforceTenantLabel(cfg.metrics.tenantLabel)),
+		metricslegacy.UIMiddleware(authorization.WithAuthorizers(authZ, rbac.Read, "metrics")))
+}
+
+// Handler for Metrics V1 API.
```
bwplotka, line 706:
```suggestion
// metricsHandler creates handler for Metrics V1 API.
```

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+		metricsv1.Registry(tCfg.reg),
+		metricsv1.HandlerInstrumenter(tCfg.ins),
+		metricsv1.SpanRoutePrefix("/api/metrics/v1/{tenant}"),
+		metricsv1.ReadMiddleware(authorization.WithAuthorizers(authZ, rbac.Read, "metrics")),
+		metricsv1.ReadMiddleware(authorization.WithEnforceTenantLabel(cfg.metrics.tenantLabel)),
+		metricsv1.WriteMiddleware(authorization.WithAuthorizers(authZ, rbac.Write, "metrics")),
+	)
+}
+
+// Handler for Logs V1 API.
```
bwplotka, line 721:
```suggestion
// logsHandler creates handler for Logs V1 API.
```

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+		cfg.logs.writeEndpoint,
+		logsv1.Logger(tCfg.logger),
+		logsv1.Registry(tCfg.reg),
+		logsv1.HandlerInstrumenter(tCfg.ins),
+		logsv1.SpanRoutePrefix("/api/logs/v1/{tenant}"),
+		logsv1.ReadMiddleware(authorization.WithAuthorizers(authZ, rbac.Read, "logs")),
+		logsv1.WriteMiddleware(authorization.WithAuthorizers(authZ, rbac.Write, "logs")))
+}
+
+// Calculate exponential backoff wait time for retrying tenant onboarding.
```
bwplotka, line 735:
```suggestion
// expBackOffWaitTime calculate exponential backoff wait time,
```
I don't think it does anything around tenancy, it's pretty generic, right?

main.go
```diff
@@ -616,6 +308,499 @@ func main() {
...
+func expBackOffWaitTime(retryCount int) uint64 {
+	if retryCount == 0 {
+		return 0
+	}
+
+	waitTime := math.Pow(2, float64(retryCount))
+
+	return uint64(waitTime)
+}
```
bwplotka, line 744:
```suggestion
	if retryCount == 0 {
		return 0
	}
	return uint64( math.Pow(2, float64(retryCount)))
}
```

bwplotka (commented):
Nice! Generally LGTM, some nits only

ratelimit/http.go
```diff
@@ -38,58 +38,54 @@ type middleware struct {
...
+			duration: c.Window.Microseconds(),
+		}}.Handler}
 	}
 
-	return combine(middlewares)
+	return combine(mWare)
 }
 
-func combine(middlewares map[string][]middleware) func(next http.Handler) http.Handler {
+func combine(mWare middleware) func(next http.Handler) http.Handler {
```
periklis, line 70: nit `mWare` is a strange naming here, suggest either typical golang `mw` or full name `middleware`.

main_test.go
```diff
@@ -0,0 +1,107 @@
...
+		t.FailNow()
+	}
+
+	cfg.metrics.readEndpoint = u
+
+	u, err = url.ParseRequestURI("http://127.0.0.1:19291")
+
+	if err != nil {
+		t.FailNow()
+	}
```
periklis, line 47: Suggest to skip checking errors for `net/url` functionality and strip down this test case.

main.go
```diff
@@ -132,15 +133,64 @@ type internalTracingConfig struct {
...
 	}
 
-	logger := logger.NewLogger(cfg.logLevel, cfg.logFormat, cfg.debug.name)
-	defer level.Info(logger).Log("msg", "exiting")
+	// Load all the command line configuration into tenantsConfig struct.
+	tCfg := tenantsConfig{}
+	loadTenantConfigs(&cfg, &tCfg)
+
+	// Onboard tenants.
+	onboardTenants(cfg, tCfg)
```
periklis, line 189: The major work of `onboardTenants` is actually `ListenAndServe`, suggest to rename to `Run()` or `ListenAndServeTenants()`

main.go
```diff
@@ -132,15 +133,64 @@ type internalTracingConfig struct {
...
 	cfg, err := parseFlags()
 	if err != nil {
 		stdlog.Fatalf("parse flag: %v", err)
 	}
 
-	logger := logger.NewLogger(cfg.logLevel, cfg.logFormat, cfg.debug.name)
-	defer level.Info(logger).Log("msg", "exiting")
+	// Load all the command line configuration into tenantsConfig struct.
+	tCfg := tenantsConfig{}
+	loadTenantConfigs(&cfg, &tCfg)
```
periklis, line 186: This is a strange pattern reminding on old C days. Prefer to return the full-backed tenants config, e.g.:
```go
tCfg := loadTenantConfig(&cfg)
```

periklis (changes requested):
Awesome work @spaparaju !  This refactoring is clearly needed here. Thank you so much.

Besides a couple of comments to improve the code here and there, I would like to raise a single concern. The monstrous `main()` is now broken into small functions in the same file, which are used once in `loadTenantsConfig`. Why not dedicating a new pkg called `tenants` and structure all this code there? WDYT?

spaparaju:
> Awesome work @spaparaju ! This refactoring is clearly needed here. Thank you so much.
> 
> Besides a couple of comments to improve the code here and there, I would like to raise a single concern. The monstrous `main()` is now broken into small functions in the same file, which are used once in `loadTenantsConfig`. Why not dedicating a new pkg called `tenants` and structure all this code there? WDYT?

makes sense, maybe as a separate PR.

periklis:
> > Awesome work @spaparaju ! This refactoring is clearly needed here. Thank you so much.
> > Besides a couple of comments to improve the code here and there, I would like to raise a single concern. The monstrous `main()` is now broken into small functions in the same file, which are used once in `loadTenantsConfig`. Why not dedicating a new pkg called `tenants` and structure all this code there? WDYT?
> 
> makes sense, maybe as a separate PR.

I mean you are already taking a great effort to split them into functions. Moving them into a separate file shouldn't hurt to be done in this PR.

authentication/mtls.go
bwplotka: ping

bwplotka (changes requested):
Some comments are still left.

ratelimit/http.go
```diff
@@ -38,58 +38,54 @@ type middleware struct {
 
 // WithLocalRateLimiter returns a middleware that controls the amount of requests per tenant using an in-memory store.
 func WithLocalRateLimiter(configs ...Config) Middleware {
-	middlewares := make(map[string][]middleware)
```
matej-g, line 41: FYI these changes seem to break rate limiting, since it looks like only the last config will be applied. Maybe it would be worthwhile to tests for rate limiting as well.

bwplotka:
ping @spaparaju - let's try to get this done this week 🤗

authentication/mtls.go
spaparaju: good catch :) right, this check was not required. removed this check.

main.go
spaparaju: this variable is used as func. parameter (setupInternalServer() )

spaparaju:
> > > Awesome work @spaparaju ! This refactoring is clearly needed here. Thank you so much.
> > > Besides a couple of comments to improve the code here and there, I would like to raise a single concern. The monstrous `main()` is now broken into small functions in the same file, which are used once in `loadTenantsConfig`. Why not dedicating a new pkg called `tenants` and structure all this code there? WDYT?
> > 
> > 
> > makes sense, maybe as a separate PR.
> 
> I mean you are already taking a great effort to split them into functions. Moving them into a separate file shouldn't hurt to be done in this PR.

@periklis gave a try to get the main() as gateway into the new 'tenants' package.

ratelimit/http.go
spaparaju: Ack. As discussed lets add more tests to cover rate limiting functionality, for now tests are passing.

tenants/tenants.go
```diff
@@ -0,0 +1,858 @@
...
+			stdlog.Fatal(err)
+		}
+	}
+
+	tCfg.ins = signalhttp.NewHandlerInstrumenter(tCfg.reg, []string{"group", "handler"})
+
+	return tCfg
+}
+
+// Create autorizer for a tenant.
```
matej-g, line 454:
```suggestion
// Create authorizer for a tenant.
```

tenants/tenants.go
```diff
@@ -0,0 +1,858 @@
...
+	}
+
+	return nil
+}
+
+// Onboard a tenant.
+func newTenant(cfg *config, tCfg *tenantsConfig, r *chi.Mux, t *tenant) {
+	retryCount := 0
+
+	retryCounterMetric, _ := registerTenantRetryMetric(tCfg)
```
matej-g, line 576: Should we be registering this inside the `newTenant`? I'm seeing error `duplicate metrics collector registration attempted` in my logs.

authentication/http.go
```diff
@@ -27,35 +25,25 @@ const (
...
-		))
-	})
-}
-
-// WithTenantID returns a middleware that finds the tenantID using the tenant
-// from the URL parameters and adds it to the request context.
-func WithTenantID(tenantIDs map[string]string) Middleware {
+// WithTenant returns a middleware with request context
+// containing both ID and name of a tenant.
+func WithTenant(tenantName, tenantID string) Middleware {
```
matej-g, line 30: I'm just thinking why we need `tenantName`, shouldn't it be available from route context via `chi.URLParam` as it was before?

tenants/tenants.go
```diff
@@ -0,0 +1,858 @@
...
+				retryCounterMetric.With(prometheus.Labels{"tenant": t.Name}).Inc()
+				retryCount++
+
+				continue
+			}
+		}
+
+		if oidcHandler != nil {
+			r.Mount("/oidc/"+t.Name, oidcHandler)
+		}
```
matej-g, line 614: Could this move to the end of preceding `ìf` block? I guess it would make it more apparent that it belongs to OIDC handler initiation part.

tenants/tenants.go
```diff
@@ -0,0 +1,858 @@
...
+	}
+
+	tCfg.ins = signalhttp.NewHandlerInstrumenter(tCfg.reg, []string{"group", "handler"})
+
+	return tCfg
+}
+
+// Create autorizer for a tenant.
+func newAuthorizer(tCfg *tenantsConfig, t *tenant) error {
```
matej-g, line 455: What do you think about passing just `logger.Logger` instead of whole `*tenantsConfig`? It seem like we do not need whole config inside.

tenants/tenants.go
```diff
@@ -0,0 +1,858 @@
...
+			}
+			t.OPA.authorizer = a
+		}
+	}
+
+	return nil
+}
+
+// Load MTLS auth information for a tenant.
+func loadMTLSConf(tCfg *tenantsConfig, t *tenant) error {
```
matej-g, line 483: Same comment about passing just logger.

tenants/tenants.go
```diff
@@ -0,0 +1,858 @@
...
+				break
+			}
+		}
+	}
+
+	return nil
+}
+
+// Load OIDC auth information for a tenant.
+func loadOIDCConf(tCfg *tenantsConfig, t *tenant) error {
```
matej-g, line 534: Same comment about passing just logger.

tenants/tenants.go
```diff
@@ -0,0 +1,858 @@
...
+		err := errors.New("tenant must specify either an OIDC or an mTLS configuration : " + t.Name)
+
+		return &oidcConf, &mTLSConf, err
+	}
+
+	return &oidcConf, &mTLSConf, nil
+}
+
+// Populate rateLimit configuration for a tenant.
+func tenantRateLimits(tCfg *tenantsConfig, t *tenant) []ratelimit.Config {
```
matej-g, line 785: Same comment about passing just logger.

tenants/tenants_test.go
```diff
@@ -0,0 +1,86 @@
...
+	cfg.tls.reloadInterval = time.Minute
+	cfg.middleware.concurrentRequestLimit = 1000
+	cfg.middleware.backLogLimitConcurrentRequests = 0
+	cfg.logLevel = "debug"
+
+	// command line configuration is now setup, populate tenantsConfig struct from commadline configuration.
+	tCfg := loadTenantConfigs(&cfg)
+
+	// onboard teants in a goroutine and watch retry metrics to go up as the tenant registration fails for
+	// a tenant with wrong configuration at ./test/conf/tenants.yaml.
```
matej-g, line 43: I was wondering whether we really need a wrongly-configured tenant for this test or if we could structure the test differently. Since we're not running Dex as a part of the test, even for 'correctly' configured tenants we would see increase in retry attempts for any tenant, since they won't be able to reach Dex.

matej-g (commented):
This is shaping up nicely after latest changes, good job @spaparaju :+1: I'm leaving some comments.

authentication/http.go
spaparaju: ahh! two reasons:

1. there was quite a bit confusion / concerns on URLParam can be manipulated vs. URLParam is implicitly gets added by 'Chi'
2. the Chi's curly notation ( that Chi uses to populate chi.URLParam automatically from the incoming request) was not working with multiple Chi mounts.

tenants/tenants_test.go
spaparaju: I see that we use Dex as part of CI tests.

main.go
```diff
@@ -1,833 +1,9 @@
...
-	})
-}
-
-type otelErrorHandler struct {
-	logger log.Logger
-}
-
-func (oh otelErrorHandler) Handle(err error) {
-	level.Error(oh.logger).Log("msg", "opentelemetry", "err", err.Error())
+	tenants.Run()
```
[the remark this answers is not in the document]
spaparaju: Make sense. Moved back the parsing logic back to main.

matej-g:
@bwplotka @periklis I have updated the PR and made some final changes, all comments seem to be addressed now, PTAL.

tenants/tenants_test.go
```diff
@@ -0,0 +1,80 @@
...
+		tenantConfig,
+		rbacConfig,
+		"",
+		MetricsConfig{},
+		LogsConfig{},
+		log.NewNopLogger(),
+		reg,
+	)
+
+	// onboard teants in a goroutine and watch retry metrics to go up as the tenant registration fails for
```
bwplotka, line 39:
```suggestion
	// Load teants in a goroutine and watch retry metrics to go up as the tenant registration fails for
```

tenants/tenants_test.go
```diff
@@ -0,0 +1,80 @@
...
+	// a tenant with wrong configuration
```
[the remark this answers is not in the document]
bwplotka: Should we put this above the `Register`?

tenants/tenants.go
```diff
@@ -0,0 +1,588 @@
...
+	Tenants            []*tenant `json:"tenants"`
+	rateLimitClient    *ratelimit.Client
+	authorizer         rbac.Authorizer
+	ins                signalhttp.HandlerInstrumenter
+	reg                *prometheus.Registry
+	retryCounterMetric *prometheus.CounterVec
+	logger             log.Logger
+}
+
+func Register(
```
bwplotka, line 82: I would really propose returning error here and avoid any logging of errors. This is really messy in this function and nested function. Just error passing will be much clear. Right now we have mix of everything here...

tenants/tenants.go
```diff
@@ -0,0 +1,588 @@
...
+	if err := yaml.Unmarshal(f, &tCfg); err != nil {
+		stdlog.Fatalf("unable to read tenant YAML: %v", err)
+	}
+
+	skip := level.Warn(log.With(tCfg.logger, "msg", "skipping invalid tenant"))
+
+	for _, t := range tCfg.Tenants {
+		err = loadOIDCConf(tCfg.logger, t)
+		if err != nil {
+			skip.Log("tenant", t.Name, "err", err)
```
bwplotka, line 151: Are we missing returns?

bwplotka (commented):
Commented on some bits, but this PR is way too big for such small semantic change TBH.  ):

matej-g:
@spaparaju @bwplotka @periklis Hi again. After some more thought, I started to feel like there is too much going on in this PR and it is becoming increasingly hard to consolidate all concerns. We were also hit again by an issue which originated from the OIDC registration failure last week. 

In interest of getting this fix into the codebase as soon as possible, I have created a new, slimmed down PR which focuses only on fixing the issue at hand. I would then suggest to leave the other parts, refactoring etc. to another time, where we could have more thorough discussion and longer exchanges with reviewers.

The PR is now open https://github.com/observatorium/api/pull/162. What are your thoughts? Do you agree to proceed in this way?

PR #148, opened 2021-07-06
Make metrics endpoint as optional

sasagarw:
This PR:
- Makes `metrics endpoint` as optional.
- fmt fixes.

/cc @periklis 
/assign @periklis 

Jira: https://issues.redhat.com/browse/LOG-1512

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,6 +22,9 @@ import (
 	"syscall"
 	"time"
 
+	metricslegacy "github.com/observatorium/api/api/metrics/legacy"
+	metricsv1 "github.com/observatorium/api/api/metrics/v1"
+
 	rbacproxytls "github.com/brancz/kube-rbac-proxy/pkg/tls"
 	"github.com/ghodss/yaml"
 	"github.com/go-chi/chi"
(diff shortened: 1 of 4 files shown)


periklis (commented):
LGTM

main.go
```diff
@@ -22,6 +22,9 @@ import (
 	"syscall"
 	"time"
 
+	metricslegacy "github.com/observatorium/api/api/metrics/legacy"
```
squat, line 25: Please maintain the imports ordered following the golang convention of grouping first standard library imports, then third party imports, and lastly imports from our module.

main.go
```diff
@@ -107,6 +108,8 @@ type metricsConfig struct {
 	writeEndpoint *url.URL
 	tenantHeader  string
 	tenantLabel   string
+	// enable metrics if at least one {read|write}Endpoint} is provided.
+	enabled bool
```
squat, line 112: I think we need to exit with an error of neither logging nor metrics are enabled. If the administrator doesn't specify any endpoints then the API can't do anything. IMO it's the logical extrapolation of the current logic where we said "at least a metrics endpoint is mandatory". Now we would be saying "specifying at least a logging or a metrics endpoint is mandatory"

squat (changes requested):
Two small details, otherwise lgtm :)

squat (approved):
lgtm! thanks @sasagarw

PR #150, opened 2021-07-08
Move metrics tenant isolation to metrics v1 pkg

periklis:
As per `WithEnforceTenantLabel` assuming PromQL specific queries, this function belongs better to `api/metrics/v1` pkg.


squat (approved):
Proposed this change offline; lgtm :+1:

Issue #153, opened 2021-07-13
Integration test for logs tail/write keeps failing silently


matej-g:
It seems that this particular test keeps failing permanently, without an explicit failure message. This applies to running it both locally and on CircleCI (see e.g. recent run here - https://app.circleci.com/pipelines/github/observatorium/api/87/workflows/d8c59c8d-f3c8-4b45-b0f3-50700da87d7a/jobs/357 - but since it exits with code `0`, I assume that is the reason is shows up as green in CI).

Relevant excerpt from output (locally run `make test-integration`):
```
-------------------------------------------
- Logs Read/Write tests: OK               -
-------------------------------------------
-------------------------------------------
- Logs Tail/Write tests                   -
-------------------------------------------
level=debug name=observatorium ts=2021-07-13T10:31:43.501425403Z caller=instrumentation.go:36 request=mgera.remote.csb/JKeZgC35Pg-000080 proto=HTTP/2.0 method=POST status=404 content=application/json path=/api/logs/v1/test-oidc/api/v1/push duration=678.177µs bytes=19
Can't use SSL_get_servername
depth=1 CN = observatorium
verify return:1
depth=0 CN = localhost
verify return:1
level=debug name=observatorium ts=2021-07-13T10:31:43.508231948Z caller=instrumentation.go:36 request=mgera.remote.csb/JKeZgC35Pg-000081 proto=HTTP/1.1 method=GET status=404 content= path=/api/logs/v1/test-oidc/api/v1/tail duration=146.55µs bytes=19
websocat: WebSocketError: Received unexpected status code
websocat: error running
level=info name=observatorium ts=2021-07-13T10:31:43.509220823Z caller=main.go:354 msg="caught interrupt"                                                                                                         
level=info name=observatorium ts=2021-07-13T10:31:43.509247696Z caller=main.go:598 msg="shutting down the HTTP server"                                                                                            
{"level":"info","msg":"Shutting down...","time":"2021-07-13T12:31:43+02:00"}
level=error ts=2021-07-13T10:31:43.509597488Z caller=transfer.go:192 msg="transfer failed" err="cannot find ingester to transfer chunks to: no pending ingesters"                                                 
{"level":"info","msg":"Server shutdown.","time":"2021-07-13T12:31:43+02:00"}
```

periklis:
Addressed by #154

matej-g:
[...]

PR #154, opened 2021-07-14
Remove otel instrumentation on websocket path

periklis:
This PR addresses issue #153. It simply removes otel instrumentation on websocket paths.


squat (approved):
LGTM; I'm curious, why is the OTEL instrumentation breaking the websocket connections?

periklis:
> LGTM; I'm curious, why is the OTEL instrumentation breaking the websocket connections?

OTEL http client transport seems to be broken regard switching protocols via ALPN. TBH I am not sure how much we can buy with instrumenting biderectional websockets. Does OTEL spans somehow work with such channels?

PR #155, opened 2021-07-19
Add support to forward the original bearer token to the authorizer

periklis:
This change set introduces support to forward the original user bearer token from the `observatorium-api` to the `opa-authorizer`. Some authorizer implementations need the original token to talk to the target authorization endpoint, e.g. `opa-openshift` talks to the OpenShift API server to request the users' namespaces.

In detail, the user bearer token pass-through can be configured per tenant and is send via the `X-Forwarded-Access-Token` header.

### DIFF
--- a/authorization/http.go
+++ b/authorization/http.go
@@ -35,7 +35,14 @@ func WithAuthorizers(authorizers map[string]rbac.Authorizer, permission rbac.Per
 				return
 			}
 
-			if statusCode, ok := a.Authorize(subject, groups, permission, resource, tenant); !ok {
+			token, ok := authentication.GetAccessToken(r.Context())
+			if !ok {
+				http.Error(w, "error finding access token", http.StatusUnauthorized)
+
+				return
+			}
+
+			if statusCode, ok := a.Authorize(subject, groups, permission, resource, tenant, token); !ok {
 				w.WriteHeader(statusCode)
 
 				return
--- a/opa/opa.go
+++ b/opa/opa.go
@@ -18,6 +18,11 @@ import (
 	"github.com/observatorium/api/rbac"
 )
 
+const (
+	contentTypeHeader           = "Content-Type"
+	xForwardedAccessTokenHeader = "X-Forwarded-Access-Token" //nolint:gosec
+)
+
 // Input models the data that is used for OPA input documents.
 type Input struct {
 	Groups     []string        `json:"groups"`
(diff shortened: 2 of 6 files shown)


opa/opa.go
```diff
@@ -78,7 +93,26 @@ func (a *restAuthorizer) Authorize(subject string, groups []string, permission r
 		return http.StatusForbidden, false
 	}
 
-	res, err := a.client.Post(a.url.String(), "application/json", bytes.NewBuffer(j))
+	req, err := http.NewRequest(http.MethodPost, a.url.String(), bytes.NewBuffer(j))
+	if err != nil {
+		level.Error(a.logger).Log("msg", "failed to build authorization request", "err", err.Error())
+
+		return http.StatusForbidden, false
```
squat, line 100: This should be an internal server error, not a forbidden error. This is important because forbidden is a 4xx error, meaning clients won't retry requests, but this is clearly a transient error.

opa/opa.go
```diff
@@ -78,7 +93,26 @@ func (a *restAuthorizer) Authorize(subject string, groups []string, permission r
...
+	}
+
+	req.Header.Set(contentTypeHeader, "application/json")
+
+	if a.RequiresAccessToken() {
+		if a.accessToken == "" {
+			level.Error(a.logger).Log("msg", "failed to forward access token to authorization request")
+
+			return http.StatusForbidden, false
```
squat, line 109: This should probably be a bad request and not a forbidden error IMO.

authorization/http.go
```diff
@@ -35,6 +35,15 @@ func WithAuthorizers(authorizers map[string]rbac.Authorizer, permission rbac.Per
...
 			}
 
+			token, ok := authentication.GetAccessToken(r.Context())
+			if !ok {
+				http.Error(w, "error finding access token", http.StatusUnauthorized)
+
+				return
+			}
+
+			a.SetAccessToken(token)
```
squat, line 45: This is _very_ dangerous. The authorizer is used highly concurrently as the API handles many requests in parallel. This line, coupled with the fact that there is no synchronization in setting the token and accessing the authorizer, means that we have no way of knowing who's token is being forwarded for a given request.

Imagine request A comes into the API. This causes the code above to set A's token on the authorizer. 1 millisecond later request B arrives and B's token is set on the authorizer. Next, the authorization middleware makes post requests to the OPA authorizer and sends B's token along for both A and B. This means that A will be authorized according to B's credentials.

squat (changes requested):
This code cannot be merged as it stands right now. It includes a serious security vulnerability as it assumes non-concurrent access to the authorizer data structure. PTAL

matej-g (commented):
Good starting point :+1: I agree with @squat's feedback, adding few comments from my side.

periklis:
@squat @matej-g PTAL, I addressed the review comments. Thank you in advance!

PR #160, opened 2021-08-23
Move E2E tests to efficientgo/e2e

matej-g:
In order to provide a more comfortable and robust experience for out integration testing, this PR moves tests from using Bash to using the [`efficientgo/e2e`](https://github.com/efficientgo/e2e) Go framework. This applies to both metrics and logs test, as well running the setup locally (interactive test). Excludes moving load tests for now.

PR #162, opened 2021-08-24
Introduce OIDC registration retry at startup

matej-g:

This PR is an attempt to introduce a registration retry mechanism for tenants using OIDC for authentication. It has been noticed that if an OIDC provider is not available during the service start-up, the registration will simply fail and tenant will not be instantiated.

In the proposed changes, if an error occurs during the attempt to reach OIDC provider, the registration is retried indefinitely with exponential backoff. This should ensure that once OIDC provider can be reached, the tenant is instantiated properly.

This PR builds on work previously done in https://github.com/observatorium/api/pull/146.


matej-g:
cc @spaparaju (cannot assign you as a reviewer)

main.go
```diff
@@ -456,16 +456,16 @@ func main() {
...
+				oh.AddOIDCForTenant("/oidc/{tenant}", oidc)
 			}
-			r.Mount("/oidc/{tenant}", oidcHandler)
+			r.Mount("/oidc/{tenant}", oh.Router())
 
 			// Metrics
 			if cfg.metrics.enabled {
 				r.Group(func(r chi.Router) {
-					r.Use(authentication.WithTenantMiddlewares(oidcTenantMiddlewares, authentication.NewMTLS(mTLSs)))
+					r.Use(authentication.WithTenantMiddlewares(oh.Middlewares(), authentication.NewMTLS(mTLSs)))
```
squat, line 468: This is only working right now because we are returning the map of middlewares, which is a reference. So even though the `Middlewares` func is only ever called ONCE, the map with the middlewares gets updated to include more tenant middlewares later. However, this hides a dangerous race: maps are not safe to access concurrently.

We need to change the `Middlewares` func to return a concurrent-safe object that can be accesses at the same time someone else who is locking the map is writing to it. This way we would normally do this would be to construct a new map that is a copy of the original `middlewares` map. However, because the `Middlewares` func is only called once EVER, it won't get updated.

authentication/oidc.go
```diff
@@ -45,71 +47,112 @@ type OIDCConfig struct {
...
+		}
+	}()
+}
 
+// Middlewares returns a collction of OIDC middlewares for all registered tenants.
+func (oh *OIDCHandlers) Middlewares() map[string]Middleware {
+	oh.mtx.RLock()
+	defer oh.mtx.RUnlock()
+
+	return oh.middlewares
```
squat, line 129: As written below, this is not thread-safe. This lock is a little bit disingenuous: it locks for access within the OIDCHandlers data structure, but then returns a reference to the map for the caller to access at will without any synchronization with the lock inside the OIDCHandlers datastructure, opening the door to races and panics.

squat (changes requested):
As written in the comments, this implementation "works" but is racy and can result in panics because of concurrent access to the middlewares map

authentication/oidc.go
```diff
@@ -45,71 +47,112 @@ type OIDCConfig struct {
...
 
-		handlers[c.Tenant] = r
-		middlewares[c.Tenant] = p.Middleware()
-	}
+			oh.mtx.Lock()
+			oh.handlers[config.Tenant] = r
+			oh.middlewares[config.Tenant] = p.Middleware()
+			oh.mtx.Unlock()
+
+			level.Debug(oh.logger).Log("msg", "OIDC provider instantiated for tenant", "tenant", config.Tenant)
```
ianbillett, line 118: nit: I would make this `info` level - it should only happen at container start-up once

authentication/oidc.go
```diff
@@ -45,71 +47,112 @@ type OIDCConfig struct {
...
+			p, err := NewProvider(ctx, oh.logger, getCookieForTenant(config.Tenant), "/"+config.Tenant, config.OIDCConfig)
+			if err != nil {
+				level.Warn(oh.logger).Log("msg", "failed to instantiate OIDC provider for tenant", "tenant", config.Tenant, "error", err)
+				oh.retryCountIncFn()
+				b.Wait()
+				continue
+			}
 
-		r := chi.NewRouter()
+			r := chi.NewRouter()
```
ianbillett, line 103: nit: There are things here that probably _should_ be instantiated outside of the for loop?

ianbillett (commented):
Generally LGTM - the only question I have is about how you have tested this change.

If the provider was not available initially, have we seen it try to retry and increment the metrics? Then see it pass when the provider becomes available?

authentication/oidc.go
```diff
@@ -45,71 +47,112 @@ type OIDCConfig struct {
...
+
+// NewOIDCHandlers instantiates OIDC handlers.
+func NewOIDCHandlers(l log.Logger, reg prometheus.Registerer) *OIDCHandlers {
 	tenantErrors := prometheus.NewGauge(prometheus.GaugeOpts{
 		Namespace: "observatorium",
 		Subsystem: "api",
 		Name:      "tenant_registration_failures",
-		Help:      "The number of tenants for which OIDC provider instantiation failed.",
+		Help:      "The number of attempts for which tenant OIDC provider instantiation failed.",
 	})
```
onprem, line 69: With the retry logic, I think this metric is not very usable. WDYT about replacing it with a gauge that shows the current number of tenants undergoing retries (instead of the total number of attempts)? This way, we can alert if there is a tenant failing to instantiate for a long time, etc.

onprem (approved):
This looks great, awesome work. Just a small suggestion about instrumentation.

authentication/oidc.go
onprem: With retries, it will log for every retry (every 5 sec in the best case), considering this, `debug` looks better to me.

matej-g: Yes, that makes sense actually, it is not too spammy and useful.

authentication/oidc.go
matej-g: Which ones do you mean? Since the loop will progress to this section only once, I think it should be fine even if it is instantiated here. From the point of code clarity, I'm impartial whether it is more 'reasonable' to do it within or outside.

authentication/oidc.go
matej-g: Makes much more sense :+1: Will adjust.

matej-g:
> Generally LGTM - the only question I have is about how you have tested this change.

@ianbillett I did some manual testing locally, trying different scenarios with/without the OIDC provider running. I also adjusted the integration tests so that the OIDC provider starts after we initialize the API - this way the re-registration _has_ to succeed, otherwise the metrics tests themselves would be failing.

Ultimately, my plan is to introduce a more robust test after https://github.com/observatorium/api/pull/160 lands, which will allow me to create more complex scenarios with the OIDC provider.

matej-g:
@squat Thanks for the invaluable feedback, I tried a different approach to make the access safe, let me know if that's you had in mind!

onprem (dismissed):
LGTM :100:

authentication/http.go
```diff
@@ -93,31 +93,31 @@ func GetGroups(ctx context.Context) ([]string, bool) {
 	return groups, ok
 }
 
+// MiddlewareFn is function type able to return
+// authentication middleware for given tenant. If
```
squat, line 97: Nit: ```suggestion
// authentication middleware for a given tenant. If
```

File: authentication/http.go [hunk 7]
```diff
@@ -93,31 +93,31 @@ func GetGroups(ctx context.Context) ([]string, bool) {
...
+// no middleware is found, second parameter should
```
squat, line 98:
```suggestion
// no middleware is found, the second return value should
```

authentication/oidc.go
```diff
@@ -45,71 +47,127 @@ type OIDCConfig struct {
...
+	}()
+}
+
+// GetTenantMiddleware an OIDC middleware for the specified tenant, if found.
+func (oh *OIDCHandlers) GetTenantMiddleware(tenant string) (Middleware, bool) {
+	oh.mtx.RLock()
+	mw, ok := oh.middlewares[tenant]
+	oh.mtx.RUnlock()
+
+	if !ok {
```
squat, line 140: We can simplify these lässt lines and always do `return mw, ok` no?

squat (commented):
Looks much better, this change is just what I had in mind. I just have a couple of small comments and everything else looks good to me

authentication/http.go
```diff
@@ -93,31 +93,31 @@ func GetGroups(ctx context.Context) ([]string, bool) {
 	return groups, ok
 }
 
+// MiddlewareFn is function type able to return
```
bwplotka, line 96:
```suggestion
// MiddlewareFunc is function type able to return
```

Not a blocker, but this would follow std lib things like e.g `http.HandlerFunc`

authentication/http.go
bwplotka: I would love if signature and naming would indicate that this returns only auth middleware. Right now it looks pretty generic

authentication/http.go
```diff
@@ -93,31 +93,31 @@ func GetGroups(ctx context.Context) ([]string, bool) {
...
+
+			for _, mwFn := range mwFns {
+				if m, ok := mwFn(tenant); ok {
+					m(next).ServeHTTP(w, r)
+					return
+				}
 			}
 
-			m(next).ServeHTTP(w, r)
+			http.Error(w, "error finding tenant", http.StatusUnauthorized)
```
bwplotka, line 120:
```suggestion
			http.Error(w, "tenant not found, have you registered it?", http.StatusUnauthorized)
```

authentication/oidc.go
```diff
@@ -45,71 +47,127 @@ type OIDCConfig struct {
...
+}
+
+// NewOIDCHandlers instantiates OIDC handlers.
+func NewOIDCHandlers(l log.Logger, reg prometheus.Registerer) *OIDCHandlers {
+	tenantsFailing := prometheus.NewGauge(prometheus.GaugeOpts{
 		Namespace: "observatorium",
 		Subsystem: "api",
-		Name:      "tenant_registration_failures",
-		Help:      "The number of tenants for which OIDC provider instantiation failed.",
+		Name:      "tenants_failing_registrations",
```
bwplotka, line 67: 👍🏽

authentication/oidc.go
```diff
@@ -45,71 +47,127 @@ type OIDCConfig struct {
...
+	const (
+		loginRoute    = "/login"
+		callbackRoute = "/callback"
+	)
+
+	ctx := context.Background()
+	b := backoff.New(ctx, backoff.Config{
+		Min:        500 * time.Millisecond,
+		Max:        5 * time.Second,
+		MaxRetries: 0, // retry indefinitely
```
bwplotka, line 95:
```suggestion
		MaxRetries: 0, // Retry indefinitely.
```

authentication/oidc.go
```diff
@@ -45,71 +47,127 @@ type OIDCConfig struct {
...
+	go func() {
+		var isFailing bool
+		for b.Reset(); b.Ongoing(); {
+			p, err := NewProvider(ctx, oh.logger, getCookieForTenant(config.Tenant), "/"+config.Tenant, config.OIDCConfig)
+			if err != nil {
+				level.Warn(oh.logger).Log("msg", "failed to instantiate OIDC provider for tenant", "tenant", config.Tenant, "error", err)
 
-		r := chi.NewRouter()
+				if !isFailing {
+					oh.retryGauge.Inc()
```
bwplotka, line 106: I would argue that retry counter could be better here. Few reasons:

* Overall prefer counters over gauges. Counters are useful when service is restarting/crashing. Gauges not.
* We need to maintain extra state using `isFailing`, making code bit more complex
*  We have similar info, if not better -> how many retries we did. Becuase what if gauge is 1 but we got stuck on second retry forever because of sum bug, and we never retry. We don't have that info with gague.

authentication/oidc.go
bwplotka: I think there some reasons why it can be more readable if we make loop more concise and smaller, I was surprised too here, same as @ianbillett . Reasons:

* Two engineers were already surprised by this part 🙈 
* If you do it in loop I expect this code to use loop functionality, as is - that we can go back to retries, but looks this large part, does cannot. Focused scoped complexity is the key
* Less indent for complex code = better.

authentication/oidc.go
```diff
@@ -45,71 +47,127 @@ type OIDCConfig struct {
...
+// the tenant OIDC configuration. If a connection with the OIDC provider
+// cannot be established, the method will retry to connect with an increasing
+// backoff.
+func (oh *OIDCHandlers) AddOIDCForTenant(prefix string, config TenantOIDCConfig) {
+	const (
+		loginRoute    = "/login"
+		callbackRoute = "/callback"
+	)
+
+	ctx := context.Background()
```
bwplotka, line 91: Since you spawn go routines I would ensure:
* We get this context hooked to interrupt handler (so when we do ctrl+c we clean go routine nicely, not aborting in random moment) - good practice
* Maybe TODO, but if we want to put this handler on HTTP then what we do if I add OIDC tenant `abc` and it's retrying in loop and  another client adds `abc` and 100 more? Do we need need to handle this somehow?

bwplotka (approved):
Sweet, some detailed comments for learning purposes. Overall amazing, focused job 👍🏽 

Test would be nice, but we really need to iterate here as we can get same incident anytime now, so ok to merge it without extra stuff.

authentication/oidc.go
matej-g: Ok now I'm convinced :upside_down_face:

authentication/oidc.go
matej-g: I see your points, as originally I used counter, but I really thought @onprem's idea with gauge was neater. There is some marginal complexity increase in keeping the state, but conversely to your points, gauge enables us to see how many tenants are failing and for how long, which I think is the crucial data point here. Recording just the number of retries won't directly tell us how widespread the issue is.

@onprem any thoughts?

authentication/oidc.go
matej-g: For your second point, do you mean if there are duplicate tenants trying to register or if there is a large number of tenants running the registration loop at the same time?

authentication/http.go
matej-g: Do you think it is not clear enough? Since when exported, it would be used as `authentication.Middleware`, making it clear this is auth middleware. Naming it e.g. `AuthMiddleware` would stutter a bit.

authentication/oidc.go
matej-g: Totally :sweat_smile:

authentication/http.go
squat, line 96:
```suggestion
// MiddlewareFn is a function type able to return
```

authentication/oidc.go
matej-g: We briefly touched upon this on the community call, it seems the general agreement in the end was to keep this simple for now, whether we do gauge or counter, and see how our needs evolve in the future, e.g. with more tenants.

authentication/oidc.go
bwplotka: Duplicate tenants

authentication/oidc.go
bwplotka: > us to see how many tenants are failing and for how long, which I think is the crucial data point here. Recording just the number of retries won't directly tell us how widespread the issue is.

You can do same with counter, no? (: But if on the community call we are happy with counter, let's go 💪🏽

authentication/oidc.go
matej-g: Hm I guess then we should have a means of detecting duplicate tenants in config, possibly before we proceed to further steps. If that's fine I'd leave it as a TODO, but I agree this should be taken care of.

bwplotka: 👍🏽

authentication/oidc.go
matej-g: I guess we were happy with either :sweat_smile: Let's keep it and come back to this if we feel it needs improving in the future.

bwplotka (approved):
I think there were still 2 unadressed comments, but happy to merge even in this state and get it to prod 💪🏽 Thanks!

authentication/http.go
```diff
@@ -93,31 +93,32 @@ func GetGroups(ctx context.Context) ([]string, bool) {
...
+type Middleware func(http.Handler) http.Handler
 
-	for _, ms := range middlewareSets {
-		for t, m := range ms {
-			middlewares[t] = m
-		}
-	}
+// MiddlewareFunc is a function type able to return authentication middleware for
+// a given tenant. If no middleware is found, the second return value should be false.
+type MiddlewareFunc func(tenant string) (Middleware, bool)
```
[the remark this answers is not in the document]
matej-g: I think was addressed, just marked as resolved by @squat  https://github.com/observatorium/api/pull/162#discussion_r695651162

authentication/oidc.go
matej-g: I did move the new router instantiation, as far as I could see that was actually the only thing being instantiated in the loop that could be done outside :slightly_smiling_face: .

PR #169, opened 2021-09-13
Add newUpRun to TestInteractiveSetup

jessicalins:
This adds the `newUpRun` function call to `TestInteractiveSetup` (as we do in https://github.com/observatorium/api/blob/main/test/e2e/metrics_test.go#L40) which makes it possible to get some metrics written to the receiver
Also a quick fix for the `token` var (we were fetching as the first return from `startBaseServices` instead of the second)


periklis (approved):
LGTM

PR #171, opened 2021-09-23
Pass Tenant ID to Authorizers

tareqmamari:
Some authorizers, in particular, OPA based authorizers relies on
Tenant ID, which coule be a crn like ID that holds information used in the
authorization process.


squat (changes requested):
No big objections from me. One thing however, to follow go convention, all letters in an abbreviation should have the same casing, ie all `tenantId` should become `tenantID`

tareqmamari:
@squat I just fixed that, thanks for the feedback.

tareqmamari:
@squat I might need to update something in the PR.

tareqmamari:
@squat the e2e test is passing locally, it looks like there is an intermittent issue in the pipeline. Could you rerun the test-e2e job please?

periklis:
> @squat the e2e test is passing locally, it looks like there is an intermittent issue in the pipeline. Could you rerun the test-e2e job please?

You can always re-run the tests yourself if you go a `git commit --amend` and then force push in to your fork's branch.

tareqmamari:
@squat this PR should be ready now, anything else is needed?

authentication/http.go
```diff
@@ -72,7 +72,10 @@ func WithTenantHeader(header string, tenantIDs map[string]string) Middleware {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			tenant := chi.URLParam(r, "tenant")
-			r.Header.Add(header, tenantIDs[tenant])
+			// tenantID will be used by loki, loki then splits it to get the chunk parts
+			// based on slashes: https://git.io/JzR97
+			tenantID := strings.ReplaceAll(tenantIDs[tenant], "/", "")
```
squat, line 77: I don't quite understand the need for this change. This seems pretty backend-specific. Can we put this somewhere else so that it doesn't cause unexpected problems?

authorization/http.go
```diff
@@ -42,7 +42,9 @@ func WithAuthorizers(authorizers map[string]rbac.Authorizer, permission rbac.Per
 				return
 			}
 
-			if statusCode, ok := a.Authorize(subject, groups, permission, resource, tenant, token); !ok {
+			tenantID, _ := authentication.GetTenantID(r.Context())
```
squat, line 45: In all of the context value getters above we handle the `!ok` case and return an error to the user. Can we do the same here?

squat (changes requested):
Just a couple small questions left @tareqmamari :)
Thanks for the work :+1:

authentication/http.go
tareqmamari: @squat It is only for the loki backend, and not sure where else we can place it? any suggestions?

squat: hmm this is tricky because it will affect all backends

tareqmamari: @squat what is about adding a new `WithLokiTenantHeader`, and keep the original WithTenantHeader as it is ?

tareqmamari: @squat I added another middleware `WithLogsTenantHeader`, basically with same logic, except the original WithTenantHeader doesn't have that slash replacement. This is not the best approach that I wanted to use, but should do the trick.

squat: Sounds ok to me. Let's cc @periklis to make sure this looks good from the rest of the logging folks

periklis: @tareqmamari Can you provide an example with slashed here? I don't quite get why we should have slashes in the tenantIDs slice in the first place. This is usually a UUID list.

tareqmamari: @periklis since we don't have restriction on the tenant id and Loki for example also accepts other forms of IDs, we want to use a crn-like identifiers, these are used - besides being identifiers - also to hold important information like where that tenant belongs ( e.g. account or group ), alongside other important information, the following is an example of how that might look like
`crn:v1:cloud:public:observatorium:global:a/59bcbfa6ea2f006b4ed7094c1a08dcdd:1a0ec336-f391-4091-a6fb-5e084a4c56f4::`

periklis: IC, but this looks now pretty custom for a single use case here. This simple replace of slashes will certainly raise questions in a couple of months again, on why we need it. This is not properly fitting here as a general case. OTH using simple UUIDs is a pretty standard thing to do. It is well-defined and does not require extra handling with upstream loki handlers. I suggest to reconsider it the ID format you use is a good choice, because removing the slash makes the ID not easy to retrieve the original form. Thus why using this format in the first place?

tareqmamari: @periklis I agree, therefore, I rolled back that part.

periklis (approved):
LGTM

PR #173, opened 2021-10-04
Introduce Plugable Authenticators

tareqmamari:
currently, only OIDC or mTLS based authentications are possible.
However, there are cases where custom authenticators are needed, such
as authenticators that might need to use private go modules, therefore,
this commit refactors the current implementation of the authentication
logic to make it pluggable in order to support new authenticators.


tareqmamari:
FYI: @periklis

tareqmamari:
@observatorium/maintainers please let me know if there is anything else we need to do in this PR to merge it.

periklis:
@tareqmamari Thanks for the latest fixes. Now it works like a charm again.

authentication/authentication.go
```diff
@@ -0,0 +1,145 @@
...
+// provider.
+type Provider interface {
+	Middleware() Middleware
+	Handler() (string, http.Handler)
+}
+
+type tenantHandlers map[string]http.Handler
+
+// AuthenticatorsHandlers manages all middleware and handlers of all authenticators.
+type AuthenticatorsHandlers struct {
```
squat, line 48: Naming is hard. My only remaining nit with this PR is that this type doesn't mean anything to me and is somewhat confusing.

The methods exposed by this type are `NewTenantAuthenticator`, `AuthenticatorMiddlewares`, and `PatternRoutes`. 
I think that `NewTenantAuthenticator` should be renamed, because it's not a constructor that returns authenticators (that is what the AuthenticatorFactory interface does) it rather configures an authenticator for a tenant. It's also somewhat confusing that a `AuthenticatorsHandlers` shoud know how to create a `NewTenantAuthenticator`, as from its name it seems to know more about handlers instead.

Morally, this type serves the purpose of maintaining a mapping of tenants <-> authenticators and allows us to get metadata about the authenticator for a certain tenant, i.e. middleware.

Maybe we can instead call this a `ProviderRepository`.

`NewTenantAuthenticator` could instead be called `ConfigureProvider`?

authentication/authentication.go
```diff
@@ -0,0 +1,145 @@
...
+func onboardNewProvider(providerType string, factory AuthenticatorFactory) {
+	authenticatorsMtx.Lock()
+	defer authenticatorsMtx.Unlock()
+
+	authenticatorFactories[providerType] = factory
+}
+
+// AuthenticatorFactory represents each authenticator factory, a function to create a new corresponding
+// authentication provider.
+type AuthenticatorFactory func(config map[string]interface{}, tenant string, registrationRetryCount *prometheus.CounterVec, logger log.Logger) (Provider, error)
```
squat, line 36: Nit: since this is a function that returns new `Providers`, maybe we can instead call this a `ProviderFactory`? It should be pretty clear what this provider is for since it's part of the `authentication` package. I think this naming would be more consistent and would reduce the cognitive overload of the difference between an `authenticator` and a `provider`, when there really is none.
```suggestion
type ProviderFactory func(config map[string]interface{}, tenant string, registrationRetryCount *prometheus.CounterVec, logger log.Logger) (Provider, error)
```

tareqmamari: Authenticator and Provider are used interchangeably, basically they are the same. As you mentioned, since it is under the authentication package, it would be more clear to use provider explicitly. I will update that.

authentication/authentication.go
tareqmamari: hmm naming is hard, I would vote to use InitializeProvider instead, that at least used before with OIDC. For the AuthenticatorsHandlers, we could either call ProviderRepository or maybe ProviderManager? What do you think?
@squat @periklis

squat: I like `ProviderManager` :+1:

tareqmamari:
@squat @periklis thanks for all of your comments, is there anything else needed in this PR to get it squashed and merged please?

squat (approved):
Thanks for your patience @tareqmamari
LGTM

PR #174, opened 2021-10-04
metrics and logs should be disabled by default

tareqmamari:
this is a bug fix, iniitially, metrics and logs should be disabled,
enable them iff at least one endpoint of each is set.

### DIFF
--- a/main.go
+++ b/main.go
@@ -726,7 +726,16 @@ func parseFlags() (config, error) {
 		rawTracingEndpointType  string
 	)
 
-	cfg := config{}
+	// if at least one metrics/logs endpoint (read or write) is set
+	// metrics/logs will be enabled.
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -726,7 +726,16 @@ func parseFlags() (config, error) {
...
+	// if at least one metrics/logs endpoint (read or write) is set
+	// metrics/logs will be enabled.
+	cfg := config{
+		metrics: metricsConfig{
+			enabled: false,
+		},
+		logs: logsConfig{
+			enabled: false,
+		},
+	}
```
periklis, line 738: Why is that manual defaulting needed? Golang's type system takes care of this.

tareqmamari: Totally right, closing it.

PR #178, opened 2021-10-14
Add Rules API specification and integration to Observatorium API

jessicalins:
Ticket: https://issues.redhat.com/browse/MON-1965

This PR contains:
* The OpenAPI Spec for the Rules API
* The client and server code generated from the OpenAPI spec
* The integration of the Rules API into the main Observatorium API

An e2e test is missing, will be added in a separate PR.

TODO

- [x] Add alerting rule to spec

Signed-off-by: Jéssica Lins <jlins@redhat.com>


rules/spec.yaml
```diff
@@ -0,0 +1,98 @@
...
+  version: "1.0.0"
+  title: Rules Backend API
+  license:
+    name: Apache 2.0
+    url: 'http://www.apache.org/licenses/LICENSE-2.0.html'
+tags:
+  - name: rulesv1
+    description: Calls related to rules
+paths:
+  /api/v1/{tenant}/rules:
```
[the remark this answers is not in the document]
onprem: For now I went with the approach to only have a single file just for the sake of simplicity. Although the Rules backend is storing the files in object storage in a structure that'll allow us to move to multiple files per tenant without breaking anything.

Also, if we allow multiple files, I think it'd be great to have one endpoint that returns all rules merged in a single response. This will make the rule syncer's work much easier.

rules/server/server.go
```diff
@@ -0,0 +1,129 @@
...
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/deepmap/oapi-codegen/pkg/runtime"
+	"github.com/go-chi/chi/v5"
+)
+
+// ServerInterface represents all server handlers.
+type ServerInterface interface {
```
bwplotka, line 15: I wished someday we have gRPC defined APIs that by default expose  HTTP route too (like https://github.com/twitchtv/twirp) 😢

rules/spec.yaml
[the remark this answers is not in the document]
bwplotka: Sorry, I don't get that. The `$ref: '#/components/schemas/Rules'`  points that Rules are set of Groups, right? there is not file notion.

bwplotka (commented):
Nice, some extra input (:

rules/spec.yaml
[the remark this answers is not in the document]
ianbillett: Ah ok - I understand this now. The identifier for the set of rules is the rule group name - tenants can choose to create and update any they like while keeping some level of semantic separation if they wish,

ianbillett (dismissed):

rules/spec.yaml
```diff
@@ -0,0 +1,98 @@
+openapi: 3.0.0
+info:
+  description: The Rules Backend API
+  version: "1.0.0"
```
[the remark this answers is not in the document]
jessicalins: changed it :)

rules/spec.yaml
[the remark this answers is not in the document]
jessicalins: I'm actually a bit confused - as I understand this still means one file per request, but using the `Rules` schema, that uses `RuleGroups`? is this what we want?
We can also test here: https://app.swaggerhub.com/apis/observatorium/rules-backend/1.0.0#/Rules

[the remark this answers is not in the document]
jessicalins: and here where we define yaml as content-type and the use of the `Rules` schema: https://github.com/observatorium/api/blob/44d2f6c58aebab68636f524b7ffdd5debcc304bd/rules/spec.yaml#L48-L53

rules/server/server.go
```diff
@@ -0,0 +1,129 @@
...
+// ServerInterfaceWrapper converts contexts to parameters.
+type ServerInterfaceWrapper struct {
+	Handler            ServerInterface
+	HandlerMiddlewares []MiddlewareFunc
+}
+
+type MiddlewareFunc func(http.HandlerFunc) http.HandlerFunc
+
+// ListRules operation middleware
+func (siw *ServerInterfaceWrapper) ListRules(w http.ResponseWriter, r *http.Request) {
```
squat, line 33: There's something funny going on here. One of the nice things of deepmap-oapi-gen is that it can generate two server interfaces for you, a basic HTTP server and also a typed, domain-aware server that unmarshals all of the requests into the correct Golang type you defined in your OpenAPI spec. Any idea why this isn't getting generated?

In any case, if we _did_ generate this, then the generated server would _have_ to be in the same package as the types (and this also the client), otherwise the server would't know what types to use.

Maybe this nice server isn't getting generated because you've put the server into a separate package?

I think it would be very worth dealing with the nicer server interface. This makes the generated server much more useful, as it automatically checks and unpacks all of the request parameters for you and puts them into the correct types, etc. This is also how grpc clients/servers are normally pacakged, so it's not an unfamiliar pattern

Makefile
```diff
@@ -202,3 +202,9 @@ JSONNETFMT_CMD := $(JSONNETFMT) -n 2 --max-blank-lines 2 --string-style s --comm
 .PHONY: jsonnet-fmt
 jsonnet-fmt: | $(JSONNETFMT)
 	PATH=$$PATH:$(BIN_DIR):$(FIRST_GOPATH)/bin echo ${JSONNET_SRC} | xargs -n 1 -- $(JSONNETFMT_CMD) -i
+
+.PHONY: generate-rules-spec
+generate-rules-spec:
```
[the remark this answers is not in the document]
squat: As mentioned below, I think it would be worth generating all of the files into the same go package, as this should allow us to produce better server stubs. BTW, in this case you can collapse these three generation calls into one single command, e.g.
```sh
@oapi-codegen -generate types,client,chi-server -package rules -o rules/rules.go
```
This means also that the Makefile target would not need to be phony.

squat, line 207: We might want to rename this as we are not generating a spec, but rather using a spec to generate code
```suggestion
generate-rules-code:
```

Makefile
```diff
@@ -202,3 +202,9 @@ JSONNETFMT_CMD := $(JSONNETFMT) -n 2 --max-blank-lines 2 --string-style s --comm
...
+	@oapi-codegen -generate client -o rules/client.go -package rules rules/spec.yaml
```
squat, line 208: We should definitely use Bingo to vendor and version oapi-gen as a tooling dependency of this project so that we can reproducibly regenerate code, so all devs are guaranteed to use the same oapi-gen version and CI can verify the generation
```suggestion
	$(OAPIGEN) -generate client -o rules/client.go -package rules rules/spec.yaml
```

rules/spec.yaml
```diff
@@ -0,0 +1,98 @@
...
+          type: string
+          example: '1m'
+        name:
+          type: string
+          example: 'telemeter.rules'
+        rules:
+          type: array
+          items:
+            $ref: '#/components/schemas/Rule'
+    Rule:
```
bwplotka, line 82: Ok, I think the problem here is that we want to return alerting rules too. So we might want to have some union here.

Also ideally this is exactly the same as Prometheus Rules HTTP API, which has some specifics... ): 

We tried to improve this on Thanos by using protobuf and this did not pan-out really: https://thanos.io/tip/proposals-accepted/202107-protobuf-openapi-httpapi.md/

jessicalins:
Holding this for now - I think we need to align:
- if the code generated from OpenAPI would work and conform with the Prometheus Rules endpoint API (probably not, as mentioned [here](https://github.com/observatorium/api/pull/178#discussion_r741091723))
- result from this ^ if we want to use OpenAPI at all

Makefile
jessicalins: @squat I've added the oapi-gen using `bingo get github.com/deepmap/oapi-codegen/cmd/oapi-codegen@v1.9.0` and modified the Makefile too - but now I get this error on CI: 
```bash
/go/bin/jsonnet-v0.16.0 -m examples/manifests examples/main.jsonnet | xargs -I{} sh -c 'cat {} | /go/bin/gojsontoyaml-v0.0.0-20200602132005-3697ded27e8c > {}.yaml && rm -f {}' -- {
/go/bin/oapi-codegen-v1.9.0 -generate types,client,chi-server -package rules -o rules/rules.go rules/spec.yaml
bash: line 1: /go/bin/oapi-codegen-v1.9.0: No such file or directory
make: *** [Makefile:207: generate-rules-code] Error 127

Exited with code exit status 2
```
I think I missed some step but not sure yet..do you have any ideas?

Makefile
```diff
@@ -202,3 +202,6 @@ JSONNETFMT_CMD := $(JSONNETFMT) -n 2 --max-blank-lines 2 --string-style s --comm
 .PHONY: jsonnet-fmt
 jsonnet-fmt: | $(JSONNETFMT)
 	PATH=$$PATH:$(BIN_DIR):$(FIRST_GOPATH)/bin echo ${JSONNET_SRC} | xargs -n 1 -- $(JSONNETFMT_CMD) -i
+
+generate-rules-code:
```
squat, line 206: Yes, @jessicalins the missing piece here is that you need to add `$(OAPI_CODEGEN)` as a dependency of the generation target, otherwise the binary will not be guaranteed to exist

jessicalins: thanks! added now

jessicalins:
I've changed the spec/generated code to also support alerting rules (using https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/ as base) could you take another look?

bwplotka (approved):
This looks ok, but how we stand with compatibility with native Prometheus rule API at the end?

 What's the decision on our goal and status for this PR? 🤗 
 
 Otherwise LGTM

squat:
> but how we stand with compatibility with native Prometheus rule API at the end?

During yesterday's community meeting we agreed that providing compete compatibility with the Prometheus rules API was out of scope for this proposal, which is geared towards providing an API for tenants to configure rules, not to read back what rules are firing etc. We agreed we should submit a new proposal to Observatorium to agree to implement compatibility with Prometheus's rules API.

Makefile
```diff
@@ -202,3 +202,6 @@ JSONNETFMT_CMD := $(JSONNETFMT) -n 2 --max-blank-lines 2 --string-style s --comm
 .PHONY: jsonnet-fmt
 jsonnet-fmt: | $(JSONNETFMT)
 	PATH=$$PATH:$(BIN_DIR):$(FIRST_GOPATH)/bin echo ${JSONNET_SRC} | xargs -n 1 -- $(JSONNETFMT_CMD) -i
+
+generate-rules-code: $(OAPI_CODEGEN)
```
squat, line 206: Let's make this a concrete target that produces a real file, rather than a phony target (or make it phony for convenience in addition to the real file). Let's also add this target as a dependency to the main API binary.
```suggestion
rules/rules.go: $(OAPI_CODEGEN)
```

jessicalins: added (also had to make it .PHONY so the file would be correctly updated when running `make rules/rules.go` or `make generate`)

regarding add as a dependency from the main API bin, you mean here?
https://github.com/observatorium/api/blob/d1f53ca3b69122164d0db2097d224f5391b862f4/Makefile#L55

I think the wildcard would already catch the `rules/rules.go` file?

Makefile
```diff
@@ -202,3 +202,7 @@ JSONNETFMT_CMD := $(JSONNETFMT) -n 2 --max-blank-lines 2 --string-style s --comm
 .PHONY: jsonnet-fmt
 jsonnet-fmt: | $(JSONNETFMT)
 	PATH=$$PATH:$(BIN_DIR):$(FIRST_GOPATH)/bin echo ${JSONNET_SRC} | xargs -n 1 -- $(JSONNETFMT_CMD) -i
+
+.PHONY: rules/rules.go
```
squat, line 206: This one should not be a `.PHONY` target, it produces a real file that can be tracked for updates
```suggestion
```

squat: I think it should still be made an explicit dependency of the binary target. Consider what happens if you `rm rules/rules.go`. If you then do `make tmp/bin/observatorium-api`, the wildcard doesn't catch this file because the file does not exist on disk, so the rules.go file is note generated.

jessicalins: if I don't make it .PHONY, when changing `spec.yaml` the `rules.go` is not regenerated/updated:
```bash
jessica@cupsofwonder:~/workspace/api|add-rules-api-spec⚡ ⇒  make rules/rules.go
make: 'rules/rules.go' is up to date.
```

same happens when using `make generate`..

good point with making the dependency explicit, will add

squat: aha. Yes, this target should _not_ be marked as `.PHONY`. The missing piece here is that the target needs to depend on the `rules/spec.yaml`, this way whenever the spec is newer than the generated file, make will know that the rules.go file needs to be regenerated.

Makefile
```diff
@@ -202,3 +202,7 @@ JSONNETFMT_CMD := $(JSONNETFMT) -n 2 --max-blank-lines 2 --string-style s --comm
...
+rules/rules.go: $(OAPI_CODEGEN)
```
squat, line 207: From above's comment:
```suggestion
rules/rules.go: $(OAPI_CODEGEN) rules/spec.yaml
```

Makefile
```diff
@@ -202,3 +202,7 @@ JSONNETFMT_CMD := $(JSONNETFMT) -n 2 --max-blank-lines 2 --string-style s --comm
...
+	$(OAPI_CODEGEN) -generate types,client,chi-server -package rules -o rules/rules.go rules/spec.yaml
```
squat, line 208: This should likely be "simplified" as such so that the output file is always in sync with the target name
```suggestion
	$(OAPI_CODEGEN) -generate types,client,chi-server -package rules -o $@ rules/spec.yaml
```

Makefile
jessicalins: nice, tested locally and worked :tada: thanks!

main.go
```diff
@@ -830,6 +833,8 @@ func parseFlags() (config, error) {
 		"The endpoint against which to send read requests for metrics. It used as a fallback to 'query.endpoint' and 'query-range.endpoint'.")
 	flag.StringVar(&rawMetricsWriteEndpoint, "metrics.write.endpoint", "",
 		"The endpoint against which to make write requests for metrics.")
+	flag.StringVar(&rawMetricsRulesEndpoint, "metrics.rules.endpoint", "",
+		"The endpoint against which to make get requests for listing recording rules and put requests for creating/updating recording rules.")
```
squat, line 837: The only last nit I have is that this is no longer _just_ recording rules, right? It can be alerting rules as well?

jessicalins: True! updated it

squat (approved):
LGTM good work @jessicalins :)

api/metrics/v1/http.go
```diff
@@ -274,5 +277,25 @@ func NewHandler(read, write *url.URL, upstreamCA []byte, opts ...HandlerOption)
...
+		if err != nil {
+			level.Warn(c.logger).Log("msg", "could not create rules endpoint client")
+			return r
+		}
+
+		rh := rulesHandler{client: client}
+
+		r.Group(func(r chi.Router) {
+			r.Use(c.uiMiddlewares...)
+			r.Get(rulesRoute, rh.get)
```
onprem, line 291: This route is same as the [Rules endpoint in Prom](https://prometheus.io/docs/prometheus/latest/querying/api/#rules) (also exposed by Thanos Querier) so it'll shadow it breaking anything (like Thanos Rules UI) that depends on it. I think we should change the `rulesRoute` to avoid confusion.

onprem (commented):
Looking good to me, just one thing I'd like to get addressed before we merge. Great work @jessicalins!

api/metrics/v1/http.go
jessicalins: very good point - I've changed it (now I think is in line with the [proposal](https://observatorium.io/docs/proposals/20201019-prometheus-rules-in-observatorium-api.md/) too)

jessicalins: changed it to `api/v1/rulesfile` since either `api/v1/rules` and `/rules` would still create conflict

api/metrics/v1/http.go
```diff
@@ -30,6 +32,7 @@ const (
 	labelNamesRoute  = "/api/v1/labels"
 	labelValuesRoute = "/api/v1/label/{label_name}/values"
 	receiveRoute     = "/api/v1/receive"
+	rulesRoute       = "/api/v1/rulesfile"
```
squat, line 35: naming is hard :see_no_evil: `/api/v1/rulesfile` sounds a bit funny to me though since when the user `put`s to the API, you're not really creating a file. hmm maybe `/api/v1/raw_rules`?

jessicalins: it is indeed :sweat_smile: we thought `rulesfile` because the content has the same format as the Prometheus rules file - but I agree that gets confusing.. I'm fine with `raw_rules` too @onprem what do you think?

onprem: Yep that sounds good. Another one I can think of from what Lucas suggested is `/api/v1/rules/raw`, which just feels a bit more natural to me. WDYT @squat @jessicalins ?

jessicalins: ^ I like it :) aligns better with the other rules endpoints too

squat: I second this

api/metrics/v1/rules.go
```diff
@@ -0,0 +1,67 @@
...
+func (rh *rulesHandler) get(w http.ResponseWriter, r *http.Request) {
+	tenant, ok := authentication.GetTenant(r.Context())
+	if !ok {
+		http.Error(w, "error finding tenant", http.StatusUnauthorized)
+		return
+	}
+
+	resp, err := rh.client.ListRules(r.Context(), tenant)
+	if err != nil {
+		http.Error(w, "error listing rules %w", http.StatusInternalServerError)
```
onprem, line 24: Maybe it makes sense to propagate the HTTP status code we got from the rules backend instead of `500`. I am thinking about the cases when it fails some validation checks on Rules Backend side and it returns a `400 Bad Request`.

api/metrics/v1/rules.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+}
+
+func (rh *rulesHandler) put(w http.ResponseWriter, r *http.Request) {
+	tenant, ok := authentication.GetTenant(r.Context())
+	if !ok {
+		http.Error(w, "error finding tenant", http.StatusUnauthorized)
+	}
+
+	resp, err := rh.client.SetRulesWithBody(r.Context(), tenant, r.Header.Get("Content-type"), r.Body)
```
onprem, line 49: ditto.

onprem (approved):
Looking great to me! Just a non-blocking suggestion we can iterate on future PRs.

squat (approved):

squat:
Great work @jessicalins :tada: :taco: Let's get this merged

onprem (approved):
Great work! :100:

PR #181, opened 2021-11-01
Respond with status OK on root path

matej-g:

Small change that makes the API respond with `200` on root path `/` instead of current `404`. This is in order to enable us to do some basic closed-box checking, to differentiate between 'real' not found case and case where API is actually serving.

Also bumps the version for `signal` package used in internal server instantiation.

### DIFF
--- a/main.go
+++ b/main.go
@@ -398,7 +398,7 @@ func main() {
 					&http.Client{Transport: t},
 					cfg.server.healthcheckURL,
 					http.MethodGet,
-					http.StatusNotFound,
+					http.StatusOK,
 					time.Second,
 				),
 			)
@@ -646,6 +646,10 @@ func main() {
 				cancel()
 			})
 		}
+
+		r.Get("/", func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusOK)
+		})
 
 		s := http.Server{
 			Addr:         cfg.server.listen,
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/gorilla/websocket v1.4.2
 	github.com/grpc-ecosystem/go-grpc-middleware v1.0.1-0.20190118093823-f849b5445de4
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
-	github.com/metalmatze/signal v0.0.0-20201002154727-d0c16e42a3cf
+	github.com/metalmatze/signal v0.0.0-20210307161603-1c9aa721a97a
 	github.com/oklog/run v1.1.0
 	github.com/open-policy-agent/opa v0.23.2
 	github.com/pkg/errors v0.9.1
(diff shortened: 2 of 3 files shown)


main.go
```diff
@@ -398,7 +398,7 @@ func main() {
 					&http.Client{Transport: t},
 					cfg.server.healthcheckURL,
 					http.MethodGet,
-					http.StatusNotFound,
+					http.StatusOK,
```
moadz, line 401: nit: Does it make more sense to return a 204 (no content) if it's simply for liveness?

moadz (commented):
LGTM % nit

squat:
Hmm this seems wrong to me. There is literally nothing to serve at this path, so from an HTTP point of view the correct status is 404

moadz:
> Hmm this seems wrong to me. There is literally nothing to serve at this path, so from an HTTP point of view the correct status is 404

@squat there are other non-4** status codes for communicating an OK status with no payload response. However, i'm not certain a 204 is appropo either.

squat:
Right, but those typically indicate that an operation was performed or was scheduled, which is not the case here. In this case it seems like a hack to enable testing. More than anything, i worry this change will be confusing as the API literally does nothing with this endpoint.

An option would be to respond with a list of available endpoints, e.g. use / as an HTTP API discovery endpoint. Of course, care would have to be taken to not leak tenant metadata, e.g. tenant X exists.

matej-g:
> Right, but those typically indicate that an operation was performed or was scheduled, which is not the case here. In this case it seems like a hack to enable testing. More than anything, i worry this change will be confusing as the API literally does nothing with this endpoint.
> 
> An option would be to respond with a list of available endpoints, e.g. use / as an HTTP API discovery endpoint. Of course, care would have to be taken to not leak tenant metadata, e.g. tenant X exists.

I can certainly see the point, although I do not think this is a hack just for testing purposes. The API might not be doing anything _per se_, but it's still a signal to the user that 'yup, you're in the right place'. The API is doing something - it's telling you're on the right address, this is the base / host root path. IMHO it's not an uncommon thing.

But I also agree that listing available endpoints would be a much nicer addition (we already had users who were unsure about paths they were using), so if that'll make this more palatable I'm happy to do it.

ianbillett:
@matej-g just wondering two things here:
* Is there a precedent set for this in any of our upstream repos - Thanos / Prometheus?
* Could we just call the `/metrics` endpoint or even `/healthy` like the kube probe that indicates service health?

matej-g:
> @matej-g just wondering two things here:
> 
> * Is there a precedent set for this in any of our upstream repos - Thanos / Prometheus?

AFAIU Prometheus / Thanos does not have this 'problem', since they redirect from `/` to `/graph` (or in case of Thanos UI, possibly to other endpoint, depending on the component). But for example, for Telemeter, we do return available [external routes](https://github.com/openshift/telemeter/blob/master/cmd/telemeter-server/main.go#L541) on `/`, which was also the suggested alternative from @squat.

> * Could we just call the `/metrics` endpoint or even `/healthy` like the kube probe that indicates service health?

Since these are served by the internal server, they are not reachable from the outside.

I will rework this to make the `/` serve list of available external endpoints, since that seems like the more agreeable option for most of us here and we can iterate / discuss on that.

matej-g:
Closing in favor of https://github.com/observatorium/api/pull/182

PR #182, opened 2021-11-03
List available paths on root path

matej-g:
This PR adds a new handler for `/` to return a list of available API paths. The purpose of this is to improve discoverability of the API paths and to improve the user experience.

The changes also including bumping the `signal` packaged user for internal server and adds some missing docs.

### DIFF
--- a/authentication/authentication.go
+++ b/authentication/authentication.go
@@ -35,7 +35,7 @@ func onboardNewProvider(providerType string, factory ProviderFactory) {
 // authentication provider.
 type ProviderFactory func(config map[string]interface{}, tenant string, registrationRetryCount *prometheus.CounterVec, logger log.Logger) (Provider, error)
 
-// Provider is an interface that should be implemented to onboard a new authentication.
+// Provider is an interface that should be implemented to onboard a new authentication
 // provider.
 type Provider interface {
 	Middleware() Middleware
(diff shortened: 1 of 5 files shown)


authentication/authentication.go
```diff
@@ -102,6 +102,7 @@ func (ah *ProviderManager) InitializeProvider(config map[string]interface{},
 	return authCh
 }
 
+// Middleware returns an authentication middleware for a tenant.
```
tareqmamari, line 105: Thanks for fixing these @matej-g. Middleware should be Middlewares

matej-g: Ah missed that, thanks! Will fix it.

ianbillett (commented):
@matej-g please can you provide an example of the response that we get from this change?

To @squat 's point in the previous PR - we don't want to be leaking any sensitive information to the world.

moadz (approved):
LGTM

matej-g:
> @matej-g please can you provide an example of the response that we get from this change?
> 
> To @squat 's point in the previous PR - we don't want to be leaking any sensitive information to the world.

It basically lists the routes but only as patterns, i.e. the actual tenant names etc. are not substituted. Example:
```
{
  "paths": [
    "/api/logs/v1/{tenant}/*",
    "/api/metrics/v1/{tenant}/*",
    "/api/v1/{tenant}/*",
    "/{tenant}"
  ]
}
```

ianbillett (approved):
LGTM - nice stuff

periklis (commented):
Looks good to me. Is it possible to list the valid HTTP verbs per path?

matej-g:
> Looks good to me. Is it possible to list the valid HTTP verbs per path?

Hah, funnily enough when I tried to list it with methods, I'm seeing we simply accept _any_ method on any endpoint. I think we should be returning `405` for the ones which do not apply. I'm creating a ticket for this. I would then suggest to revisit this once we have the incorrect methods behaving correctly.

authentication/authentication.go
squat: ping here

PR #184, opened 2021-11-08
Introduce pluggable authorization

tareqmamari:
This commit introduces new pluggable authorization setup in
observatorium api, it includes some minor refactoring on the
existing authorization implementation and converting the opa
and rbac authorizers as plugins.

The proposal of this implementation should be found here: https://github.com/observatorium/observatorium/pull/438

Signed-off-by: Neyyan Sony <neyysony@in.ibm.com>


tareqmamari:
@observatorium/maintainers based on the last community call, we opened a separate proposal for the pluggable authorization portion + this implementation PR, we have also applied the feedback we got from the previous pluggable authentication PR. @periklis I know you had few concerns before, please let me know if anything is needed in this PR, thank you!

periklis:
> @observatorium/maintainers based on the last community call, we opened a separate proposal for the pluggable authorization portion + this implementation PR, we have also applied the feedback we got from the previous pluggable authentication PR. @periklis I know you had few concerns before, please let me know if anything is needed in this PR, thank you!

As mentioned on our call, switching authorizers to a pluggable model is not really a goal for observatorium-api. The core case is that OPA is the actual interface to add your custom authorizer. Event the static RBAC is based upon it using the in-process OPA authorizer.

sven-lange-last:
@periklis we discussed the topic of pluggable authorization "vs" OPA based authorization in today's Observatorium community call. See the minutes.

Summary:

* Today, we have two different OPA based authorizers in the Observatorium API: REST API and in-process.
* REST API based authorizer: we expect performance problems with that approach and challenges with running the authorizer in a sidecar container to the API container when using ServiceMesh (Istio). For this reason, we would prefer to go with the in-process approach.
* In-process: at the moment, OPA / Rego does not come with the required support for the proprietary authorization system we have to use. We have to use a proprietary Go library for making authorization decisions that implements its own caching mechanism.
* OPA provides extension mechanisms to add functions to Rego: https://www.openpolicyagent.org/docs/latest/extensions/.
* At the moment, we don't have any experience with or knowledge about extending OPA. We would have to explore this approach.
* As mentioned: the proprietary Go library for making authorization decisions implements its own caching mechanism. For this reason, we have to turn off any OPA caching.
* What’s also important from our perspective: we would like to add an extension mechanism to Observatorium API that allows adopters to add their extensions with a minimal risk of merge conflicts when the community code evolves. That’s why we suggested https://github.com/observatorium/observatorium/pull/428.
* We discussed to perform a time-boxed exploration of what we can achieve with an in-process OPA based authorizer and share our results in the next community meeting. Based on the results, we should discuss how we want to move on.

@tareqmamari @squat feel free to chime in.

@periklis what do you think?

tareqmamari:
To address some of the concerns above, we are closing this in favor of https://github.com/observatorium/api/pull/197

PR #185, opened 2021-11-09
OIDC Auth: Redirect to a valid URL after `/callback`

matej-g:

Currently, if a user goes through the authentication flow in the browser, the last step is to call `/callback` endpooint, which subsequently redirects user further with authentication set up. However, at the moment we are not setting any redirect URL, causing the user to end up at route `/oidc/{tenant}`, which does not exist.

This PR adjusts the flow to redirect the user to `/{tenant}` which in turn redirects to the Thanos UI (`/graph` endpoint), which is much better user experience.

Tested locally.


onprem (approved):
LGTM, thanks for the quick fix!

periklis (commented):
LGTM

PR #186, opened 2021-11-11
Simplify Rules API

squat:
Rather than depending on the exact implementation of the rules API
client, we should be depending on the interface. This makes it possible
to better test our code as well as allows for generally better
abstraction.

This commit allows the rules API server to stream the responses from the
rules API backend back to the client, allowing us to avoid buffering the
responses in the Observatorium API.


matej-g (approved):
Looks good!

onprem (approved):
LGTM, Thanks!

PR #187, opened 2021-11-12
Add metric for skipped tenant configuration

matej-g:

If an issue with a tenant configuration occurs, that prevents the tenant from being configured on the API start up, we do not interrupt the process, since we still want the API to come up with the rest of correctly configured tenants. We only issue a warning for tenants which could not be configured.

To increase the visibility of such issue, this PR introduces a metric to track any tenants skipped due to misconfiguration.

### DIFF
--- a/main.go
+++ b/main.go
@@ -224,6 +224,7 @@ func main() {
 		collectors.NewGoCollector(),
 		collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}),
 	)
+
 	skippedTenants := promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
 		Namespace: "observatorium",
 		Subsystem: "api",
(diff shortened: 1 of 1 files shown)


onprem (approved):
Looks good!

main.go
```diff
@@ -217,6 +218,20 @@ func main() {
...
+	reg.MustRegister(
+		version.NewCollector("observatorium"),
+		collectors.NewGoCollector(),
+		collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}),
+	)
+
+	skippedTenants := promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+		Namespace: "observatorium",
+		Subsystem: "api",
+		Name:      "tenants_skipped_invalid_configuration",
```
ianbillett, line 231: My only thought here is whether `tenants_skipped` is slightly ambiguous - perhaps `tenants_invalid_configuration` is more direct? If a tenant has invalid configuration - they are not being registered!

ianbillett: Not a blocker tho :man_shrugging:

periklis (commented):
LGTM

PR #188, opened 2021-11-12
api/metrics/v1: fix rules panic

squat:
This commit fixes two bugs where a failed HTTP request from the rules
client can cause a panic in the Observatorium API.
1. status codes were read prematurely when the HTTP response may have
   been nil, causing a panic; and
2. the set handler did not return after a failed request, causing the
   HTTP response to be set twice.
It also fixes a formatting bug in the HTTP status where an extraneous
`%w` was printed.


periklis (approved):

jessicalins (approved):
good catch! thanks for fixing this :pray: :tada:

PR #190, opened 2021-11-14
Makefile: Adjust `container-test` target to work for MacOS

### LINKED ISSUE #189: "make test-interactive" gives "exec format error" on MacOS
"make test-interactive" gives "exec format error" on MacOS.

Almost certainly the problem is that the _Makefile_ builds _observatorium_api_ for the OS, but then packages the OS-specific binary in the Linux-specific container image.

Work around:

```bash
rm ./observatorium-api
OS=linux GOARCH=amd64 make test-interactive
```

Output without the work-around
```
16:13:17 Starting observatorium_api
16:13:18 observatorium_api: standard_init_linux.go:228: exec user process caused: exec format error
16:13:52 
Error: No such object: e2e_interactive-observatorium_api

    interactive_test.go:33: interactive_test.go:33:
        
         unexpected error: docker container observatorium_api failed to start: exit status 1
```

matej-g:

We have a report that running the interactive and E2E tests does not work on MacOS, which is due to the fact that we use single stage Docker build for local tests and simply copy the built API binary into the image. This does not work for e.g. MacOS users.

The fix is to use the 'full' build for other OS users and build the API for right platform instead of copying the binary.

Resolves #189.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,10 +122,18 @@ proto: ratelimit/gubernator/proto/google ratelimit/gubernator/gubernator.proto $
 	PATH=$$PATH:$(BIN_DIR):$(FIRST_GOPATH)/bin scripts/generate_proto.sh
 
 .PHONY: container-test
+container-test: # Use 'shortcut' to build test image if on Linux, otherwise full build.
+ifeq ($(OS)_$(ARCH), linux_x86_64)
 container-test: build
 	@docker build \
 		-f Dockerfile.e2e-test \
 		-t $(DOCKER_REPO):local_e2e_test .
+else
+container-test:
+	@docker build \
+		-f Dockerfile \
+		-t $(DOCKER_REPO):local_e2e_test .
+endif
 
 .PHONY: container
 container: Dockerfile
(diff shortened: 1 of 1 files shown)


esnible:
Works for me (macOS Big Sur 11.6; Docker version 20.10.8)

Makefile
```diff
@@ -122,10 +122,18 @@ proto: ratelimit/gubernator/proto/google ratelimit/gubernator/gubernator.proto $
 	PATH=$$PATH:$(BIN_DIR):$(FIRST_GOPATH)/bin scripts/generate_proto.sh
 
 .PHONY: container-test
+container-test: # Use 'shortcut' to build test image if on Linux, otherwise full build.
+ifeq ($(OS)_$(ARCH), linux_x86_64)
```
squat, line 126: Hmm this doesn't seem right to me. If I do my development on an ARM machine then this change means I also don't benefit from binaries built as past of the dev flow? Maybe this should check for Linux in general?

matej-g: I adjusted it, although if we want to realistically support running tests on ARM, we would need to have other images we are using available for ARM platform, which does not seem to be the case.

squat: Thanks :) yes I think it's a big shame that Thanos and Observatorium images are not multi-arch by default. This is definitely not a best practice and is quite antiquated at this point.

matej-g: I actually thought Thanos can be built for ARM :thinking:, but I see it does not push images for general use. I also think many 'side-services' such as Gubernator are also not readily available for ARM. 

I agree with the sentiment though, would be nice to have the support :+1:.

squat (approved):

PR #191, opened 2021-11-14
Enforce Labels on Returned Rules

squat:
Currently, recording and alerting rules that are fetched from the new
rules/raw API will not have any tenant label attached to them. This
means that when the rules are configured on a Thanos Ruler, the time
series that are generated will not have a tenant label on them. This is
a big problem, because without this tenant label, the very tenant to
whom the rules belong cannot query for the time series, making them in
effect useless. This PR enforces that all rules returned by the
rules/raw endpoint have a tenant_id label attached to them.

Note: this is a distinct problem from ensuring that the rules written by
a tenant are guaranteed to only include data from that tenant.
Currently, the rules from a tenant can select data from any tenant. This
is a distinct problem and should absolutely be tackled in a follow up.

In order to be able to apply the labels, we must assert the type of the rules.
Currently, when a group of rules is unmarshaled, the concrete type in Go
will never be a rules.RecordingRule or a rules.AlertingRule, although
those types are defined in the rules package. Instead those rules will
always be some map[string]interface{} type. This means that it's not
possible for users of the package to perform type assertions on the
rules that are fetched via the client, which is a serious annoyance for
anyone using the package.

This PR also fixes this by providing a custom unmarshaling function.
Furthermore, in order to fix unmarshaling via both JSON and YAML, the
PR switches the YAML package to ghodss/yaml, which is a superset of
the current YAML package, go-yaml/yaml.

Finally, rules are either Recording Rules or Alerting Rules, but not both. This
PR fixes this fact in the OpenAPI specification.

### DIFF
--- a/api/metrics/v1/rules.go
+++ b/api/metrics/v1/rules.go
@@ -2,14 +2,20 @@ package v1
 
 import (
 	"io"
+	"io/ioutil"
 	"net/http"
 
+	"github.com/ghodss/yaml"
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
 	"github.com/observatorium/api/authentication"
 	"github.com/observatorium/api/rules"
 )
 
 type rulesHandler struct {
-	client rules.ClientInterface
+	client      rules.ClientInterface
+	logger      log.Logger
+	tenantLabel string
 }
 
 func (rh *rulesHandler) get(w http.ResponseWriter, r *http.Request) {
@@ -19,8 +25,16 @@ func (rh *rulesHandler) get(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	id, ok := authentication.GetTenantID(r.Context())
+	if !ok {
+		http.Error(w, "error finding tenant ID", http.StatusUnauthorized)
+		return
+	}
+
 	resp, err := rh.client.ListRules(r.Context(), tenant)
 	if err != nil {
+		level.Error(rh.logger).Log("msg", "could not list rules", "err", err.Error())
+
 		sc := http.StatusInternalServerError
 		if resp != nil {
 			sc = resp.StatusCode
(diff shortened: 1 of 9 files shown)


api/metrics/v1/rules.go
```diff
@@ -33,8 +47,56 @@ func (rh *rulesHandler) get(w http.ResponseWriter, r *http.Request) {
...
+			case rules.AlertingRule:
+				if r.Labels.AdditionalProperties == nil {
+					r.Labels.AdditionalProperties = make(map[string]string)
+				}
+
+				r.Labels.AdditionalProperties[rh.tenantLabel] = id
+				rawRules.Groups[i].Rules[j] = r
+			}
+		}
+	}
```
onprem, line 88: Do we need to enforce this label on the read path as well? Since the label is enforced during writes, we can just assume that it will be already present.

squat: The labels are not enforced anywhere in the write path currently. Also even if they wereb enforced on the write path, it wouldn't be enough because there's no telling how rules made it into a particular rule backend. In other words, we cannot be sure that the write path was even used at all to configure the backend. Perhaps the backend is read only and was configured using a static file outside of the workflow of making a PUT request to Observatorium, so there's no guarantee that the rules in the file have the right tenant ID. Finally, even though we proxy the tenant name in all requests to the backend, that's not good enough for the backend to enforce labels as the labels must include the tenant ID, not the tenant name.

rules/custom_types_test.go
```diff
@@ -0,0 +1,206 @@
...
+- record: bar
+  expr: vector(1)
+- alert: HighRequestLatency
+  expr: job:request_latency_seconds:mean5m{job="myjob"} > 0.5
+  for: 10m`),
+			out: RuleGroup{
+				Name:     "foo",
+				Interval: "5s",
+				Rules: []interface{}{
+					RecordingRule{
```
jessicalins, line 75: so we can still get both recording and alerting rules in the same request, but we wouldn't like to have a rule that has both `alert` and `record` properties, right? is that why `anyOf` was changed to `oneOf`? https://github.com/observatorium/api/pull/191/commits/d08124476f3ec0716dadafe6bc8005767439be51#diff-d523f469970611999b9968425756e5cc8c840626d7e592cf514492c52215c9e6L81
Just wanted to make sure I understand since here we expect one `RuleGroup` with both alerting/recording rules

rules/custom_types_test.go
```diff
@@ -0,0 +1,206 @@
...
+						Expr:   "vector(1)",
+						Labels: RecordingRule_Labels{AdditionalProperties: make(map[string]string)},
+					},
+				},
+			},
+		},
+		{
+			name: "one alerting rule",
+			raw: []byte(`name: foo
+interval: 5s
```
jessicalins, line 42: I think would be more readable if we format/align the content of `raw` (same in the other test inputs)

jessicalins (commented):
cool, thanks for catching that!
left one question/suggestion

rules/custom_types_test.go
squat: That's correct. If it was anyOf, then a single rule would be allowed to have both `alert` and `record` fields defined, which is not what we want. https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/ that's my reading at least. The section `Difference Between anyOf and oneOf` seems to suggest that

rules/custom_types_test.go
squat: This is a literal YAML string, and indentation matters in YAML, that's why it's like this :/

squat: We could instead add an extra newline before the first line so they are all aligned totally to the left. WDYT?

rules/custom_types_test.go
jessicalins: oki, this makes sense :+1:

rules/custom_types_test.go
jessicalins: ah right :/
cool, I think doing as you suggested reads better!

squat: OK, updated :heart:

jessicalins (approved):
awesome :tada: 
I've created a ticket for the note you mentioned in the description of the PR, so we don't forget about it: https://issues.redhat.com/browse/MON-2063

matej-g (approved):
Looks good :+1:

onprem (approved):
LGTM, Thanks!!

PR #193, opened 2021-11-17
Debug problems of authorization

esnible:
It is quite difficult to debug 403 problems.  This allows the admin to see the problems.

For example, when I tried to substitute Keycloak as OIDC provider, I must have misconfigured it.  Without this PR, nothing is logged and the user just sees "403".  With this log the user sees "403 authorization declined", which gives a hint to where to look in the code.

The **admin** sees 

```
tenant "test-oidc" unknown; valid tenants are map[test:{map[{admin@example.com user}:{}] map[{admin@example.com user}:{}]}]
```

(Probably this currently logs too much.  That is another TODO)

### DIFF
--- a/authorization/http.go
+++ b/authorization/http.go
@@ -75,7 +75,8 @@ func WithAuthorizers(authorizers map[string]rbac.Authorizer, permission rbac.Per
 
 			statusCode, ok, data := a.Authorize(subject, groups, permission, resource, tenant, tenantID, token)
 			if !ok {
-				http.Error(w, "authorization declined", statusCode)
+				// Send 403 http.StatusForbidden
+				w.WriteHeader(statusCode)
 
 				return
 			}
--- a/authorization/http.go
+++ b/authorization/http.go
@@ -75,7 +75,7 @@ func WithAuthorizers(authorizers map[string]rbac.Authorizer, permission rbac.Per
 
 			statusCode, ok, data := a.Authorize(subject, groups, permission, resource, tenant, tenantID, token)
 			if !ok {
-				w.WriteHeader(statusCode)
+				http.Error(w, "authorization declined", statusCode)
 
 				return
 			}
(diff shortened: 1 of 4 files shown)


esnible:
@squat Can you look at this for the basic direction?

I realize it is not ready yet.  Three things.
1. Is there a clever way to get the logger or should I start passing it down into _rbac.go_ methods?
2. I am confused why `ts`, which is supposed to contain tenants, contains users.  I will dig deeper into this today.
3. Adding "authorization declined" in the 403 response to the client should not be a problem, right?

matej-g (commented):
Looks good, thanks you! :+1: Only thing missing is passing the lint check, afterwards should be good to go.

esnible:
@matej-g fixed lint issue; ready!

matej-g (dismissed):

authorization/http.go
```diff
@@ -75,7 +75,7 @@ func WithAuthorizers(authorizers map[string]rbac.Authorizer, permission rbac.Per
 
 			statusCode, ok, data := a.Authorize(subject, groups, permission, resource, tenant, tenantID, token)
 			if !ok {
-				w.WriteHeader(statusCode)
+				http.Error(w, "authorization declined", statusCode)
```
squat, line 78: I'm not convinced that saying "authorization declined" is any more informative to the user than the interpretation of a 403 as "forbidden".
But not a blocker

squat (approved):
lgtm modulo the single comment

authorization/http.go
esnible: The reason for this code is because as an outsider it took me a long time to realize where to start looking in the code to understand the reason for the 403.  The code generating the 403 forbidden was `w.WriteHeader(statusCode)` which has no "403" and no "forbidden".

The alternative I considered was a comment.  Which do you think is preferrable?

```
// Send 403 http.StatusForbidden
w.WriteHeader(statusCode)
```

squat: I think the second is better. Let's do that and merge this :)

esnible: OK, done.  (The commit of the suggested alternative caused Github to believe I dismissed your reviews.)

squat (approved):
Thanks @esnible

PR #198, opened 2021-12-17
Rules API: Add e2e test

jessicalins:
This PR ~~starts to~~ adds an e2e test to the Rules API. ~~Already opening as a draft to get some feedback about the overall structure.~~

~~What is currently missing:~~
~~- Test thanos-ruler-syncer flow~~ - after talking with @onprem we thought not adding a test here since this is already covered in the other repos (`rules-objstore` / `thanos-rule-syncer`) 
~~- Test 'unhappy' cases~~


test/e2e/rules_test.go
```diff
@@ -0,0 +1,123 @@
...
+func TestRulesAPI(t *testing.T) {
+	t.Parallel()
+
+	e, err := e2e.NewDockerEnvironment(envRulesAPIName)
+	testutil.Ok(t, err)
+	t.Cleanup(e.Close)
+
+	prepareConfigsAndCerts(t, rules, e)
+	_, token, rateLimiterAddr := startBaseServices(t, e, rules)
+	_, _, rulesEndpoint, _ := startServicesForMetrics(t, e) // TODO: create another function just for rules?
```
matej-g, line 24: I think that TODO would make sense :+1:

test/e2e/rules_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+	api, err := newObservatoriumAPIService(
+		e,
+		withMetricsEndpoints("", "", "http://"+rulesEndpoint),
```
matej-g, line 28: I think we might need to provide some write/read endpoint, otherwise API will complain that neither metrics nor logs endpoints are provided?

test/e2e/services.go
```diff
@@ -25,6 +25,7 @@ const (
 	dexImage        = "dexidp/dex:v2.30.0"
 	opaImage        = "openpolicyagent/opa:0.31.0"
 	gubernatorImage = "thrawn01/gubernator:1.0.0-rc.8"
+	rulesBackendImage = "quay.io/observatorium/rules-objstore:main-2021-11-22-cc68e25"
```
matej-g, line 28: Maybe
```suggestion
	rulesObjectStoreImage = "quay.io/observatorium/rules-objstore:main-2021-11-22-cc68e25"
```

test/e2e/services.go
```diff
@@ -174,6 +186,27 @@ func newLokiService(e e2e.Environment) *e2e.InstrumentedRunnable {
 	)
 }
 
+func newRulesBackendService(e e2e.Environment) *e2e.InstrumentedRunnable {
+	ports := map[string]int{"http": 8080, "internal": 8081}
+
+	args := e2e.BuildArgs(map[string]string{
+		"-log.level":            logLevelDebug,
```
matej-g, line 193: Is the object store using `-` for long options instead of `--`?

matej-g (commented):
Looks good so far :+1: Will wait on the remaining parts of the PR.

test/e2e/services.go
jessicalins: added `--` to be consistent :)

test/e2e/rules_test.go
jessicalins: I think for the test environment, since is using https://github.com/observatorium/api/blob/main/test/e2e/services.go#L224 this does not apply? It seems to append in case they are not empty: https://github.com/observatorium/api/blob/main/test/e2e/services.go#L250-L253 but no validation in case they are empty.

I also searched in the test output for this error message `"Neither logging nor metrics endpoints are enabled. Specifying at least a logging or a metrics endpoint is mandatory"` from https://github.com/observatorium/api/blob/main/main.go#L199-L202 but I haven't found any matches

test/e2e/rules_test.go
```diff
@@ -0,0 +1,170 @@
...
+		testutil.Ok(t, err)
+		testutil.Equals(t, http.StatusOK, res.StatusCode)
+
+		body, err := ioutil.ReadAll(res.Body)
+		bodyStr := string(body)
+		assertResponse(t, bodyStr, "record: job:up:avg")
+		assertResponse(t, bodyStr, "alert: ManyInstancesDown")
+	})
+
+	t.Run("put-invalid-rules", func(t *testing.T) {
```
jessicalins, line 156: @onprem for the case that invalid rules are submitted, I keep receiving no error/200 HTTP response status code, instead of https://github.com/observatorium/rules-objstore/blob/main/pkg/server/server.go#L80 
I got this in the test output:
```bash
17:10:54 rules_objstore: level=debug name=rules-objstore ts=2022-01-19T20:10:54.115959093Z caller=server.go:81 component=server handler=setrules tenant=test-oidc msg="request body failed rule group validation" errs="unsupported value type"
17:10:54 observatorium_api: level=debug name=observatorium ts=2022-01-19T20:10:54.116175567Z caller=instrumentation.go:36 request=observatorium_api/NKsl1S6qrw-000009 proto=HTTP/1.1 method=PUT status=200 content= path=/api/metrics/v1/test-oidc/api/v1/rules/raw duration=1.492672ms bytes=42
```
I'll try to take a better look on this tomorrow, but I think either obs API is not fetching err/status code or rules-objstore is not returning/propagating them properly

test/e2e/configs.go
```diff
@@ -18,6 +18,7 @@ type testType string
 
 const (
 	metrics     testType = "metrics"
+	rules 		testType = "rules"
```
ianbillett, line 21: `make format` should fix these rogue whitespaces :+1:

test/e2e/rules_test.go
```diff
@@ -0,0 +1,170 @@
...
+	rulesEndpointURL := "https://" + api.Endpoint("https") + "/api/metrics/v1/test-oidc/api/v1/rules/raw"
+	tr := &http.Transport{
+		TLSClientConfig: getTLSClientConfig(t, e),
+	}
+
+	client := &http.Client{
+		Transport: &tokenRoundTripper{rt: tr, token: token},
+	}
+
+	t.Run("get-put-recording-rules", func(t *testing.T) {
```
ianbillett, line 43: `get-put-recording-rules` this is kind of confusing to me as to what the test is aiming to do?

test/e2e/rules_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	api, err := newObservatoriumAPIService(
+		e,
+		withRulesEndpoint("http://"+rulesEndpoint),
+		withRateLimiter(rateLimiterAddr),
+	)
+	testutil.Ok(t, err)
+	testutil.Ok(t, e2e.StartAndWaitReady(api))
+
+	rulesEndpointURL := "https://" + api.Endpoint("https") + "/api/metrics/v1/test-oidc/api/v1/rules/raw"
```
ianbillett, line 34: Is the `test-oidc` tenant name defined anywhere else? I just imagine if we change that we'll unexpectedly break this test

test/e2e/configs.go
jessicalins: weirdly if I run `make format` locally I get:
```bash
jessica@cupsofwonder:~/workspace/api|rules-add-e2e ⇒  make format
/home/jessica/go/bin/golangci-lint-v1.21.0 run --fix --enable-all -c .golangci.yml
WARN [runner/nolint] Found unknown linters in //nolint directives: cyclop, exhaustivestruct, goerr113, gomnd, intefacer, paralleltest, testpackage 
```
so somehow it did not give any more output than that and also did not change any of the files.

I then tried to run with my local `golangci-lint` (with the latest version - 1.43.0) and it found some issues in the code, so I think we may tackle this separately? I wonder if would be good to also update the `golangci-lint` version (with the latest version I get also some warnings of linters that are deprecated and we use them in the `nolinter` directive (e.g. `WARN [runner/nolint] Found unknown linters in //nolint directives: intefacer`)

Should I create a separate issue for that? Maybe makes sense to also link to this issue @matej-g already opened: https://github.com/observatorium/api/issues/212
For now I run manually `gofmt` to this config file, so the whitespaces should be fixed

test/e2e/rules_test.go
jessicalins: I renamed to `read-write-recording-rules` (also to the further ones) - hopefully now is clearer?
It basically tests the reading (`GET`) / writing (`PUT`) of rules via the rules API defined in observatorium/api going through `rules-objectstore` https://github.com/observatorium/api/pull/198/files#diff-368252fd7020798243fc9b3793b03207f23440bf25424123e38b77eabf42c3d3R28

test/e2e/configs.go
jessicalins: Also, shouldn't the lint step fail?

test/e2e/rules_test.go
jessicalins: good point..we define it manually everywhere in the tests: https://github.com/observatorium/api/search?q=test-oidc 
maybe makes sense to define it as a constant in `configs.go` and use its value? I could open another pr for that or just already change it here (?)

test/e2e/rules_test.go
```diff
@@ -0,0 +1,170 @@
...
+		res, err = client.Do(r)
+		defer res.Body.Close()
+
+		testutil.Ok(t, err)
+		testutil.Equals(t, http.StatusOK, res.StatusCode)
+
+		body, err := ioutil.ReadAll(res.Body)
+		bodyStr := string(body)
+
+		assertResponse(t, bodyStr, "sum by (job) (http_inprogress_requests)")
```
saswatamcode, line 86: It might be a good idea to check for tenant labels here? 🙂
With something like: `assertResponse(t, bodyStr, "tenant_id: 1610b0c3-c509-4592-a256-a1871353dbfa")` (id taken from [here](https://github.com/observatorium/api/blob/8fcfcc34a15759905dad0ed9c75a893594b2877f/test/e2e/configs.go#L44)).

test/e2e/rules_test.go
jessicalins: added as a constant here: https://github.com/observatorium/api/pull/198/commits/0b6000559c7d58f91e53a48627eef6fda60f87dc

test/e2e/rules_test.go
jessicalins: good idea! added here: https://github.com/observatorium/api/pull/198/commits/7d50a67a817c4ab7cf7d3717acdfb7c6dc7e871c

test/e2e/configs.go
ianbillett: Seems to be fixed now :shrug:

test/e2e/rules_test.go
```diff
@@ -0,0 +1,173 @@
...
+	rulesEndpointURL := "https://" + api.Endpoint("https") + "/api/metrics/v1/"+defaultTenantName+"/api/v1/rules/raw"
+	tr := &http.Transport{
+		TLSClientConfig: getTLSClientConfig(t, e),
+	}
+
+	client := &http.Client{
+		Transport: &tokenRoundTripper{rt: tr, token: token},
+	}
+
+	t.Run("read-write-recording-rules", func(t *testing.T) {
```
ianbillett, line 43: Sorry to be a [pedant](https://en.wikipedia.org/wiki/Pedant) with this - but the actual flow is more `write-then-read-recording-rules` in the test happy path which would be the most accurate description. Not a blocker.

test/e2e/rules_test.go
```diff
@@ -0,0 +1,173 @@
...
+		testutil.Equals(t, http.StatusOK, res.StatusCode)
+
+		body, err := ioutil.ReadAll(res.Body)
+		bodyStr := string(body)
+
+		assertResponse(t, bodyStr, "sum by (job) (http_inprogress_requests)")
+		assertResponse(t, bodyStr, "tenant_id: "+defaultTenantID)
+	})
+
+	t.Run("read-write-alerting-rules", func(t *testing.T) {
```
ianbillett, line 90: Ditto - `write-then-read-alerting-rules` would be my first choice but not a blocker.

test/e2e/configs.go
jessicalins: you mean the whitespaces or `make format`?
To fix the whitespaces I ran `gofmt` manually, but I thought would be fixed by running `make format` instead (which does not seem to be the case) :(

test/e2e/rules_test.go
```diff
@@ -0,0 +1,173 @@
...
+		// Set an invalid rules file
+		invalidRules := []byte(invalidRulesYamlTpl)
+		r, err := http.NewRequest(
+			http.MethodPut,
+			rulesEndpointURL,
+			bytes.NewReader(invalidRules),
+		)
+		testutil.Ok(t, err)
+		res, err := client.Do(r)
+		//TODO: an error/http status code is not being returned to the API
```
ianbillett, line 169: If we're not fixing this issue in this PR then please let's create a ticket and reference it so we don't loose sight of this problem :)

` TODO (MON-XXXX) - an error/http...`

ianbillett (dismissed):
Couple of non-blocking naming nits & ticket to be created.

LGTM :tada:

ianbillett:
Also - we can see that your tests are passing in CI [here](https://app.circleci.com/pipelines/github/observatorium/api/356/workflows/ae2a86e2-b333-4b22-bbde-5905640cb5df/jobs/1700)!

test/e2e/rules_test.go
jessicalins: actually for this test case is more like `read-then-write-then-read` (because we read first, even though we have no rules submitted yet) but I'd prefer your suggestion too :)

test/e2e/rules_test.go
jessicalins: Added a bug ticket [here](https://issues.redhat.com/browse/MON-2176)

saswatamcode: This should get fixed by #218. 🙂

test/e2e/configs.go
```diff
@@ -29,13 +30,16 @@ const (
...
+	envMetricsName  = "e2e_metrics_read_write"
+	envRulesAPIName = "e2e_rules_api"
+	envLogsName     = "e2e_logs_read_write_tail"
+	envTenantsName  = "e2e_tenants"
+	envInteractive  = "e2e_interactive"
 
 	defaultTenantID = "1610b0c3-c509-4592-a256-a1871353dbfa"
 	mtlsTenantID    = "845cdfd9-f936-443c-979c-2ee7dc91f646"
+
+	defaultTenantName = "test-oidc"
```
matej-g, line 42: :+1: :+1: :+1:

test/e2e/rules_test.go
```diff
@@ -0,0 +1,158 @@
...
+		testutil.Ok(t, err)
+		testutil.Equals(t, http.StatusOK, res.StatusCode)
+
+		body, err := ioutil.ReadAll(res.Body)
+		bodyStr := string(body)
+		assertResponse(t, bodyStr, "record: job:up:avg")
+		assertResponse(t, bodyStr, "alert: ManyInstancesDown")
+		assertResponse(t, bodyStr, "tenant_id: "+defaultTenantID)
+	})
+}
```
matej-g, line 158: Nit: Missing newline

matej-g (approved):
I think we addressed everything, thanks again @jessicalins :tada:

Issue #158, opened 2021-08-04
Make gRPC interceptors to Support Authz/Authn


squat:
We can't implement the custom authz logic for Observatorum in OpenTelemetry, so we will need to proxy the authorized gRPC requests. Unfortunately, in order to provide authz, we also need authn. This means we will need to build a gRPC auth interceptor proxy into Observatorium. The proxies must be application agnostic, that is, they shouldn't know anything about the application-level proto for the endpoints to which they are forwarding. They should only look at frame headers for auth information.

esnible:
Can we close this because we now have https://github.com/observatorium/api/blob/main/authorization/grpc.go ?

matej-g:
Yes, looks like this has been satisfied, closing, thanks @esnible :+1:

PR #199, opened 2021-12-23
GRPC trace API

esnible:
Signed-off-by: Ed Snible <snible@us.ibm.com>

This PR creates an Open Telemetry gRPC endpoint for the method `/opentelemetry.proto.collector.trace.v1.TraceService/Export`.  **This proof-of-concept does not (yet) contain RBAC security or tenancy.**  This version uses https://github.com/mwitkow/grpc-proxy to proxy the gRPC without deserializing it.

Previously I created [pull 196](https://github.com/observatorium/api/pull/196) which exposed the OTLP Trace **HTTP** endpoint.  That was straightforward using the Chi-router.

The final implementation needs RBAC and tenancy.  I haven't yet found anything that is the equivalent to the `chi.Router` for gRPC.  Maybe https://github.com/grpc-ecosystem/go-grpc-middleware ?

It is too early to review this code, beyond the architectural ideas and the choice of libraries, parameter names, etc.

### DIFF
--- a/api/traces/v1/api.go
+++ b/api/traces/v1/api.go
@@ -42,7 +42,7 @@ func NewOTelConnection(write string, opts ...ClientOption) (*grpc.ClientConn, er
 	// service supporting opentelemetry.proto.collector.trace.v1.TraceService
 	level.Info(c.logger).Log("msg", "gRPC dialing OTel", "endpoint", write)
 
-	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 
 	return grpc.DialContext(ctx, write,
--- a/api/traces/v1/api.go
+++ b/api/traces/v1/api.go
@@ -53,7 +53,6 @@ func NewHandler(write string, opts ...HandlerOption) grpcproxy.StreamDirector {
 				// Create the connection lazily, when we first receive a trace to forward
 				level.Info(c.logger).Log("msg", "gRPC dialing OTel collector")
 
-				// TODO test where the keep-alive fails and the connection closes
 				conn, err = grpc.DialContext(ctx, write,
 					// Note that CustomCodec() is deprecated.  The fix for this isn't calling WithDefaultCallOptions(ForceCodec(...)) as suggested,
 					// because the codec we need to register is also deprecated.  A better fix, if Google removes
--- a/api/traces/v1/api.go
+++ b/api/traces/v1/api.go
@@ -2,7 +2,6 @@ package v1
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
@@ -44,17 +43,11 @@ func NewHandler(write string, opts ...HandlerOption) grpcproxy.StreamDirector {
 	var conn *grpc.ClientConn
 
 	director := func(ctx context.Context, fullMethodName string) (context.Context, *grpc.ClientConn, error) {
-		fmt.Printf("@@@ ecs REACHED NewHandler/director\n")
 		// example method name is /opentelemetry.proto.collector.trace.v1.TraceService/Export
 		// md will include headers and also pseudo-headers such as ":authority"
-		md, ok := metadata.FromIncomingContext(ctx)
-		if !ok {
-			return ctx, nil, status.Error(codes.Internal, "metadata failure")
-		}
+		md, _ := metadata.FromIncomingContext(ctx)
 
-		outCtx, _ := context.WithCancel(ctx) // TODO call cancel function
-		outCtx = metadata.NewOutgoingContext(outCtx, md.Copy())
-		// @@@ outCtx := metadata.NewOutgoingContext(ctx, md.Copy())
+		outCtx := metadata.NewOutgoingContext(ctx, md.Copy())
 
 		if fullMethodName == traceRoute {
 			var err error
--- a/api/traces/v1/api.go
+++ b/api/traces/v1/api.go
@@ -8,6 +8,7 @@ import (
 	grpcproxy "github.com/mwitkow/grpc-proxy/proxy"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 )
@@ -58,7 +59,7 @@ func NewHandler(write string, opts ...HandlerOption) grpcproxy.StreamDirector {
 					// because the codec we need to register is also deprecated.  A better fix, if Google removes
 					// the deprecated type, is https://github.com/mwitkow/grpc-proxy/pull/48
 					grpc.WithCodec(grpcproxy.Codec()), // nolint: staticcheck
-					grpc.WithInsecure(),               // nolint: staticcheck
+					grpc.WithTransportCredentials(insecure.NewCredentials()),
 					grpc.WithBlock())
 
 				if err == nil {
--- a/api/traces/v1/api.go
+++ b/api/traces/v1/api.go
@@ -2,18 +2,16 @@ package v1
 
 import (
 	"context"
+	"time"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	grpcproxy "github.com/mwitkow/grpc-proxy/proxy"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
-	"google.golang.org/grpc/metadata"
-	"google.golang.org/grpc/status"
 )
 
-const traceRoute = "/opentelemetry.proto.collector.trace.v1.TraceService/Export"
+const TraceRoute = "/opentelemetry.proto.collector.trace.v1.TraceService/Export"
 
 type handlerConfiguration struct {
 	logger log.Logger
@@ -30,7 +28,7 @@ func WithLogger(logger log.Logger) HandlerOption {
 }
 
 // NewHandler creates the new traces v1 handler.
-func NewHandler(write string, opts ...HandlerOption) grpcproxy.StreamDirector {
+func NewOTelConnection(write string, opts ...HandlerOption) (*grpc.ClientConn, error) {
 	c := &handlerConfiguration{
 		logger: log.NewNopLogger(),
 	}
@@ -39,43 +37,23 @@ func NewHandler(write string, opts ...HandlerOption) grpcproxy.StreamDirector {
 		o(c)
 	}
 
-	var conn *grpc.ClientConn
+	level.Info(c.logger).Log("msg", "gRPC dialing OTel collector")
 
-	director := func(ctx context.Context, fullMethodName string) (context.Context, *grpc.ClientConn, error) {
-		md, _ := metadata.FromIncomingContext(ctx)
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	defer cancel()
 
-		outCtx := metadata.NewOutgoingContext(ctx, md.Copy())
+	connOtel, err := grpc.DialContext(ctx, write,
+		// Note that CustomCodec() is deprecated.  The fix for this isn't calling WithDefaultCallOptions(ForceCodec(...)) as suggested,
+		// because the codec we need to register is also deprecated.  A better fix, is the newer
+		// version of mwitkow/grpc-proxy, but that version doesn't (currently) work with OTel protocol.
+		grpc.WithCodec(grpcproxy.Codec()), // nolint: staticcheck
+		grpc.WithTransportCredentials(insecure.NewCredentials()))
 
-		if fullMethodName == traceRoute {
-			var err error
-
-			if conn == nil {
-				// Create the connection lazily, when we first receive a trace to forward
-				level.Info(c.logger).Log("msg", "gRPC dialing OTel collector")
-
-				conn, err = grpc.DialContext(ctx, write,
-					// Note that CustomCodec() is deprecated.  The fix for this isn't calling WithDefaultCallOptions(ForceCodec(...)) as suggested,
-					// because the codec we need to register is also deprecated.  A better fix, if Google removes
-					// the deprecated type, is https://github.com/mwitkow/grpc-proxy/pull/48
-					grpc.WithCodec(grpcproxy.Codec()), // nolint: staticcheck
-					grpc.WithTransportCredentials(insecure.NewCredentials()),
-					grpc.WithBlock())
-
-				if err == nil {
-					level.Info(c.logger).Log("msg", "gRPC connected to OTel collector")
-				} else {
-					conn = nil
-					level.Warn(c.logger).Log("msg", "gRPC did not connect to OTel collector")
-				}
-			}
-
-			return outCtx, conn, err
-		}
-
-		level.Info(c.logger).Log("msg", "gRPC reverse proxy director caught unknown method", "methodName", fullMethodName)
-
-		return outCtx, nil, status.Errorf(codes.Unimplemented, "Unknown method")
+	// Note that err == nil does not mean connected; this is a non-blocking ClientConn.
+	if err != nil {
+		level.Warn(c.logger).Log("msg", "gRPC did not dial to OTel collector", "target", write)
+		return nil, err
 	}
 
-	return director
+	return connOtel, nil
 }
--- a/api/traces/v1/api.go
+++ b/api/traces/v1/api.go
@@ -1,0 +1,90 @@
+package v1
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	grpcproxy "github.com/mwitkow/grpc-proxy/proxy"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+const (
+	traceRoute = "/opentelemetry.proto.collector.trace.v1.TraceService/Export"
+)
+
+type handlerConfiguration struct {
+	logger log.Logger
+}
+
+// HandlerOption modifies the handler's configuration.
+type HandlerOption func(h *handlerConfiguration)
+
+// WithLogger add a custom logger for the handler to use.
+func WithLogger(logger log.Logger) HandlerOption {
+	return func(h *handlerConfiguration) {
+		h.logger = logger
+	}
+}
+
+// NewHandler creates the new traces v1 handler.
+func NewHandler(write string, opts ...HandlerOption) grpcproxy.StreamDirector {
+	c := &handlerConfiguration{
+		logger: log.NewNopLogger(),
+	}
+
+	for _, o := range opts {
+		o(c)
+	}
+
+	var conn *grpc.ClientConn
+
+	director := func(ctx context.Context, fullMethodName string) (context.Context, *grpc.ClientConn, error) {
+		fmt.Printf("@@@ ecs REACHED NewHandler/director\n")
+		// example method name is /opentelemetry.proto.collector.trace.v1.TraceService/Export
+		// md will include headers and also pseudo-headers such as ":authority"
+		md, ok := metadata.FromIncomingContext(ctx)
+		if !ok {
+			return ctx, nil, status.Error(codes.Internal, "metadata failure")
+		}
+
+		outCtx, _ := context.WithCancel(ctx) // TODO call cancel function
+		outCtx = metadata.NewOutgoingContext(outCtx, md.Copy())
+		// @@@ outCtx := metadata.NewOutgoingContext(ctx, md.Copy())
+
+		if fullMethodName == traceRoute {
+			var err error
+
+			if conn == nil {
+				// Create the connection lazily, when we first receive a trace to forward
+				level.Info(c.logger).Log("msg", "gRPC dialing OTel collector")
+
+				conn, err = grpc.DialContext(ctx, write,
+					// Note that CustomCodec() is deprecated.  The fix for this isn't calling WithDefaultCallOptions(ForceCodec(...)) as suggested,
+					// because the codec we need to register is also deprecated.  A better fix, if Google removes
+					// the deprecated type, is https://github.com/mwitkow/grpc-proxy/pull/48
+					grpc.WithCodec(grpcproxy.Codec()), // nolint: staticcheck
+					grpc.WithInsecure(),
+					grpc.WithBlock())
+
+				if err == nil {
+					level.Info(c.logger).Log("msg", "gRPC connected to OTel collector")
+				} else {
+					level.Warn(c.logger).Log("msg", "gRPC did not connect to OTel collector")
+				}
+			}
+
+			return outCtx, conn, err
+		}
+
+		level.Info(c.logger).Log("msg", "gRPC reverse proxy director caught unknown method", "methodName", fullMethodName)
+
+		return outCtx, nil, status.Errorf(codes.Unimplemented, "Unknown method")
+	}
+
+	return director
+}
--- a/api/traces/v1/api.go
+++ b/api/traces/v1/api.go
@@ -11,25 +11,26 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 )
 
+// TraceRoute represents the fully-qualified gRPC method name for exporting a trace
 const TraceRoute = "/opentelemetry.proto.collector.trace.v1.TraceService/Export"
 
-type handlerConfiguration struct {
+type connOptions struct {
 	logger log.Logger
 }
 
-// HandlerOption modifies the handler's configuration.
-type HandlerOption func(h *handlerConfiguration)
+// ClientOption modifies the connection's configuration.
+type ClientOption func(h *connOptions)
 
 // WithLogger add a custom logger for the handler to use.
-func WithLogger(logger log.Logger) HandlerOption {
-	return func(h *handlerConfiguration) {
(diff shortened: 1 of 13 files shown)


go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
 	github.com/metalmatze/signal v0.0.0-20210307161603-1c9aa721a97a
 	github.com/mitchellh/mapstructure v1.4.1
+	github.com/mwitkow/grpc-proxy v0.0.0-20181017164139-0f1106ef9c76
```
pavolloffay, line 25: is this a safe to use? Last commit is from October 2018

esnible: @pavolloffay I had the same fear.  Reasons to keep it for now:
- I saw commits there from @bwplotka , so probably approves of mwitkow's approach
- I couldn't find anything else on github which forwards without deserializing and reserializing
- Because I have read mwitkow's code I can't recreate it myself (for legal reasons; anything I make might be seen as derivative to mwitkow's work).

matej-g:
Thanks for this @esnible :+1: I've pinged people to take a look at this, I'll be back with my review as well.

esnible:
@matej-g Any feedback?

matej-g:
Hey @esnible,
Although there seem to be still a lot of comments and to-do's, I believe this is in essence how we can proceed, i.e. I imagine we want to have a proxy to forward requests and before that we'll need interceptors for every authn / authz (OIDC, OPA etc.). I'd personally say that we can develop this PR further and continue iterate / give feedback? I'm happy to also chat about this on the tomorrow's community call.

I believe @squat could also give us qualified feedback here, as he was initially having thoughts about how to implement the gRPC interceptors (https://github.com/observatorium/api/issues/158).

go.mod
matej-g: Yes, it seems there are not many alternatives (short of creating our own proxy implementation). I also found https://github.com/bradleyjkemp/grpc-tools which seem to be maintained a bit better, but I have no idea how it compares to this package.

bwplotka:
Thanks for this and make sure to ping us on the #observatorium CNCF channel or through private means next - that got blocked too long 🙈 

I would love if we can add gRPC API if Otel gRPC is more efficient and popular. That should fit perfectly in our space.

In terms of  https://github.com/mwitkow/grpc-proxy feel free to use it. I have somehow written permissions and looks like many wants to jump in and help: https://github.com/mwitkow/grpc-proxy/pull/52. Definitely something running in production in many places. (: 

In terms of interceptors please use `v2` branch and if anything is missing, feel free to put issue on  https://github.com/grpc-ecosystem/go-grpc-middleware/

For authorization - this interceptor might be a good start: https://github.com/grpc-ecosystem/go-grpc-middleware/blob/v2/interceptors/auth/auth.go We might want to convert some middleware to interceptors:  https://github.com/observatorium/api/blob/main/main.go#L552 

Overall all decisions here makes sense to me! 💪🏽

esnible:
**NOT READY TO REVIEW.** The code stopped working after the rebase; investigating.

matej-g:
> Note: This code requires the 2018 version of mwitkow/grpc-proxy. The repo was untouched; I thought it was orphaned, but just a week ago it got an update that breaks this. I have opened an issue [mwitkow/grpc-proxy#54](https://github.com/mwitkow/grpc-proxy/issues/54)

What seems to be the issue exactly with using latest version? On a quick look, the changes seem to be more on the surface to modernize the codebase, rather than changing how the library works.

My thinking is, if we're already introducing this dependency and it has been just updated, I would not make these change to require to use the old version.

authentication/oidc.go
```diff
@@ -367,3 +372,97 @@ func (a oidcAuthenticator) Middleware() Middleware {
...
+		var token string
+
+		md, ok := metadata.FromIncomingContext(ctx)
+		if !ok {
+			return ctx, status.Error(codes.Internal, "metadata error")
+		}
+
+		authorizationHeader := ""
+		authorizationHeaders := md["authorization"]
+		if len(authorizationHeaders) > 0 {
```
matej-g, line 388: Could we compress this into one `if` block in which we do the whole `token` population logic?
```suggestion
		if len(md["authorization"]) > 0 && md["authorization"][0] {
```

authentication/oidc.go
```diff
@@ -367,3 +372,97 @@ func (a oidcAuthenticator) Middleware() Middleware {
...
+			if len(authorization) != 2 {
+				const msg = "invalid Authorization header"
+				level.Debug(a.logger).Log("msg", msg)
+				return ctx, status.Error(codes.InvalidArgument, msg)
+			}
+
+			token = authorization[1]
+		}
+
+		idToken, err := a.verifier.Verify(oidc.ClientContext(ctx, a.client), token)
```
matej-g, line 403: It looks like from this line below, the code is very similar with how HTTP middleware handles this. Do you think we could extract those parts and reuse a single method for both HTTP and gRPC middleware?

authentication/mtls.go
```diff
@@ -143,6 +147,12 @@ func (a MTLSAuthenticator) Middleware() Middleware {
 	}
 }
 
+func (a MTLSAuthenticator) GRPCMiddleware() grpc.StreamServerInterceptor {
+	return grpc_middleware_auth.StreamServerInterceptor(func(ctx context.Context) (context.Context, error) {
```
matej-g, line 151: Should we not support mTLS authentication as well? Or you'd like to this in a separate PR? Maybe we could at least at a to-do.

main.go
```diff
@@ -814,6 +858,12 @@ func parseFlags() (config, error) {
 		"The name of the HTTP header containing the tenant ID to forward to the metrics upstreams.")
 	flag.StringVar(&cfg.metrics.tenantLabel, "metrics.tenant-label", "tenant_id",
 		"The name of the PromQL label that should hold the tenant ID in metrics upstreams.")
+	flag.StringVar(&rawTracesWriteEndpoint, "traces.write.endpoint", "",
+		"The endpoint against which to make gRPC write requests for traces.")
+	flag.StringVar(&cfg.traces.upstreamCAFile, "traces.tls.ca-file", "",
+		"File containing the TLS CA against which to upstream OTLP trace servers. Leave blank to disable TLS.")
+	flag.StringVar(&cfg.traces.tenantHeader, "traces.tenant-header", "X-Tenant",
+		"The name of the HTTP header containing the tenant ID to forward to the logs upstream.")
```
matej-g, line 866: Seems like also need to regenerate README to add these options, it seem to be causing CI to fail currently.

matej-g (commented):
Thanks for this @esnible! I think the direction is looking good.

It would also be great to have E2E tests for tracing API, but I'm happy to have them after the fact (as long as we don't forget about it :slightly_smiling_face:).

esnible:
@matej-g I have spent two days trying to get the latest version to proxy OpenTelemetry protobufs through gRPC.  Sending data either with _grpcurl_ or with the Open Telemetry gRPC trace exporter gives similar problems.  I discuss the problem at https://github.com/mwitkow/grpc-proxy/issues/55

I am **extremely** concerned about requiring a back-level mwitkow/grpc-proxy.  I am looking for a work-around.

> What seems to be the issue exactly with using latest version? On a quick look, the changes seem to be more on the surface to modernize the codebase, rather than changing how the library works.
> 
> My thinking is, if we're already introducing this dependency and it has been just updated, I would not make these changes to require to use the old version.

matej-g:
> @matej-g I have spent two days trying to get the latest version to proxy OpenTelemetry protobufs through gRPC. Sending data either with _grpcurl_ or with the Open Telemetry gRPC trace exporter gives similar problems. I discuss the problem at [mwitkow/grpc-proxy#55](https://github.com/mwitkow/grpc-proxy/issues/55)

Understood, thanks for putting in the effort @esnible 👍, then let's see if get an answer from upstream and in the meantime we can perhaps settle the remaining questions.

api/traces/v1/api.go
```diff
@@ -0,0 +1,83 @@
...
+			if conn == nil {
+				// Create the connection lazily, when we first receive a trace to forward
+				level.Info(c.logger).Log("msg", "gRPC dialing OTel collector")
+
+				conn, err = grpc.DialContext(ctx, write,
+					// Note that CustomCodec() is deprecated.  The fix for this isn't calling WithDefaultCallOptions(ForceCodec(...)) as suggested,
+					// because the codec we need to register is also deprecated.  A better fix, if Google removes
+					// the deprecated type, is https://github.com/mwitkow/grpc-proxy/pull/48
+					grpc.WithCodec(grpcproxy.Codec()), // nolint: staticcheck
+					grpc.WithInsecure(),               // nolint: staticcheck
```
frzifus, line 64:
```suggestion
					grpc.WithTransportCredentials(insecure.NewCredentials())
```
secondary, but so the linter would give us a green light.

main.go
```diff
@@ -440,15 +456,15 @@ func main() {
...
-			pm := authentication.NewProviderManager(logger, registerTenantsFailingMetric)
+		tenantIDs := map[string]string{}
+		authorizers := map[string]rbac.Authorizer{}
+		var rateLimits []ratelimit.Config
+		// registrationRetryCount used by authenticator providers to count
+		// registration failures per tenant.
+		registerTenantsFailingMetric := authentication.RegisterTenantsFailingMetric(reg)
+		pm := authentication.NewProviderManager(logger, registerTenantsFailingMetric)
 
+		r.Group(func(r chi.Router) {
```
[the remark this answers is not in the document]
esnible: > Looks like this could be moved even further down around line 515, since the tenant ID population logic etc. is not related to the router.

@matej-g `main()` is 900 lines long!  I am nervous about refactoring it as part of this PR.  You missed the `r` at line 505.  It wasn't obvious to me how the looping through `tenant` and pushing the `r.Group()` relate.  I think the refactor should be done -- other people will also struggle with this code -- but perhaps not as part of my first PR.

main.go
```diff
@@ -974,3 +1043,40 @@ type otelErrorHandler struct {
...
+		grpc.CustomCodec(grpcproxy.Codec()), // nolint: staticcheck
+		grpc.UnknownServiceHandler(grpcproxy.TransparentHandler(director)),
+		// Authorization (tenant)
+		grpc.ChainStreamInterceptor(
+			authentication.WithGRPCTenantHeader(tenantHeader, tenantIDs),
+			authentication.WithGRPCAccessToken(),
+			authentication.WithGRPCTenantInterceptors(pm.GRPCMiddlewares),
+			auth.StreamServerInterceptor(
+				authorization.WithGRPCAuthorizers(authorizers, rbac.Write, "traces", logger)),
+		),
```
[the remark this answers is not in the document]
esnible: Agreed that we **need** rate limiting to meet our goal of stopping DDOS and customer abuse.
Agreed about the desirability of metrics.

My thought was that the first GRPC PR should be minimal for ease of reviewing it and to unblock @pavolloffay so that he can test the Jaeger backend with the Observatorium headers.

api/traces/v1/api.go
```diff
@@ -0,0 +1,83 @@
...
+		// example method name is /opentelemetry.proto.collector.trace.v1.TraceService/Export
+		// md will include headers and also pseudo-headers such as ":authority"
+		md, _ := metadata.FromIncomingContext(ctx)
+
+		outCtx := metadata.NewOutgoingContext(ctx, md.Copy())
+
+		if fullMethodName == traceRoute {
+			var err error
+
+			if conn == nil {
```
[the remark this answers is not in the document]
esnible: @matej-g I am now explicitly setting `conn` to `nil` on failure.

I haven't tested this code yet for performance or reliability.  It passes my simple test: Observatorium-api forwards a trace.  Even if the OTel collector is restarted, and the Observatorium-api forwards the next trace.  More testing is needed for correct keep-alives and for the OTel collector restarting while Observatorium-api is streaming traces.

esnible:
@matej-g I would also like to have tests.  I see tests in the _test/e2e_ directory.  I did not understand them.  Can you point me to some docs on test writing?

I have some ideas on how to test the GRPC piece end-to-end with just 2 Open Telemetry agents -- no actual Jaeger or span storage.  Would this be desirable?  Or should the E2E tests install the full Jaeger, possibly including ElasticSearch, and be as close as possible to a valid deployment?

> It would also be great to have E2E tests for tracing API, but I'm happy to have them after the fact (as long as we don't forget about it 🙂).

esnible:
@matej-g I discovered a bug where the Observatorium-api gets an interrupt and restarts every few minutes.  I am looking into it; please review but don't merge yet.

main.go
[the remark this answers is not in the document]
matej-g: You're right, this requires more thorough refactoring and that's beyond this PR, let's leave it aside.

main.go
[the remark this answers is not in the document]
matej-g: As long as we keep track of it, I'm happy to do it separately. Let's do that to keep this moving then.

matej-g:
> @matej-g I would also like to have tests. I see tests in the _test/e2e_ directory. I did not understand them. Can you point me to some docs on test writing?

We are using the https://github.com/efficientgo/e2e framework, you'll find everything documented there.

main.go
```diff
@@ -974,3 +1043,85 @@ type otelErrorHandler struct {
...
+
+// Permissions required for each gRPC method
+//nolint:gochecknoglobals (this would be a const if Golang allowed const maps)
+var gRPCRBAC = authorization.GRPCRBac{
+	// opentelemetry.proto.collector.trace.v1.TraceService/Export requires trace write perm
+	tracesv1.TraceRoute: {
+		Permission: rbac.Write,
+		Resource:   "traces",
+	},
+	// in the future we will add trace read permission for Jaeger queries, etc.
```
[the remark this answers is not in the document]
bwplotka: I know there is make lint issue - I guess it just ask for issue number?

main.go
```diff
@@ -974,3 +1043,85 @@ type otelErrorHandler struct {
...
+}
+
+//nolint:lll
```
bwplotka, line 1058: lll?

main.go
```diff
@@ -974,3 +1043,85 @@ type otelErrorHandler struct {
...
+func gRPCServer(cfg *config, tenantHeader string, tenantIDs map[string]string, pmis authentication.GRPCMiddlewareFunc, authorizers map[string]rbac.Authorizer, logger log.Logger) (*grpc.Server, error) {
+	connOtel, err := tracesv1.NewOTelConnection(cfg.traces.writeEndpoint,
+		tracesv1.WithLogger(logger),
+	)
```
bwplotka, line 1062:
```suggestion
	connOtel, err := tracesv1.NewOTelConnection(
	        cfg.traces.writeEndpoint,
		tracesv1.WithLogger(logger),
	)
```

main.go
```diff
@@ -974,3 +1043,85 @@ type otelErrorHandler struct {
...
+	// opentelemetry.proto.collector.trace.v1.TraceService/Export requires trace write perm
+	tracesv1.TraceRoute: {
+		Permission: rbac.Write,
+		Resource:   "traces",
+	},
+	// in the future we will add trace read permission for Jaeger queries, etc.
+}
+
+//nolint:lll
+func gRPCServer(cfg *config, tenantHeader string, tenantIDs map[string]string, pmis authentication.GRPCMiddlewareFunc, authorizers map[string]rbac.Authorizer, logger log.Logger) (*grpc.Server, error) {
```
bwplotka, line 1059:
```suggestion
func newOtelGRPCServer(cfg *config, tenantHeader string, tenantIDs map[string]string, pmis authentication.GRPCMiddlewareFunc, authorizers map[string]rbac.Authorizer, logger log.Logger) (*grpc.Server, error) {
```

main.go
```diff
@@ -974,3 +1043,85 @@ type otelErrorHandler struct {
...
+	connOtel, err := tracesv1.NewOTelConnection(cfg.traces.writeEndpoint,
+		tracesv1.WithLogger(logger),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// Currently we only proxy TraceService/Export to OTel collectors.
+	// In the future we will pass queries to Jaeger, and possibly other
+	// gRPC methods for logs and metrics to different connections
```
bwplotka, line 1069:
```suggestion
	// gRPC methods for logs and metrics to different connections.
```

main.go
```diff
@@ -974,3 +1043,85 @@ type otelErrorHandler struct {
...
+		return outCtx, nil, status.Errorf(codes.Unimplemented, "Unknown method")
+	}
+
+	opts := []grpc.ServerOption{
+		// Note that CustomCodec() is deprecated.  The fix for this isn't calling RegisterCodec() as suggested,
+		// because the codec we need to register is also deprecated.  A better fix, if Google removes
+		// the deprecated type, is https://github.com/mwitkow/grpc-proxy/pull/48
+		grpc.CustomCodec(grpcproxy.Codec()), // nolint: staticcheck
+		grpc.UnknownServiceHandler(grpcproxy.TransparentHandler(director)),
+		// Authorization (tenant)
```
bwplotka, line 1096: I think it's visible what interceptors are, we could remove this comment.

main.go
```diff
@@ -974,3 +1043,85 @@ type otelErrorHandler struct {
...
+		grpc.ChainStreamInterceptor(
+			authentication.WithGRPCTenantHeader(tenantHeader, tenantIDs, logger),
+			authentication.WithGRPCAccessToken(),
+			authentication.WithGRPCTenantInterceptors(logger, pmis),
+			auth.StreamServerInterceptor(
+				authorization.WithGRPCAuthorizers(authorizers, gRPCRBAC, logger)),
+			// Log when a trace cannot be forwarded
```
bwplotka, line 1103: .. why?

bwplotka (commented):
Thanks

Overall, functionally makes sense. Let us know if you need help in setting e2e tests. Code looks fine too, just make sure to not handle error twice. If you see something like 

```
foo, err := bar()
if err != nil {
  // produce error/warning/debug log line
  return err
}
```

That is generally quite bad. The reason is that the passed error is *most likely* handled in some way - this some way is usually to log it on server side, put metric, return error to user. If every function would add log line when returning error we have lots of wasted information and effort. 

HTTP middleware are tricky as usually indeed the returned error is handled by returning HTTP/gRPC response to client/user. But in order to log this I would rather implement logging middleware/interceptor than spending 100 code lines to add log line to every error handling. WDYT?

api/traces/v1/api.go
```diff
@@ -0,0 +1,59 @@
...
+
+	level.Info(c.logger).Log("msg", "gRPC dialing OTel collector", "endpoint", write)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	defer cancel()
+
+	connOtel, err := grpc.DialContext(ctx, write,
+		// Note that CustomCodec() is deprecated.  The fix for this isn't calling WithDefaultCallOptions(ForceCodec(...)) as suggested,
+		// because the codec we need to register is also deprecated.  A better fix, is the newer
+		// version of mwitkow/grpc-proxy, but that version doesn't (currently) work with OTel protocol.
```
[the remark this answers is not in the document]
esnible: @bwplotka already done; the Issue is https://github.com/mwitkow/grpc-proxy/issues/55

api/traces/v1/api.go
```diff
@@ -0,0 +1,59 @@
...
+		grpc.WithCodec(grpcproxy.Codec()), // nolint: staticcheck
+		grpc.WithTransportCredentials(insecure.NewCredentials()))
+
+	// Note that err == nil does not mean connected; this is a non-blocking ClientConn.
+	if err != nil {
+		level.Warn(c.logger).Log("msg", "gRPC did not dial to OTel collector", "target", write)
```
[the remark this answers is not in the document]
esnible: In this case two parties are interested in the error.  The client sending traces to Observatorium will want to know the trace was not preserved.  The operators of Observatorium will want to know they are dropping traces sent by clients.
The former look at the return; the latter look at Observatorium logs.

authentication/oidc.go
```diff
@@ -298,72 +303,134 @@ func (a oidcAuthenticator) Middleware() Middleware {
...
+	}
+
+	sub := idToken.Subject
+
+	if a.config.UsernameClaim != "" {
+		claims := map[string]interface{}{}
+		if err := idToken.Claims(&claims); err != nil {
+			const msg = "failed to read claims"
+
+			level.Warn(a.logger).Log("msg", msg, "err", err)
```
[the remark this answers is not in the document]
esnible: This is not new code; I merely refactored existing code.  The error message will be important for debugging the interaction between Observatorium and OIDC provider, and must not be given to the caller.

api/traces/v1/api.go
```diff
@@ -0,0 +1,54 @@
...
+
+	for _, o := range opts {
+		o(c)
+	}
+
+	// The endpoint is typically an OTel collector, but can be any gRPC
+	// service supporting opentelemetry.proto.collector.trace.v1.TraceService
+	level.Info(c.logger).Log("msg", "gRPC dialing OTel", "endpoint", write)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
```
matej-g, line 45: Any reason for the 1 minute time out here? Could we do with shorter time out?

authentication/grpc.go
```diff
@@ -0,0 +1,91 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+// GRPCMiddlewareFunc is a function type able to return authentication middleware for
+// a given tenant. If no middleware is found, the second return value should be false.
+type GRPCMiddlewareFunc func(tenant string) (grpc.StreamServerInterceptor, bool)
+
+// WithGRPCTenantHeader returns a new StreamServerInterceptor that adds the tenant and tenantid
```
matej-g, line 21:
```suggestion
// WithGRPCTenantHeader returns a new StreamServerInterceptor that adds the tenant and tenantID
```

authentication/grpc.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		tenant := headerTenants[0]
+		ctx = context.WithValue(ctx, tenantKey, tenant)
+
+		id, ok := tenantIDs[tenant]
+		if !ok {
+			// Rather than letting the caller know their tenant was invalid, just record that it was invalid
+			id = "__no_tenant_id"
+		}
```
matej-g, line 48: What is the rationale behind this instead of returning error?

authentication/grpc.go
```diff
@@ -0,0 +1,91 @@
...
+
+		return handler(srv, wrapped)
+	}
+}
+
+func WithGRPCAccessToken() grpc.StreamServerInterceptor {
+	return grpc_middleware_auth.StreamServerInterceptor(func(ctx context.Context) (context.Context, error) {
+		md, ok := metadata.FromIncomingContext(ctx)
+		if !ok {
+			return nil, status.Error(codes.Internal, "metadata error")
```
matej-g, line 62: Nit: perhaps we could align the error message with the other interceptor (if we switch to 'not found' code there, it should also apply to this)

matej-g (commented):
Thanks for addressing my comments @esnible, this is looking nicely :+1:, I have few more comments but I would not like to stall this PR any further. I'll be happy to approve once you get a chance to respond to the latest review.

authentication/grpc.go
```diff
@@ -0,0 +1,91 @@
...
+// WithGRPCTenantHeader returns a new StreamServerInterceptor that adds the tenant and tenantid
+// to the stream Context.
+func WithGRPCTenantHeader(header string, tenantIDs map[string]string, logger log.Logger) grpc.StreamServerInterceptor {
+	header = strings.ToLower(header)
+
+	return func(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+		ctx := ss.Context()
+		md, ok := metadata.FromIncomingContext(ctx)
+		if !ok {
+			return status.Errorf(codes.Internal, "metadata not found")
```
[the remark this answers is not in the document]
esnible: No.  Not having metadata should be impossible.

authentication/grpc.go
esnible: Normally for security if someone doesn't have a valid `Authorization` we don't give them any hints about their message.  Such as "invalid tenant ID", which would let an unauthorized user check if a tenant existed.

Later I realized that the Observatorium token validation endpoint is different for each tenant, thus it is impossible to authorize without a valid tenant header.  Thus not failing here doesn't provide the additional security I wanted -- a user with an invalid bearer JWT can check if tenants exist.

main.go
```diff
@@ -128,6 +139,13 @@ type logsConfig struct {
 	enabled bool
 }
 
+type tracesConfig struct {
+	writeEndpoint string
+	tenantHeader  string
+	// enable traces if writeEndpoint is provided.
+	enabled bool
```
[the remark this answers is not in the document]
esnible: I added a TODO.  There is an argument that if we support neither metrics nor logs we shouldn't listen on HTTP at all, and I want to avoid the changes needed for that with this initial PR.

authentication/grpc.go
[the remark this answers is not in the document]
matej-g: I see now why it has to be there :+1: I retract the comment.

authentication/grpc.go
matej-g: Got it, thanks for the explanation :+1:

main.go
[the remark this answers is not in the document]
matej-g: Cool, makes sense :+1:

matej-g (approved):
Thanks you @esnible! This is look good to me now!

api/traces/v1/api.go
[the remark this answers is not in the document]
bwplotka: Thanks great job! Putting issue here would help to link things

bwplotka (approved):
Thanks! It looks definitely good enough 👍🏽

PR #200, opened 2022-01-04
rules spec: Add an endpoint to return rules for all tenants

onprem:
Following up on the discussion [here](https://github.com/observatorium/thanos-rule-syncer/pull/8#pullrequestreview-841868405) this PR extends the rules backend spec with another endpoint to return rules for all tenants.

I also changes the previous path to return rules for a tenant from `/api/v1/{tenant}/rules` to `/api/v1/rules/{tenant}`, because the previous format scoped everything under this API to a tenant.

Signed-off-by: Prem Saraswat <prmsrswt@gmail.com>

### DIFF
--- a/rules/rules.go
+++ b/rules/rules.go
@@ -294,11 +294,26 @@ func WithRequestEditorFn(fn RequestEditorFn) ClientOption {
 
 // The interface specification for the client above.
 type ClientInterface interface {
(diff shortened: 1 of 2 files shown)


squat (approved):
Looks nice, @onprem

PR #203, opened 2022-01-13
Introduce dependabot config

tareqmamari:
This would help speed up bumping dependencies to avoid any
vulnerabilities.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+- package-ecosystem: gomod
+  directory: "/"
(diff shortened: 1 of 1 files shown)


tareqmamari:
As a replacement of this PR, could we use renovate bot? https://github.com/marketplace/renovate
@observatorium-maintainers

kakkoyun (approved):
LGTM

matej-g (approved):
Looks good, I agree with @kakkoyun - let's get Dependabot in and we can reevaluate if we see the need to change to some other solution

PR #204, opened 2022-01-17
Add matcher based tenancy to api/v1/rules

saswatamcode:
Currently, a call to the `/api/metrics/v1/{tenant}/api/v1/rules` endpoint, returns rules for all tenants, as it proxies the request to Thanos Querier.

This PR adds tenant isolation to `/api/metrics/v1/{tenant}/api/v1/rules` endpoint, by adapting [`prom-label-proxy`](https://github.com/prometheus-community/prom-label-proxy) code to modify the API response from the Thanos Query `api/v1/rules` endpoint and filter via the tenant label (this is possible as we [enforce tenant labels on rules](https://github.com/observatorium/api/pull/191)). So each tenant can only retrieve their own rules.

Relevant issue: https://issues.redhat.com/browse/MON-2141

### DIFF
--- a/api/metrics/v1/http.go
+++ b/api/metrics/v1/http.go
@@ -32,7 +32,8 @@ const (
 	labelNamesRoute  = "/api/v1/labels"
 	labelValuesRoute = "/api/v1/label/{label_name}/values"
 	receiveRoute     = "/api/v1/receive"
-	rulesRoute       = "/api/v1/rules/raw"
+	rulesRoute       = "/api/v1/rules"
+	rulesRawRoute    = "/api/v1/rules/raw"
 )
 
 type handlerConfiguration struct {
@@ -214,6 +215,12 @@ func NewHandler(read, write, rulesEndpoint *url.URL, upstreamCA []byte, opts ...
 				prometheus.Labels{"group": "metricsv1", "handler": "label_values"},
 				otelhttp.WithRouteTag(c.spanRoutePrefix+labelValuesRoute, proxyRead),
 			))
+			// Thanos Query Rules API supports matchers from v0.25 so the WithEnforceTenancyOnMatchers
+			// middleware will not work here if prior versions are used.
+			r.Handle(rulesRoute, c.instrument.NewHandler(
+				prometheus.Labels{"group": "metricsv1", "handler": "rules"},
+				otelhttp.WithRouteTag(c.spanRoutePrefix+rulesRoute, proxyRead),
+			))
 		})
 
 		var uiProxy http.Handler
@@ -296,12 +303,12 @@ func NewHandler(read, write, rulesEndpoint *url.URL, upstreamCA []byte, opts ...
 
 		r.Group(func(r chi.Router) {
 			r.Use(c.uiMiddlewares...)
-			r.Get(rulesRoute, rh.get)
+			r.Get(rulesRawRoute, rh.get)
 		})
 
 		r.Group(func(r chi.Router) {
 			r.Use(c.writeMiddlewares...)
-			r.Put(rulesRoute, rh.put)
+			r.Put(rulesRawRoute, rh.put)
 		})
 	}
 


squat:
Hey @saswatamcode, regarding:
> a call to the /api/metrics/v1/{tenant}/api/v1/rules endpoint, returns rules for all tenants

I don't quite agree with this statement. I think the ticket that you linked in the PR description is incorrect. The rules API is already tenant-aware: the `/rules` API is filtering results so that we only get the rules that were defined by one single tenant. Furthermore, the client must be authorized to read metrics from the specified tenant in order to fetch the rules. Please look here: https://github.com/observatorium/api/blob/9401841557797afdcfde05800c552158cc36a483/api/metrics/v1/rules.go#L34

However, as I mentioned in #191 
> Note: this is a distinct problem from ensuring that the rules written by
a tenant are guaranteed to only include data from that tenant.
Currently, the rules from a tenant can select data from any tenant. This
is a distinct problem and should absolutely be tackled in a follow up.

I think this is the problem you, and the Jira ticket, are trying to address?

saswatamcode:
Hi @squat! :)
Yes, the naming is a bit confusing, as both deal with rules, but let me clarify the purpose of this ticket/PR,
- Currently, the rules API which hits the Rules storage backend is tenant-aware. A tenant can get and set rules in YAML format, via GET and PUT requests to the `api/metrics/v1/{tenant}/api/v1/rules/raw` endpoint of the API.
- However when a tenant sends a GET request to `api/metrics/v1/{tenant}/api/v1/rules`, the request is proxied by the API to the read upstream endpoint (for eg. Thanos Querier). The querier sends back all rules of all tenants in JSON. This is handled using [`uiProxy`](https://github.com/observatorium/api/blob/main/api/metrics/v1/http.go#L241).

> Note: this is a distinct problem from ensuring that the rules written by
a tenant are guaranteed to only include data from that tenant.
Currently, the rules from a tenant can select data from any tenant. This
is a distinct problem and should absolutely be tackled in a follow up.

I think this is the problem you, and the Jira ticket, are trying to address?

This is still a distinct problem.

squat:
ack. I wasn't aware of this second endpoint. hmm this endpoint seems to be essentially part of an attempt trying to turn Observatorium into an equivalent Prometheus/Thanos Querier replacement

ianbillett:
Nice PR @saswatamcode :fist_raised: I think the changes you have proposed would work, but I am wondering if there is a different solution that better maintains existing properties of the API :thinking: 

As @squat alludes to - the API was always designed and intended to be a light-weight authentication and authorization proxy that enforced the relevant tenant configurations on requests. With this in mind, it feels like adding rule specific logic in this PR start to take us down a path we have tried our best to resist so far!

Did you consider any upstream / Thanos first approaches? For example if the Thanos Query [rules API](https://github.com/thanos-io/thanos/blob/e982ff899a54d888e28efa00cd5b7caf4643ea6a/pkg/api/query/v1.go#L183) supported matchers (like [/labels](https://github.com/thanos-io/thanos/blob/e982ff899a54d888e28efa00cd5b7caf4643ea6a/pkg/api/query/v1.go#L178) does), then the API would then append the right labels and proxy the request forward.

This would be more work (requires change in Rules [proto](https://github.com/thanos-io/thanos/blob/28b66579c7ecdfdd4db5f8ba4b598ac673e6c89c/pkg/rules/rulespb/rpc.pb.go#L108)), but feels like it would maintain the lightweight API properties and could also be useful for Thanos.

saswatamcode:
Hi @ianbillett! 
Yes, I did consider the upstream first approach before, while discussing this with @onprem. 🙂

Adding matcher support to Thanos Query rules API is certainly possible but I was a bit hesitant to go with that, as this would make the Query rules API diverge from the [Prometheus rules API](https://prometheus.io/docs/prometheus/latest/querying/api/#rules) (one would have matcher support while the other wouldn't). 

Also, I found precedent that bits of prom-label-proxy were already adapted in Observatorium API [here](https://github.com/observatorium/api/blob/main/api/metrics/v1/labels_enforcer.go#L27), so I decided to go with this.

If the upstream approach seems better for this use case, I can raise a PR to Thanos and update this PR once merged. WDYT? 🙂

bwplotka:
SGTM for changing Prom and Thanos API first (:

Sounds like this is a great starting point: https://github.com/prometheus/prometheus/issues/9627

ianbillett:
We just discussed this issue at the Thanos Community hours [meeting](https://docs.google.com/document/d/137XnxfOT2p1NcNUq6NWZjwmtlSdA6Wyti86Pd6cyQhs/edit#heading=h.sxuklvmkmbax) :+1: 

We're happy with the idea of adding matchers to the Thanos `/rules` calls. Though this would make Thanos differ from the Prometheus API that is ok, as long as all valid Prometheus requests are still compatible. In fact, there is an upstream Prometheus issue https://github.com/prometheus/prometheus/issues/9627 that is proposing the same thing.

Next steps are to follow up with the upstream issue first, and if we can't progress with that then add it in Thanos :+1:

squat:
one thing I'm curious about is why we want to add this as a first-class feature in the Thanos/Prometheus API vs making the prom-label proxy part of the Observatorium infrastructure/deployment and add the enforcement there?

saswatamcode:
I think adding this as a first-class feature to Thanos/Prometheus API would be better as it solves a couple of issues, 🙂
- There are already issues asking for label-based matcher support in rules API upstream
- By adding it here, we avoid having another moving part in our deployment but don't compromise on tenancy

saswatamcode:
This can be merged once [thanos#5111](https://github.com/thanos-io/thanos/pull/5111) makes its way to main! 🙂

matej-g (approved):
Looks good, thank you @saswatamcode :+1:

PR #210, opened 2022-01-19
Use `go-kit/log` package instead of the deprecated `go-kit/kit/log`

matej-g:

This PR replaces usage of the deprecated log package(s) by it's current replacement. See notes in here: https://github.com/go-kit/kit/tree/master/log#package-log


squat (approved):
:+1:

PR #218, opened 2022-01-25
Propagate rules client status code

saswatamcode:
Currently, the API returns 200 status codes for all responses of the rules client. So even if the rules file in a PUT request is invalid, API returns 200 with the body `request body failed rule group validation` as observed in #198.

This PR ensures that the correct HTTP status code received from the rules client is propagated by the API, for the `rules/raw` endpoint.

Fixes [MON-2176](https://issues.redhat.com/browse/MON-2176).

### DIFF
--- a/api/metrics/v1/rules.go
+++ b/api/metrics/v1/rules.go
@@ -122,6 +122,8 @@ func (rh *rulesHandler) put(w http.ResponseWriter, r *http.Request) {
 
 	defer resp.Body.Close()
 
+	w.WriteHeader(resp.StatusCode)
+
 	if _, err := io.Copy(w, resp.Body); err != nil {
 		http.Error(w, "error writing rules response", http.StatusInternalServerError)
 		return


matej-g (approved):
Thanks for swiftly jumping in @saswatamcode! I'm wondering whether we should add a test case for this in here or still in https://github.com/observatorium/api/pull/198 :thinking:. Maybe we could finally merge the tests PR, rebase this PR and add the test / remove `TODO` as part of this?

saswatamcode:
Yes, the test can be added here once the rule tests PR is merged! 🙂

api/metrics/v1/rules.go
```diff
@@ -122,6 +122,8 @@ func (rh *rulesHandler) put(w http.ResponseWriter, r *http.Request) {
 
 	defer resp.Body.Close()
 
+	w.WriteHeader(resp.StatusCode)
+
```
matej-g, line 126: One more nit: Actually, on second look, I'm thinking if we should not be doing `WriteHeader` as the last thing, after `io.Copy`? If `io.Copy` fails, the `http.Error` method will also do `WriteHeader` behind the scenes. Not sure that will lead to [multiple `WriteHeader` calls ](https://stackoverflow.com/questions/42906490/whats-the-bad-effect-of-http-multiple-response-writeheader-calls)issue in this case, but perhaps also from readability standpoint, setting this before the end of the method might make it more clear? WDYT?

saswatamcode: Yes, I thought so too! But seems like `io.Copy` calls the `Write` method of the `io.Writer` passed, which in this case is an `http.ResponseWriter` and as per https://pkg.go.dev/net/http#ResponseWriter seems like the `Write` method will implicitly call `WriteHeader(http.StatusOK)` if no `WriteHeader` has been called before. So we need to call it before the `io.Copy` I think so that the original bug doesn't persist. 

This is what I found out from [here](https://stackoverflow.com/questions/26096944/when-responding-using-io-copy-who-should-be-responsible-for-the-error). So tested this locally with API, rules-objstore and minio,

Before,
```bash
curl -v -X PUT http://127.0.0.1:8080/api/metrics/v1/oidc-one/api/v1/rules/raw \
--header "Content-Type: application/yaml" \
-d "invalid:
- name: testing
 invalid_rules:
 - rule1: job:up:avg
   expr: avg without(instance)(up{job="node"})
 - rule2: ManyInstancesDown
   expr: job:up:avg{job="node"} < 0.5"

*   Trying 127.0.0.1:8080...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 8080 (#0)
> PUT /api/metrics/v1/oidc-one/api/v1/rules/raw HTTP/1.1
> Host: 127.0.0.1:8080
> User-Agent: curl/7.68.0
> Accept: */*
> Content-Type: application/yaml
> Content-Length: 170
> 
* upload completely sent off: 170 out of 170 bytes
* Mark bundle as not supporting multiuse
< HTTP/1.1 400 Bad Request
< Content-Length: 42
< Content-Type: text/plain; charset=utf-8
< Date: Tue, 25 Jan 2022 10:21:39 GMT
< 
request body failed rule group validation
* Connection #0 to host 127.0.0.1 left intact
```

With `w.WriteHeader` at end of method,
```bash
curl -v -X PUT http://127.0.0.1:8080/api/metrics/v1/oidc-one/api/v1/rules/raw \
--header "Content-Type: application/yaml" \
-d "invalid:
- name: testing
 invalid_rules:
 - rule1: job:up:avg
   expr: avg without(instance)(up{job="node"})
 - rule2: ManyInstancesDown
   expr: job:up:avg{job="node"} < 0.5"

*   Trying 127.0.0.1:8080...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 8080 (#0)
> PUT /api/metrics/v1/oidc-one/api/v1/rules/raw HTTP/1.1
> Host: 127.0.0.1:8080
> User-Agent: curl/7.68.0
> Accept: */*
> Content-Type: application/yaml
> Content-Length: 170
> 
* upload completely sent off: 170 out of 170 bytes
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Content-Length: 42
< Content-Type: text/plain; charset=utf-8
< Date: Tue, 25 Jan 2022 10:22:43 GMT
< 
request body failed rule group validation
* Connection #0 to host 127.0.0.1 left intact
```

matej-g: You're right, thanks for the full explanation! In that case this is resolved.

squat (approved):

jessicalins:
@saswatamcode awesome! I'll re-add the test in https://github.com/observatorium/api/pull/198 :)

PR #219, opened 2022-01-25
Enforce tenant labels on rule storage PUT request

saswatamcode:
Currently, we enforce tenant labels on rules by adding the labels to them when Observatorium API gets it from the rules storage client i.e, the rules will only have the tenant labels when we get it from the API `rules/raw` endpoint.

With the recent changes in thanos-rule-syncer and rules-objstore, this means,
- Observatorium API doesn’t add a tenant label while writing rules to rules-objstore
- With the `listallrules` endpoint, thanos-rule-syncer will try to get all rules by hitting rules-objstore directly (so enforcing labels won’t work as we’re no longer relying on the API read path, but read from rules-objstore directly)
- So the rule file generated by thanos-rule-syncer, which is to be used by Thanos Ruler won’t have any tenant labels, so any resultant series will also not have labels

This PR fixes the above issue, by ensuring that the API adds tenant labels while making the PUT request to the rules storage client and validating the labels when a GET request is made. This way tenant labels are propagated to the rules storage client and then eventually to thanos-rule-syncer and Thanos Ruler.

For more diagrammatic context see [MON-2169](https://issues.redhat.com/browse/MON-2169).

### DIFF
--- a/api/metrics/v1/rules.go
+++ b/api/metrics/v1/rules.go
@@ -13,6 +13,29 @@ import (
 	"github.com/observatorium/api/rules"
 )
 
+func marshalRules(rawRules rules.Rules) ([]byte, error) {
+	body, err := yaml.Marshal(rawRules)
+	if err != nil {
+		return nil, err
+	}
+
+	return body, nil
+}
+
+func unmarshalRules(r io.Reader) (rules.Rules, error) {
+	body, err := ioutil.ReadAll(r)
+	if err != nil {
+		return rules.Rules{}, err
+	}
+
+	var rawRules rules.Rules
+	if err := yaml.Unmarshal(body, &rawRules); err != nil {
+		return rules.Rules{}, err
+	}
+
+	return rawRules, nil
+}
+
 type rulesHandler struct {
 	client      rules.ClientInterface
 	logger      log.Logger
@@ -53,16 +76,10 @@ func (rh *rulesHandler) get(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	body, err := ioutil.ReadAll(resp.Body)
+	rawRules, err := unmarshalRules(resp.Body)
 	if err != nil {
-		http.Error(w, "error listing rules", http.StatusInternalServerError)
-		return
-	}
-
-	var rawRules rules.Rules
-	if err := yaml.Unmarshal(body, &rawRules); err != nil {
 		level.Error(rh.logger).Log("msg", "could not unmarshal rules", "err", err.Error())
-		http.Error(w, "error unmarshaling rules", http.StatusInternalServerError)
+		http.Error(w, "error unmarshalling rules", http.StatusInternalServerError)
 
 		return
 	}
@@ -94,10 +111,10 @@ func (rh *rulesHandler) get(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	body, err = yaml.Marshal(rawRules)
+	body, err := marshalRules(rawRules)
 	if err != nil {
-		level.Error(rh.logger).Log("msg", "could not marshal YAML", "err", err.Error())
-		http.Error(w, "error marshaling YAML", http.StatusInternalServerError)
+		level.Error(rh.logger).Log("msg", "could not marshal rules YAML", "err", err.Error())
+		http.Error(w, "error marshaling rules YAML", http.StatusInternalServerError)
 
 		return
 	}
@@ -120,16 +137,10 @@ func (rh *rulesHandler) put(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	body, err := ioutil.ReadAll(r.Body)
+	rawRules, err := unmarshalRules(r.Body)
 	if err != nil {
-		http.Error(w, "error reading request body", http.StatusInternalServerError)
-		return
-	}
-
-	var rawRules rules.Rules
-	if err := yaml.Unmarshal(body, &rawRules); err != nil {
 		level.Error(rh.logger).Log("msg", "could not unmarshal rules", "err", err.Error())
-		http.Error(w, "error unmarshaling rules", http.StatusInternalServerError)
+		http.Error(w, "error unmarshalling rules", http.StatusInternalServerError)
 
 		return
 	}
@@ -155,10 +166,10 @@ func (rh *rulesHandler) put(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	body, err = yaml.Marshal(rawRules)
+	body, err := marshalRules(rawRules)
 	if err != nil {
-		level.Error(rh.logger).Log("msg", "could not marshal YAML", "err", err.Error())
-		http.Error(w, "error marshaling YAML", http.StatusInternalServerError)
+		level.Error(rh.logger).Log("msg", "could not marshal rules YAML", "err", err.Error())
+		http.Error(w, "error marshaling rules YAML", http.StatusInternalServerError)
 
 		return
 	}
--- a/api/metrics/v1/rules.go
+++ b/api/metrics/v1/rules.go
@@ -13,6 +13,29 @@ import (
 	"github.com/observatorium/api/rules"
 )
 
+func enforceLabelsInRules(rawRules rules.Rules, tenantLabel string, tenantID string) {
+	for i := range rawRules.Groups {
+		for j := range rawRules.Groups[i].Rules {
+			switch r := rawRules.Groups[i].Rules[j].(type) {
+			case rules.RecordingRule:
+				if r.Labels.AdditionalProperties == nil {
+					r.Labels.AdditionalProperties = make(map[string]string)
+				}
+
+				r.Labels.AdditionalProperties[tenantLabel] = tenantID
+				rawRules.Groups[i].Rules[j] = r
+			case rules.AlertingRule:
+				if r.Labels.AdditionalProperties == nil {
+					r.Labels.AdditionalProperties = make(map[string]string)
+				}
+
+				r.Labels.AdditionalProperties[tenantLabel] = tenantID
+				rawRules.Groups[i].Rules[j] = r
+			}
+		}
+	}
+}
+
 func unmarshalRules(r io.Reader) (rules.Rules, error) {
 	body, err := ioutil.ReadAll(r)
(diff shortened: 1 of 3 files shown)


api/metrics/v1/rules.go
```diff
@@ -107,7 +114,56 @@ func (rh *rulesHandler) put(w http.ResponseWriter, r *http.Request) {
...
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		http.Error(w, "error reading request body", http.StatusInternalServerError)
+		return
+	}
+
+	var rawRules rules.Rules
+	if err := yaml.Unmarshal(body, &rawRules); err != nil {
+		level.Error(rh.logger).Log("msg", "could not unmarshal rules", "err", err.Error())
+		http.Error(w, "error unmarshaling rules", http.StatusInternalServerError)
```
jessicalins, line 132:
```suggestion
		http.Error(w, "error unmarshalling rules", http.StatusInternalServerError)
```

api/metrics/v1/rules.go
```diff
@@ -107,7 +114,56 @@ func (rh *rulesHandler) put(w http.ResponseWriter, r *http.Request) {
...
+
+				r.Labels.AdditionalProperties[rh.tenantLabel] = id
+				rawRules.Groups[i].Rules[j] = r
+			}
+		}
+	}
+
+	body, err = yaml.Marshal(rawRules)
+	if err != nil {
+		level.Error(rh.logger).Log("msg", "could not marshal YAML", "err", err.Error())
```
jessicalins, line 160:
```suggestion
		level.Error(rh.logger).Log("msg", "could not marshal rules YAML", "err", err.Error())
```

api/metrics/v1/rules.go
```diff
@@ -107,7 +114,56 @@ func (rh *rulesHandler) put(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, "error marshaling YAML", http.StatusInternalServerError)
```
jessicalins, line 161:
```suggestion
		http.Error(w, "error marshaling rules YAML", http.StatusInternalServerError)
```

jessicalins: I think since we have the same marshaling logic also in the `get` [method](https://github.com/observatorium/api/blob/80ff927bebedc78bf7092d494526bd88a5428b3b/api/metrics/v1/rules.go#L97-L103) maybe we could extract this to a separate function and reuse it in both get/put endpoints? same for the [auth logic](https://github.com/observatorium/api/blob/80ff927bebedc78bf7092d494526bd88a5428b3b/api/metrics/v1/rules.go#L23-L33)

jessicalins (commented):
cool! :triangular_ruler: I think after we merge https://github.com/observatorium/api/pull/198 we'd need to update the tests for the write cases too

api/metrics/v1/rules.go
saswatamcode: Yes, I think we can do it for both marshaling and unmarshalling rules!

matej-g (commented):
Looking good and great PR description @saswatamcode! Few questions, mostly just curious :grin:

api/metrics/v1/rules.go
```diff
@@ -71,26 +89,32 @@ func (rh *rulesHandler) get(w http.ResponseWriter, r *http.Request) {
 			switch r := rawRules.Groups[i].Rules[j].(type) {
 			case rules.RecordingRule:
 				if r.Labels.AdditionalProperties == nil {
-					r.Labels.AdditionalProperties = make(map[string]string)
+					http.Error(w, "rule has no label", http.StatusInternalServerError)
```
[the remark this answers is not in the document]
saswatamcode: No, you're not missing anything the cases are exactly the same. But we switch by type so that we can get the `Labels`. If we use no switch or use `cases rules.RecordingRule, rules.AlertingRule:` then it's just an interface.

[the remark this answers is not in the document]
saswatamcode: I'm on the fence too about this. 
But it is kind of a fault on our end, as this switch case is only entered when there is a rule in a rule group. 
So we have a rule, but with no label which is saved in rules storage, which means something is very wrong, as we do not allow rules without tenant labels to be saved in the first place. So a `5xx` status code would seem apt, right?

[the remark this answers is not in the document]
matej-g: Ah, that's an interesting gotcha, now I see, thanks :+1:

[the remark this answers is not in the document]
squat: i kind of think that we should continue to enforce the labels on the read path rather than return a non-200 error code. If the rules were loaded into the rules backend some other way than through the observatorium API, then this current changeset would make the rules un-fetchable. What's the downside to enforcing again and returning 200s?

matej-g (approved):
Nothing more to add :partying_face:

squat (changes requested):
just one comment on continuing to enforce on read path

api/metrics/v1/rules.go
```diff
@@ -12,6 +13,29 @@ import (
 	"github.com/observatorium/api/rules"
 )
 
+func marshalRules(rawRules rules.Rules) ([]byte, error) {
+	body, err := yaml.Marshal(rawRules)
```
squat, line 17: as a rule, we try to avoid defining functions that only call one function on the inside. As it stands there is no benefit for the caller to call `marshalRules` over directly using `yaml.Marshal`

api/metrics/v1/rules.go
```diff
@@ -52,16 +76,10 @@ func (rh *rulesHandler) get(w http.ResponseWriter, r *http.Request) {
...
 	if err != nil {
-		http.Error(w, "error listing rules", http.StatusInternalServerError)
-		return
-	}
-
-	var rawRules rules.Rules
-	if err := yaml.Unmarshal(body, &rawRules); err != nil {
 		level.Error(rh.logger).Log("msg", "could not unmarshal rules", "err", err.Error())
-		http.Error(w, "error unmarshaling rules", http.StatusInternalServerError)
+		http.Error(w, "error unmarshalling rules", http.StatusInternalServerError)
```
squat, line 82:
```suggestion
		http.Error(w, "error unmarshaling rules", http.StatusInternalServerError)
```

api/metrics/v1/rules.go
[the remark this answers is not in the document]
saswatamcode: The downside is that the original problem mentioned in the ticket would persist.

If we add rules to the rules backend some other way and don't add in the tenant labels, the rules would again be propagated and configured on the Ruler and the series generated won't have tenant labels on them as well. So, the tenant the rules belong to can't query them.
But if we enforce it on the read path, then in the above condition the observatorium API will return a 200 status code and the rules with added tenant labels, even though the series generated from rules aren't actually being queryable (due to missing labels).

This changeset ensures that all rules in the rules backend have a tenant label. So even if we load the rules some other way in the rules backend, we must provide the tenant label. 🙂

api/metrics/v1/rules.go
saswatamcode: Sure, let me change!

saswatamcode:
Let me also update this with the test cases from #198.

test/e2e/configs.go
```diff
@@ -205,6 +205,7 @@ func createRulesYAML(
 const recordingRuleYamlTpl = `
 groups:
   - name: example
+    interval: 30s
```
saswatamcode, line 208: Had to update the tests a bit with rule group `interval` and alert `for` and `annotations` fields. Currently, as it's defined in [spec.yaml](https://github.com/observatorium/api/blob/main/rules/spec.yaml#L83), all these fields are `required` which means that (un)marshaling them will lead to `interval: ""` or `annotations: {}` if they aren't provided by the tenant.

Is there a way to handle this better or allow tenant to not have to provide these fields?

test/e2e/configs.go
```diff
@@ -241,3 +247,16 @@ invalid:
  - rule2: ManyInstancesDown
    expr: job:up:avg{job="node"} < 0.5
 `
+
+const validYamlWithInvalidRulesYamlTpl = `
```
saswatamcode, line 251: Also, had to add a new invalid case. 
So if YAML is completely invalid, then Observatorium API won't be able to parse it and respond with a `500` status code, as is covered by the `write-invalid-rules` test case.

In the new `write-valid-yaml-with-invalid-rules` case, the YAML is valid and API is able to parse it and sends it to rules backend, after which validation fails and it responds with `400` status code.

Open to suggestions on how this can be handled better. 🙂

api/metrics/v1/rules.go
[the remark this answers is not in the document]
squat: @saswatamcode i think you perhaps misunderstood my comment. I'm proposing we enforce labels on both the PUT and the GET (for a single tenant). As far as I can tell, there is no downside at all to this and it protects us in the event that rules are sideloaded into the ruler.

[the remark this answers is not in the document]
saswatamcode: Ack! Yes, there's no downside then. Let me push a commit!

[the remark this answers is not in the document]
squat: Ok nice :) one last thing, it looks like both the PUT and the GET handlers use the exact same logic to enforce labels. Maybe this whole code block could be abstracted to a func that takes the tenant name and rules pre-typeswitch and applies the labels. WDYT?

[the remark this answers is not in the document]
saswatamcode: Yes, that would be better!

squat (approved):
Thanks @saswatamcode

test/e2e/configs.go
matej-g: I think this is correct, in the end user will end up with `4xx`, understanding that an invalid rule has been provided.

test/e2e/configs.go
matej-g: Wait, I'm confused here, so the API is accepting a rule file even if those empty parameters are required? That does not sound right.

matej-g: We clarified this F2F with @saswatamcode as of now this is indeed how it works in API, but it is outside the scope of this PR, resolving.

matej-g (approved):

PR #228, opened 2022-02-03
Interactive test: Run rules services as part of the interactive test

matej-g:

Adds rules service to run when running the interactive setup.


jessicalins (approved):
:straight_ruler: cool! would be nice to have it also in the `newUpRun` function but I think is a bit more complicated to change there since we'd need to change the lib too I think (to support an [extra](https://github.com/observatorium/api/blob/e486b8926a8b00cfb108f6e36e224729936714b3/test/e2e/services.go#L365) arg)

Issue #194, opened 2021-11-18
Rules API: Validate tenant rules


jessicalins:
As mentioned by @squat  [here](https://github.com/observatorium/api/pull/191):

"Note: this is a distinct problem from ensuring that the rules written by
a tenant are guaranteed to only include data from that tenant.
Currently, the rules from a tenant can select data from any tenant. This
is a distinct problem and should absolutely be tackled in a follow up."

jessicalins:
solved by https://github.com/observatorium/api/pull/230 - closing

PR #230, opened 2022-02-04
Rules API: Enforce tenant label and id in rule expression

jessicalins:
To help with tenant rules validation (as mentioned in https://github.com/observatorium/api/issues/194), this PR adds a `tenant_id` label to metrics present in a `expr` field in a recording/alerting rule that is written via the Rules API (via `/api/v1/rules/raw` endpoint)

So for the following rule that is written via Rules API:
```yaml
groups:
- name: node_rules
  interval: 30s
  rules:
  - record: job:up:avg
    expr: avg without(instance)(up{job="node"})
  - alert: ManyInstancesDown
    expr: job:up:avg{job="node"} < 0.5
    for: 10m
    annotations:
      Summary: Many instances down
```
We expect that when the rule is read via `GET` request, the rule is changed and a `tenant_id` label is appended to `up` and `avg`:

`expr: avg without(instance)(up{job="node",tenant_id="tenantIDHere"})`
`expr: job:up:avg{job="node",tenant_id="tenantIDHere"} < 0.5`

I've also modified the current e2e tests to support this change.

Signed-off-by: Jéssica Lins <jessicaalins@gmail.com>

### DIFF
--- a/api/metrics/v1/rules.go
+++ b/api/metrics/v1/rules.go
@@ -3,7 +3,6 @@ package v1
 import (
 	"bytes"
 	"fmt"
-	"github.com/prometheus-community/prom-label-proxy/injectproxy"
 	"io"
 	"io/ioutil"
 	"net/http"
@@ -13,8 +12,9 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/observatorium/api/authentication"
 	"github.com/observatorium/api/rules"
-	"github.com/prometheus/prometheus/promql/parser"
 	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/promql/parser"
+	"github.com/prometheus-community/prom-label-proxy/injectproxy"
 )
 
 func enforceLabelsInRules(rawRules rules.Rules, tenantLabel string, tenantID string) error {
@@ -59,17 +59,20 @@ func enforceLabelsInRules(rawRules rules.Rules, tenantLabel string, tenantID str
 			}
 		}
 	}
+
 	return nil
 }
 
-func enforceLabelsInExpr(e *injectproxy.Enforcer, expr string) (string, error){
+func enforceLabelsInExpr(e *injectproxy.Enforcer, expr string) (string, error) {
 	parsedExpr, err := parser.ParseExpr(expr)
 	if err != nil {
 		return "", fmt.Errorf("parse expr error: %w", err)
 	}
+
 	if err := e.EnforceNode(parsedExpr); err != nil {
 		return "", fmt.Errorf("enforce node error: %w", err)
 	}
+
 	return parsedExpr.String(), nil
 }
 
@@ -142,7 +145,6 @@ func (rh *rulesHandler) get(w http.ResponseWriter, r *http.Request) {
 
 		return
 	}
-
 
 	body, err := yaml.Marshal(rawRules)
 	if err != nil {
--- a/api/metrics/v1/rules.go
+++ b/api/metrics/v1/rules.go
@@ -12,9 +12,9 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/observatorium/api/authentication"
 	"github.com/observatorium/api/rules"
+	"github.com/prometheus-community/prom-label-proxy/injectproxy"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/promql/parser"
-	"github.com/prometheus-community/prom-label-proxy/injectproxy"
 )
 
 func enforceLabelsInRules(rawRules rules.Rules, tenantLabel string, tenantID string) error {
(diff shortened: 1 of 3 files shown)


ianbillett (changes requested):
Nice work @jessicalins :+1: 

One big question though - have you considered and discounted adding the tenant label into the rule spec itself rather than parsing and adding it into the query itself?

From the Prometheus rule [spec](https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/):
```
# Labels to add or overwrite before storing the result.
labels:
  [ <labelname>: <labelvalue> ]
``` 

This would manifest here as an additional entry to the rule object, rather than parsing into the query itself.

jessicalins:
@ianbillett we actually already enforce the `tenant_id` label on returned rules (Lucas added [here](https://github.com/observatorium/api/pull/191/files)) and in the [spec](https://github.com/observatorium/api/blob/main/rules/spec.yaml#L112)

So a returned rule (before this PR) was already like: 
```yaml
groups:
- interval: 15s
  name: ruletwo
  rules:
  - expr: demo_test_metric
    labels:
      tenant_id: sdffsd34-c509-4592-a256-a18713657bfa
    record: job:demo_test_metric:ruletwometric
```

The problem is that when we query for a rule itself (e.g in this case, for `job:demo_test_metric:ruletwometric` - e.g. `http://127.0.0.1:8082/api/metrics/v1/oidc-two/api/v1/query?query=job:demo_test_metric:ruletwometric`) it would return metrics that match that rule, including rules from other tenants, instead of only the tenant rules.
So adding the tenant label to the `expr` helps with that. Discussed this with @onprem @saswatamcode - let me know if I'm missing something:)

test/e2e/rules_test.go
```diff
@@ -117,7 +117,9 @@ func TestRulesAPI(t *testing.T) {
 
 		body, err := ioutil.ReadAll(res.Body)
 		bodyStr := string(body)
+
 		assertResponse(t, bodyStr, "alert: HighRequestLatency")
+		assertResponse(t, bodyStr, "job:request_latency_seconds:mean5m{job=\"myjob\",tenant_id=\""+defaultTenantID+"\"}")
```
saswatamcode, line 122: Maybe adding a more complex query like the ones we already use in rhobs/configuration would be a good idea here? 🙂

test/e2e/configs.go
```diff
@@ -207,35 +207,35 @@ groups:
   - name: example
     interval: 30s
     rules:
-      - record: job:http_inprogress_requests:sum
-        expr: sum by (job) (http_inprogress_requests)
+      - record: id_network_type
+        expr: 0 * topk by (ebs_account) (1, max by (ebs_account,account_type,internal,email_domain) (label_replace(label_replace(label_replace(subscription_labels{email_domain="domain1.com"}*0+5, "class", "Internal", "class", ".*") or label_replace(subscription_labels{class!="Customer",email_domain=~"(.*\\.|^)domain2.com"}*0+4, "class", "Internal", "class", ".*") or (subscription_labels{class="Customer"}*0+3) or (subscription_labels{class="Partner"}*0+2) or (subscription_labels{class="Evaluation"}*0+1) or label_replace(subscription_labels{class!~"Evaluation|Customer|Partner"}*0+0, "class", "", "class", ".*"), "account_type", "$1", "class", "(.+)"), "internal", "true", "email_domain", "domain1.com|(.*\\.|^)domain2.com") ))
```
jessicalins, line 211: got it from [observatorium-metrics-template.yaml ](https://github.com/rhobs/configuration/blob/3978c8609271eb30fe16492af5f6b507ea86e139/resources/services/observatorium-metrics-template.yaml#L1340) in `rhobs/configuration`
I've just modified to be a more general expr, replacing `redhat` and `ibm` by `domain1` and `domain2`

ianbillett (dismissed):
Righto! I see - thanks for explaining. If you've discussed and agreed on this path forward with @onprem  & @saswatamcode then this solution works for me.

api/metrics/v1/rules.go
```diff
@@ -98,7 +138,13 @@ func (rh *rulesHandler) get(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	enforceLabelsInRules(rawRules, rh.tenantLabel, id)
+	err = enforceLabelsInRules(rawRules, rh.tenantLabel, id)
+	if err != nil {
+		level.Error(rh.logger).Log("msg", "could not enforce labels in rules", "err", err.Error())
+		http.Error(w, "error enforcing labels in rules", http.StatusInternalServerError)
```
matej-g, line 144: Nit: I feel like this might be a bit cryptic in the response for the user, maybe something along the lines 'failed to process rules' would suffice.

matej-g (approved):
Nice! 🍪

matej-g (approved):

PR #235, opened 2022-02-08
Rules API: Make response more verbose

matej-g:
When working with rules API, the response message is not always very useful for the user. I suggest to:
- If we don't have any rules stored for the tenant, return message saying no rules found instead of 'error loading rule'
- If (un)marshalling fails, provide the error in the response. This is useful in particular if the error message including line number, which can be helpful for the user during troubleshooting.


saswatamcode (approved):
Thanks! This makes sense! 🙂

Just need to fix lint! But we ran across [this](https://app.circleci.com/pipelines/github/observatorium/api/436/workflows/e7b67c80-e63e-4a88-a9c1-78192c1206da/jobs/2111?invite=true#step-102-349) again. Worth making an issue?

api/metrics/v1/rules.go
```diff
@@ -86,14 +87,20 @@ func (rh *rulesHandler) get(w http.ResponseWriter, r *http.Request) {
...
+		}
+
 		return
 	}
 
 	rawRules, err := unmarshalRules(resp.Body)
 	if err != nil {
 		level.Error(rh.logger).Log("msg", "could not unmarshal rules", "err", err.Error())
-		http.Error(w, "error unmarshaling rules", http.StatusInternalServerError)
+		http.Error(w, fmt.Sprintf("error unmarshaling rules: %v", err), http.StatusInternalServerError)
```
squat, line 103: We normally don't want to return arbitrary errors to the user from API. We don't do this (AFAIK) anywhere in Observatorium or other projects because it raises the possibility of injection attacks where users trigger a known error to get the API to construct a reply with an injected value, which can later be exploited.

squat: In other words, it's our convention to _not_ echo untrusted values back to users

Makefile
```diff
@@ -90,10 +90,12 @@ test-unit:
 
 .PHONY: test-e2e
 test-e2e: container-test
-	CGO_ENABLED=1 GO111MODULE=on go test -v -race -short --tags integration ./test/e2e
+	@rm -rf test/e2e/e2e_*
```
jessicalins, line 93: nice! :space_invader:

test/e2e/rules_test.go
```diff
@@ -12,6 +12,65 @@ import (
 	"github.com/efficientgo/tools/core/pkg/testutil"
 )
 
+const recordingRuleYamlTpl = `
```
jessicalins, line 15: why not keep them in `test/e2e/configs.go`?

api/metrics/v1/rules.go
matej-g: Done :+1: I removed this change.

matej-g: Done :+1: I removed this change.

test/e2e/rules_test.go
matej-g: I felt like it is easier to find these in the same file as rule tests, since it's strictly not a test service configuration, but rather test data.

jessicalins, line 15: I think to make the e2e tests pass we need now to update this const as in [here](https://github.com/observatorium/api/blob/main/test/e2e/configs.go#L205-L212) 
after that LGTM :)

squat (approved):

test/e2e/rules_test.go
matej-g: Whoops, missed this during merge conflict, thank you :bow:

jessicalins (approved):
:watermelon:

PR #237, opened 2022-02-10
Enable Rate Limiting for Logs endpoints

tareqmamari:
Currently, Rate Limiting is only enabled for metrics, this commit
enables it also for logs endpoints.

### DIFF
--- a/main.go
+++ b/main.go
@@ -568,6 +568,12 @@ func main() {
 					r.Use(authentication.WithTenantMiddlewares(pm.Middlewares))
 					r.Use(authentication.WithTenantHeader(cfg.logs.tenantHeader, tenantIDs))
 
+					if rateLimitClient != nil {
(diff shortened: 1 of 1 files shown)


matej-g:
@tareqmamari Could you please take a look at the failing tests? It seems related.

periklis:
The reason we don't have this supported yet is that Loki supports a more special-purpose rate limiting on the stream level. It is not per se wrong to rate limit on the API, but till now the usage scenario is to use Loki's rate limiting capabilities on the distributor (See `limits_config`)

tareqmamari:
@periklis Thanks for the clarification, indeed that is interesting. Our goal to have a dynamic tenants management, putting all tenants configuration in one place alongside its rate limiting configuration would enable us to achieve that better without the need to reconfigure/update loki configuration.

tareqmamari:
@matej-g we will open a new PR, a bigger one, and that would touch the part updated here as well. I am closing this one, and we can discuss it later in the new PR.

PR #246, opened 2022-03-07
gRPC tests for trace handling

esnible:


test/e2e/traces_test.go
```diff
@@ -0,0 +1,244 @@
...
+			}
+			]
+		}
+		]
+	}
+	]
+}`
+)
+
+func TestTracesExport(t *testing.T) {
```
saswatamcode, line 102: This method doesn't run a test yet [as per CI](https://app.circleci.com/pipelines/github/observatorium/api/471/workflows/3f2c2901-2ee6-49bc-b80b-6a7d1c50a918/jobs/2289), the reason being that it doesn't have a `t.Run` call yet. 

I think that [this block](https://github.com/observatorium/api/pull/246/files#diff-55863b90b036d7b2b31833355b1bfaef5ad7140e9f67e7364ac710f3f2e7af81R192-R224) can be added within a `t.Run` as it is sending the trace and then querying for it, so `t.Run("write-then-query-single-trace", func..` would be a good idea! 🙂

test/e2e/traces_test.go
```diff
@@ -0,0 +1,244 @@
...
+
+	e, err := e2e.NewDockerEnvironment(envTracesName)
+	testutil.Ok(t, err)
+	t.Cleanup(e.Close)
+
+	prepareConfigsAndCerts(t, traces, e)
+	dex, token, _ := startBaseServices(t, e, traces)
+	internalOtlpEndpoint, httpQueryEndpoint := startServicesForTraces(t, e)
+
+	// oidcEndpoint := dex.Endpoint("https")
```
saswatamcode, line 113: Is this comment necessary?

test/e2e/traces_test.go
```diff
@@ -0,0 +1,244 @@
...
+	testutil.Ok(t, err)
+
+	// Note that we don't wait for the trace to be committed to storage.
+	// (If we were using a buffered Jaeger storage backend we would
+	// not give up on the first fetch attempt.)
+	response, err = client.Do(request)
+	testutil.Ok(t, err)
+	defer response.Body.Close()
+
+	if response.StatusCode != http.StatusOK {
```
saswatamcode, line 221: Same as above.

test/e2e/traces_test.go
```diff
@@ -0,0 +1,244 @@
...
+		responseBytes, _ := io.ReadAll(response.Body)
+		t.Fatalf("reading trace from Jaeger not OK: %d\n%s", response.StatusCode, responseBytes)
+	}
+	// Parse and verify the trace is the same?
```
saswatamcode, line 225: I think we should verify the response in this case, we already have an `assertResponse` helper method [here](https://github.com/observatorium/api/blob/main/test/e2e/helpers.go#L114). 🙂

test/e2e/traces_test.go
```diff
@@ -0,0 +1,244 @@
...
+	// responseBytes, _ := io.ReadAll(response.Body)
+	// ...
+
+	// Uncomment for interactive test
```
saswatamcode, line 229: Might be a good idea to add tracing setup to `TestInteractiveSetup`, as was done for Rules API in https://github.com/observatorium/api/pull/228. This can be a separate PR too I think. 🙂

test/e2e/traces_test.go
```diff
@@ -0,0 +1,244 @@
...
+		"POST",
+		fmt.Sprintf("http://%s/v1/traces", otel.Endpoint("http")),
+		bytes.NewBuffer([]byte(traceJSON)))
+	testutil.Ok(t, err)
+	request.Header.Set("Content-Type", "application/json")
+	response, err := client.Do(request)
+	testutil.Ok(t, err)
+	defer response.Body.Close()
+
+	if response.StatusCode != http.StatusOK {
```
saswatamcode, line 203: Might be a good idea to replace this by `testutil.Equals(t, http.StatusOK, res.StatusCode)`.

test/e2e/traces_test.go
```diff
@@ -0,0 +1,244 @@
...
+		tokenUrl, -1)
+	forwardingConfig = strings.Replace(forwardingConfig,
+		"{{OBS_GRPC_ENDPOINT}}",
+		api.InternalEndpoint("grpc"), -1)
+	forwardingConfig = strings.Replace(forwardingConfig,
+		"{{DEX_TOKEN}}",
+		token, -1)
+
+	otelFile, err := ioutil.TempFile(e.SharedDir(), "fwd-coll.yaml")
+	if err != nil {
```
saswatamcode, line 153: These can be replaced with `testutil.Ok(t, err)`.

saswatamcode (commented):
Thanks for the awesome work! 💫
Just some suggestions!

test/e2e/traces_test.go
moadz: Yup, response could be `200` but empty body.

moadz (commented):
LGTM % Response verification and Saswata's comments.

test/e2e/services.go
```diff
@@ -1,11 +1,15 @@
-// +build integration interactive
+//go:build integration || interactive || experimentalintegration
```
jessicalins, line 1: I think here we need to keep the `// +build` syntax, since we are using go 1.16 (as mentioned in the docs [here](https://go.dev/doc/go1.16#:~:text=Note%20that%20%2F%2Fgo%3Abuild,introduced%20into%20Go%20programs%20yet.))

test/e2e/services.go
```diff
@@ -73,6 +105,64 @@ func startServicesForLogs(t *testing.T, e e2e.Environment) (
...
+				"http.query":  16686, // Query
+			}).
+		Init(e2e.StartOptions{Image: "jaegertracing/all-in-one:1.31"})
+
+	// Warn if a YAML change introduced a tab character
+	if strings.ContainsRune(otelConfig, '\t') {
+		t.Errorf("Tab in the YAML")
+	}
+
+	config := strings.Replace(otelConfig,
```
jessicalins, line 123: I think we could move the creation of the yaml config file to https://github.com/observatorium/api/blob/main/test/e2e/configs.go and add a function like `createTracesYAML` or something similar - what do you think?
So [this](https://github.com/observatorium/api/blob/ca9ffa0586a31db2ef842713ccfc1363fc783a96/test/e2e/services.go#L37-L63) could also be moved there

test/e2e/services.go
```diff
@@ -73,6 +105,64 @@ func startServicesForLogs(t *testing.T, e e2e.Environment) (
...
+	}
+
+	otel := e.Runnable("otel-collector").
+		WithPorts(
+			map[string]int{
+				"grpc": 4317,
+				"http": 4318,
+			}).
+		Init(e2e.StartOptions{
+			Image:   "otel/opentelemetry-collector:0.45.0",
```
jessicalins, line 153: I think we could move this to a const [here](https://github.com/observatorium/api/blob/ca9ffa0586a31db2ef842713ccfc1363fc783a96/test/e2e/services.go#L21) so is easier to replace in case the image is updated

test/e2e/services.go
```diff
@@ -308,6 +412,24 @@ func newObservatoriumAPIService(
 		args = append(args, "--middleware.rate-limiter.grpc-address="+opts.ratelimiterAddr)
 	}
 
+	if opts.tracesWriteEndpoint != "" {
+		args = append(args, "--traces.write.endpoint="+opts.tracesWriteEndpoint)
```
jessicalins, line 416: I think we could also update [usage](https://github.com/observatorium/api#usage) with those flags

test/e2e/traces_test.go
```diff
@@ -0,0 +1,245 @@
+//go:build experimentalintegration
```
jessicalins, line 1: same here

test/e2e/traces_test.go
```diff
@@ -0,0 +1,245 @@
...
+	otelForwardingConfig = `
+receivers:
+    otlp:
+      protocols:
+        http:
+            endpoint: 0.0.0.0:4318
+        grpc:
+            endpoint: 0.0.0.0:4317
+
+# extensions:
```
jessicalins, line 31: will the parts that are commented out be used at some point? if not, I think we can remove them?

test/e2e/traces_test.go
```diff
@@ -0,0 +1,245 @@
...
+		body, err = ioutil.ReadAll(response.Body)
+		testutil.Ok(t, err)
+
+		bodyStr = string(body)
+		//nolint:lll
+		assertResponse(t, bodyStr, `{"result":{"resourceSpans":[{"resource":{"attributes":[{"key":"host.name","value":{"stringValue":"testHost"}}]},"instrumentationLibrarySpans":[{"instrumentationLibrary":{},"spans":[{"traceId":"W47/95gDgQPSabYzgT/GDA==","spanId":"7uGbfsPBsXM=","parentSpanId":"AAAAAAAAAAA=","name":"testSpan","kind":"SPAN_KIND_INTERNAL","startTimeUnixNano":"1544712660000000000","endTimeUnixNano":"1544712661000000000","attributes":[{"key":"attr1","value":{"intValue":"55"}},{"key":"internal.span.format","value":{"stringValue":"proto"}}]}]}]}]}}`)
+
+		testutil.Equals(t, http.StatusOK, response.StatusCode)
+
+		// Uncomment for interactive test
```
jessicalins, line 229: would this happen in this same PR? I think if not, we could also remove the commented lines

jessicalins (commented):
looking good, thanks for adding this! left some suggestions

test/e2e/services.go
esnible: These were added in the previous PR, the one that added the functionality.

- https://github.com/observatorium/api/pull/199/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R78
- https://github.com/observatorium/api/pull/199/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R78

esnible:
@jessicalins Thank you for your comments.  I have addressed them.  The test is passing, but the build is not passing consistently, because of the flaky TestRulesAPI reported in https://github.com/observatorium/api/issues/249

test/e2e/services.go
jessicalins: from the [docs](https://go.dev/doc/go1.16#:~:text=Note%20that%20%2F%2Fgo%3Abuild,introduced%20into%20Go%20programs%20yet.): 
"Note that //go:build lines are not supported in Go 1.16 and should not be introduced into Go programs yet." 

As far as I understand looks like for 1.16 is recommended to use only `// + build` - I think we could for now follow this and when updating to `1.17` update the build tags to use the newest syntax
Could be a bit annoying with your setup - maybe you could install go 1.16 as well and configure to use it just with observatorium/api project?

jessicalins:
@esnible regarding the failing e2e tests, me and @matej-g are taking a look at it. We would need to do changes in https://github.com/efficientgo/e2e , we'll let you know once this is merged and fixed here too

jessicalins:
@esnible could you try rebasing with latest main? this [pr](https://github.com/observatorium/api/pull/251) should have fixed CI

esnible:
@jessicalins All tests now pass.

test/e2e/traces_test.go
```diff
@@ -0,0 +1,231 @@
...
+	testutil.Ok(t, e2e.StartAndWaitReady(api))
+
+	t.Run("write-then-query-single-trace", func(t *testing.T) {
+
+		forwardingConfig := strings.Replace(otelForwardingConfig,
+			"{{OBS_GRPC_ENDPOINT}}",
+			api.InternalEndpoint("grpc"), -1)
+		forwardingConfig = strings.Replace(forwardingConfig,
+			"{{DEX_TOKEN}}",
+			token, -1)
```
matej-g, line 136: Why not use the more traditional `fmt.Sprintf`? :slightly_smiling_face:

test/e2e/traces_test.go
```diff
@@ -0,0 +1,231 @@
...
+)
+
+const (
+	// Note that if the forwarding collector uses OIDC flow instead of hard-coding
+	// the bearer token we would need
+	// "otel/opentelemetry-collector-contrib:0.45.0" instead.
+	otelFwdCollectorImage = "otel/opentelemetry-collector:0.45.0"
+
+	// OTel trace collector that receives in HTTP w/o security, but exports in gRPC with security.
+	otelForwardingConfig = `
```
matej-g, line 30: What do you think about moving these to `config.go` and moving the forwarder initialization method to `services.go`? It's a bit more natural fit at least for me.

matej-g (commented):
This is looking great @esnible! :+1: I have some very minor comments, but otherwise I'm happy to get this PR in.

test/e2e/configs.go
```diff
@@ -1,3 +1,4 @@
+//go:build integration || interactive
```
jessicalins, line 1: I would still remove this line (as mentioned [here](https://github.com/observatorium/api/pull/246#discussion_r826815823))

test/e2e/helpers.go
```diff
@@ -1,3 +1,4 @@
+//go:build integration || interactive
```
jessicalins, line 1: same here

jessicalins (commented):
looking good, just a suggestion to remove the new tag syntax + Matej's comments

saswatamcode (approved):

matej-g (approved):
Thanks @esnible! This is looking great!

jessicalins (approved):
LGTM! Thanks for this @esnible

PR #251, opened 2022-03-15
test

PhilipGough:


test/e2e/services.go
```diff
@@ -1,11 +1,14 @@
+//go:build integration || interactive
```
matej-g, line 1: Since we're not yet on 1.17, we don't use the new syntax

test/e2e/services.go
```diff
@@ -53,10 +56,40 @@ func startServicesForMetrics(t *testing.T, e e2e.Environment) (
...
+		// https://docs.min.io/docs/minio-kms-quickstart-guide.html
+		"MINIO_KMS_KES_ENDPOINT=" + "https://play.min.io:7373",
+		"MINIO_KMS_KES_KEY_FILE=" + "root.key",
+		"MINIO_KMS_KES_CERT_FILE=" + "root.cert",
+		"MINIO_KMS_KES_KEY_NAME=" + "my-minio-key",
+	}
+	f := e2e.NewInstrumentedRunnable(e, "rules-minio", ports, e2edb.AccessPortName)
+	runnable := f.Init(
+		e2e.StartOptions{
+			Image: "minio/minio:RELEASE.2022-03-03T21-21-16Z",
```
matej-g, line 75: Why not latest release?

matej-g (commented):
Thanks for this @PhilipGough! This will be good to unblock us in here, would be super cool if we can fix this in the upstream as well :heart:.

test/e2e/services.go
PhilipGough: @matej-g did you mean `:latest` or that I should have picked up the latest release. If it is the latter, I think you are correct and have done so in https://github.com/efficientgo/e2e/pull/26

I initially tested with the release I felt was going to fix the issue.

matej-g: I meant whatever is the latest tagged release, not `latest`, if I seen correctly there is a newer version. Thanks @PhilipGough :+1:

PR #252, opened 2022-03-16
Trace query API

esnible:
This PR adds proxying of the Jaeger V2 HTTP query API and the Jaeger UI assets.

With this change, the Jaeger UI is available at http://localhost:8080/api/traces/v1/test-oidc/search/

Note that the first time logging in, it redirects away from traces.  Not sure if that is correct or not.

Note that there are currently no multi-tenant backends for Jaeger, so the UI currently sees all traces.  The work to divide the tenants will be done in Jaeger itself or another front-end, probably not in observatorium/api.

### DIFF
--- a/api/traces/v1/http.go
+++ b/api/traces/v1/http.go
@@ -91,8 +91,11 @@ func (n nopInstrumentHandler) NewHandler(labels prometheus.Labels, handler http.
 	return handler.ServeHTTP
 }
 
-// NewV2APIHandler creates a trace query handler for Jaeger V2 HTTP queries
-func NewV2APIHandler(read *url.URL, opts ...HandlerOption) http.Handler {
+// NewV2Handler creates a trace handler for Jaeger V2 API, web UI, and web UI static content
+// The web UI handler is able to rewrite
+// HTML to change the <base> attribute so that it works with the Observatorium-style
+// "/api/v1/traces/{tenant}/" URLs.
+func NewV2Handler(read *url.URL, opts ...HandlerOption) http.Handler {
 	if read == nil {
 		panic("missing Jaeger read url")
 	}
@@ -111,7 +114,7 @@ func NewV2APIHandler(read *url.URL, opts ...HandlerOption) http.Handler {
 
 	var proxyRead http.Handler
 	{
-		level.Debug(c.logger).Log("msg", "Configuring upstream Jaeger query v2", "queryv2", read)
+		level.Debug(c.logger).Log("msg", "Configuring upstream Jaeger", "queryv2", read)
 		middlewares := proxy.Middlewares(
 			proxy.MiddlewareSetUpstream(read),
 			proxy.MiddlewareSetPrefixHeader(),
@@ -129,117 +132,31 @@ func NewV2APIHandler(read *url.URL, opts ...HandlerOption) http.Handler {
 			Director:  middlewares,
 			ErrorLog:  proxy.Logger(c.logger),
 			Transport: otelhttp.NewTransport(t),
-		}
-	}
 
-	r.Group(func(r chi.Router) {
-		r.Use(c.readMiddlewares...)
-		r.Get("/traces*", c.instrument.NewHandler(
-			prometheus.Labels{"group": "tracesv1api", "handler": "api"},
-			proxyRead))
-		r.Get("/services*", c.instrument.NewHandler(
-			prometheus.Labels{"group": "tracesv1api", "handler": "api"},
-			proxyRead))
-		r.Get("/dependencies*", c.instrument.NewHandler(
-			prometheus.Labels{"group": "tracesv1api", "handler": "api"},
-			proxyRead))
-	})
-
-	return r
-}
-
-// NewUIStaticHandler creates a trace handler for Jaeger UI static assets (.js and .css)
-func NewUIStaticHandler(read *url.URL, opts ...HandlerOption) http.Handler {
-	c := &handlerConfiguration{
-		logger:     log.NewNopLogger(),
-		registry:   prometheus.NewRegistry(),
-		instrument: nopInstrumentHandler{},
-	}
-
-	for _, o := range opts {
-		o(c)
-	}
-
-	r := chi.NewRouter()
-
-	var proxyRead http.Handler
-	{
-		level.Debug(c.logger).Log("msg", "Configuring upstream Jaeger UI static", "ui", read)
-		middlewares := proxy.Middlewares(
-			proxy.MiddlewareSetUpstream(read),
-			proxy.MiddlewareSetPrefixHeader(),
-			proxy.MiddlewareLogger(c.logger),
-			proxy.MiddlewareMetrics(c.registry, prometheus.Labels{"proxy": "tracesv1-ui-static"}),
-		)
-
-		t := &http.Transport{
-			DialContext: (&net.Dialer{
-				Timeout: ReadTimeout,
-			}).DialContext,
-		}
-
-		proxyRead = &httputil.ReverseProxy{
-			Director:  middlewares,
-			ErrorLog:  proxy.Logger(c.logger),
-			Transport: otelhttp.NewTransport(t),
-		}
-	}
-	r.Group(func(r chi.Router) {
-		r.Use(c.readMiddlewares...)
-		r.Get("/*", c.instrument.NewHandler(
-			prometheus.Labels{"group": "tracesv1ui", "handler": "search"},
-			proxyRead))
-	})
-
-	return r
-}
-
-// NewUIHandler creates a trace handler for Jaeger UI that is able to re-write
-// HTML to change the <base> attribute so that it works with the Observatorium-style
-// "/api/v1/traces/{tenant}/" URLs.
-func NewUIHandler(read *url.URL, opts ...HandlerOption) http.Handler {
-	c := &handlerConfiguration{
-		logger:     log.NewNopLogger(),
-		registry:   prometheus.NewRegistry(),
-		instrument: nopInstrumentHandler{},
-	}
-
-	for _, o := range opts {
-		o(c)
-	}
-
-	r := chi.NewRouter()
-
-	var proxyRead http.Handler
-	{
-		level.Debug(c.logger).Log("msg", "Configuring upstream Jaeger UI", "ui", read)
-		middlewares := proxy.Middlewares(
-			proxy.MiddlewareSetUpstream(read),
-			proxy.MiddlewareSetPrefixHeader(),
-			proxy.MiddlewareLogger(c.logger),
-			proxy.MiddlewareMetrics(c.registry, prometheus.Labels{"proxy": "tracesv1-ui"}),
-		)
-
-		t := &http.Transport{
-			DialContext: (&net.Dialer{
-				Timeout: ReadTimeout,
-			}).DialContext,
-		}
-
-		proxyRead = &httputil.ReverseProxy{
-			Director:  middlewares,
-			ErrorLog:  proxy.Logger(c.logger),
-			Transport: otelhttp.NewTransport(t),
-
-			// This is the key piece, it changes <base href=> tags
+			// This is a key piece, it changes <base href=> tags on text/html content
 			ModifyResponse: jaegerUIResponseModifier,
 		}
 	}
 
 	r.Group(func(r chi.Router) {
 		r.Use(c.readMiddlewares...)
-		r.Get("/*", c.instrument.NewHandler(
-			prometheus.Labels{"group": "tracesv1ui", "handler": "search"},
+		r.Get("/api/traces*", c.instrument.NewHandler(
+			prometheus.Labels{"group": "tracesv1api", "handler": "traces"},
+			proxyRead))
+		r.Get("/api/services*", c.instrument.NewHandler(
+			prometheus.Labels{"group": "tracesv1api", "handler": "services"},
+			proxyRead))
+		r.Get("/api/dependencies*", c.instrument.NewHandler(
+			prometheus.Labels{"group": "tracesv1api", "handler": "dependencies"},
+			proxyRead))
+		r.Get("/static/*", c.instrument.NewHandler(
+			prometheus.Labels{"group": "tracesv1static", "handler": "ui"},
+			proxyRead))
+		r.Get("/search*", c.instrument.NewHandler(
+			prometheus.Labels{"group": "tracesv1ui", "handler": "ui"},
+			proxyRead))
+		r.Get("/favicon.ico", c.instrument.NewHandler(
+			prometheus.Labels{"group": "tracesv1ui", "handler": "ui"},
 			proxyRead))
 	})
 
@@ -247,7 +164,7 @@ func NewUIHandler(read *url.URL, opts ...HandlerOption) http.Handler {
 }
 
 func jaegerUIResponseModifier(response *http.Response) error {
-	// Only modify successful HTTP
+	// Only modify successful HTTP with HTML content
 	if response.StatusCode == http.StatusOK && strings.HasPrefix(response.Header.Get("Content-Type"), "text/html") {
 		// Do man-in-the-middle rewriting of the UI HTML.
 		var err error
--- a/api/traces/v1/http.go
+++ b/api/traces/v1/http.go
@@ -1,11 +1,17 @@
 package v1
 
 import (
+	"bytes"
+	"compress/flate"
+	"compress/gzip"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/go-chi/chi"
@@ -182,8 +188,9 @@ func NewUIHandler(read *url.URL, opts ...HandlerOption) http.Handler {
 			ErrorLog:  proxy.Logger(c.logger),
 			Transport: otelhttp.NewTransport(t),
 			ErrorHandler: func(rw http.ResponseWriter, r *http.Request, e error) {
-				fmt.Printf("@@@ ecs in NewUIHandler anon ErrorHandler for request %#v\n", r)
+				fmt.Printf("@@@ ecs in NewUIHandler anon ErrorHandler for request %q: %v\n", r.URL.String(), e)
 			},
+			ModifyResponse: jaegerUIResponseModifier,
 		}
 	}
 	r.Group(func(r chi.Router) {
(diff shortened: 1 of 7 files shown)


api/traces/v1/http.go
```diff
@@ -0,0 +1,291 @@
...
+	"github.com/go-kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+
+	"github.com/observatorium/api/proxy"
+)
+
+const (
+	ReadTimeout  = 15 * time.Minute
+	WriteTimeout = time.Minute
```
jessicalins, line 28: is `WriteTimeout` being used somewhere?

api/traces/v1/http.go
```diff
@@ -0,0 +1,291 @@
...
+type HandlerOption func(h *handlerConfiguration)
+
+// Logger add a custom logger for the handler to use.
+func Logger(logger log.Logger) HandlerOption {
+	return func(h *handlerConfiguration) {
+		h.logger = logger
+	}
+}
+
+// WithJaegerQueryV3 adds a custom Jaeger query for the handler to use.
```
jessicalins, line 50: I think we could update the comment to `// WithRegistry ...`

api/traces/v1/http.go
```diff
@@ -0,0 +1,291 @@
...
+		proxyRead = &httputil.ReverseProxy{
+			Director:  middlewares,
+			ErrorLog:  proxy.Logger(c.logger),
+			Transport: otelhttp.NewTransport(t),
+		}
+	}
+	r.Group(func(r chi.Router) {
+		r.Use(c.readMiddlewares...)
+		r.Handle("/*", c.instrument.NewHandler(
+			prometheus.Labels{"group": "tracesv1ui", "handler": "search"},
```
jessicalins, line 181: since `tracesv1ui` and `search` are reused in `NewUIHandler` maybe they could be moved to consts?

api/traces/v1/http.go
```diff
@@ -0,0 +1,291 @@
...
+			}).DialContext,
+		}
+
+		proxyRead = &httputil.ReverseProxy{
+			Director:  middlewares,
+			ErrorLog:  proxy.Logger(c.logger),
+			Transport: otelhttp.NewTransport(t),
+
+			// This is the key piece, it changes <base href=> tags
+			ModifyResponse: jaegerUIResponseModifier,
```
jessicalins, line 226: I wonder what happens here in case `jaegerUIResponseModifier` returns an error?
should `ErrorHandler` be implemented as well https://go.dev/src/net/http/httputil/reverseproxy.go ?

api/traces/v1/http.go
```diff
@@ -0,0 +1,291 @@
...
+		// At this point we have read the body.  Even if it didn't have a <body href=>
+		// to modify, we need to create a new Reader.  This code thus executes all the time.
+
+		// JaegerUI insists on a <base>, so create one but use Observatorium's
+		// opinion of the base href, not Jaeger Query's opinion.
+		forwardedPrefix := response.Request.Header.Get(proxy.PrefixHeader)
+
+		// The <base href=> tag generated by Jaeger to tell the UI where to fetch static
+		// assets and query /api
+		const expectedBaseTag = `<base href="/" data-inject-target="BASE_URL"/>`
```
jessicalins, line 278: is Jaeger used by Obs API in a specific version? or would we expect the `<base href=..>` tag to be changed at some point?

main.go
```diff
@@ -607,6 +608,63 @@ func main() {
 					)
 				})
 			}
+
+			// Traces.
+			if cfg.traces.enabled {
```
jessicalins, line 613: :tada:

jessicalins (commented):
looking good, thanks @esnible - just a few suggestions/questions from my side
I think would be nice to also add a `traces_test.go` ?

api/traces/v1/http.go
```diff
@@ -0,0 +1,291 @@
...
+
+// Logger add a custom logger for the handler to use.
+func Logger(logger log.Logger) HandlerOption {
+	return func(h *handlerConfiguration) {
+		h.logger = logger
+	}
+}
+
+// WithJaegerQueryV3 adds a custom Jaeger query for the handler to use.
+func WithRegistry(r *prometheus.Registry) HandlerOption {
```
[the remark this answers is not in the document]
matej-g: I also feel like there's quiet some code repetition between metrics / logs / traces for these common handler options. We might want to refactor this, but I'd leave that out to a separate stream of work, just putting the idea out there.

proxy/proxy.go
```diff
@@ -18,7 +18,7 @@ type contextKey string
 const (
 	prefixKey contextKey = "prefix"
 
-	prefixHeader string = "X-Forwarded-Prefix"
```
matej-g, line 21: Any reason to export this in here?

main.go
```diff
@@ -607,6 +608,63 @@ func main() {
...
+								tracesv1.WithHandlerInstrumenter(ins),
+								tracesv1.WithSpanRoutePrefix("/api/traces/v1/{tenant}"),
+								tracesv1.WithReadMiddleware(authorization.WithAuthorizers(authorizers, rbac.Read, "traces")),
+								tracesv1.WithReadMiddleware(logsv1.WithEnforceAuthorizationLabels()),
+								tracesv1.WithWriteMiddleware(authorization.WithAuthorizers(authorizers, rbac.Write, "traces")),
+							),
+						),
+					)
+
+					// Static UI (.js and .css), not protected by RBAC -- any user can see UI
```
matej-g, line 634: I'm wondering if there's any reason we do not want to enforce authorization specifically on UI?

Also I'm wondering if following the convention we have with other signals would make sense here as well, i.e. having everything organize under one handler instead of splitting it into API, UI etc.

api/traces/v1/http.go
esnible: I am trying to follow the style of https://github.com/observatorium/api/blob/main/api/metrics/v1/http.go#L167 where the labels are explicit.  I like that style; it is very readable.  The compiler will optimize the literals together.

jessicalins: sounds good :+1:

api/traces/v1/http.go
esnible: The Jaeger API is versioned, but the UI is not.  This handler is for the UI.

I want to bring @pavolloffay into this discussion.  Jaeger does something weird, which was introduced in https://github.com/jaegertracing/jaeger-ui/issues/42 .  The introduced `<base>` to support reverse proxies.  I don't like their solution and I am doing something of the opposite.

To support some reverse proxy, Jaeger introduced a mechanism where the path known to the proxy is told to Jaeger.  So if I have two tenants, `tenant1` and `tenant2`, I can start a Jaeger instance that is told the base URL is `/api/traces/v1/tenant1` and another that is told the base URL is `/api/traces/v1/tenant2`.  The reverse proxy is responsible for remembering the host (or IP) of each tenant and forwarding properly.

What Observatorium does with this PR is offer the endpoint {host}/api/traces/v1/{tenant}/{suffix} and proxy adding header `x-tenant: ...` to {jaeger}/{suffix}.  All queries are sent to the same `--traces.read.endpoint={host:port}`

The approach I am taking will be great when Jaeger has a multi-tenant back-end.  This is in-plan, see https://github.com/jaegertracing/jaeger/issues/3427 .  (We might need a temporary work-around to get to that approach, and I plan to start on that work next).

The other approach is that when we add multitenancy to Jaeger we have Jaeger itself alter the BASE_URL instead of doing it in `jaegerUIResponseModifier()`.  If that approach is taken we still don't need to version the `<base>` tag, we merely delete jaegerUIResponseModifier() from the Observatorium/API code.

@pavolloffay , Does the Jaeger community want to pass the tenant to the UI as a segment in the path or in a header?

api/traces/v1/http.go
```diff
@@ -0,0 +1,291 @@
...
+	return func(h *handlerConfiguration) {
+		h.writeMiddlewares = append(h.writeMiddlewares, m)
+	}
+}
+
+type handlerInstrumenter interface {
+	NewHandler(labels prometheus.Labels, handler http.Handler) http.HandlerFunc
+}
+
+type nopInstrumentHandler struct{}
```
pavolloffay, line 89: it is "mandatory" to have this type? I would avoid it if it is optional.

proxy/proxy.go
esnible: What I'm trying to do is have Observatorium replace `<base href="/">` with `<base href="/api/traces/v1/{tenant}/">` for the reasons I explain at https://github.com/observatorium/api/pull/252#discussion_r831156906

In the place I am doing this inside of httputil.ReverseProxy.ModifyResponse, and at this point the HTTP request Context doesn't have the prefix that it used.  I could try to reconstruct the value applied by `MiddlewareSetPrefixHeader()`, but instead of doing that I just lookup that value Observatorium supplied for `X-Forwarded-Prefix` which happens to be exactly the value /api/traces/v1/{tenant}/ that I need to make the UI work.

I don't need to get the value from here, I could just use the literal `"X-Forwarded-Prefix"` myself.  Or we can talk about the whole approach.  What is your preference?

main.go
esnible: If I go to http://localhost:8080/api/traces/v1/test-oidc/search/ I am redirected to http://localhost:5556/dex/auth/local/login?back=&state=vcqh2yns46maemaiyfh2k33xm where I am presented with a Dex login screen.  So we do enforce **authentication** through the `r.Use(authentication.WithAccessToken())` earlier in the function.

I felt like the RBAC permission for traces for for reading and writing traces, not UI.  By not doing RBAC the UI comes up the UI tries to fetch trace "service names", and the user sees the Unauthorized message within a nice-looking Jaeger UI.  Do you want me to add RBAC here to block the UI from coming up at all?

I originally used a single handler until I chose to use `httputil.ReverseProxy.ModifyResponse`.  I didn't want that logic to run over the /static resources of the Jaeger UI, only the generated HTML.  I split into two handlers to make that logic more clear.

proxy/proxy.go
```diff
@@ -32,6 +32,11 @@ func Middlewares(middlewares ...Middleware) func(r *http.Request) {
 }
 
 func MiddlewareSetUpstream(upstream *url.URL) Middleware {
+	// Verify that the middleware upstream is valid now; rather than failing during processing
+	if upstream == nil {
+		panic("no upstream")
```
jessicalins, line 37: @esnible I think this is what is failing in the tests now (https://app.circleci.com/pipelines/github/observatorium/api/515/workflows/541d2f45-b811-4c09-8548-1909ab818240/jobs/2530?invite=true#step-102-886 )

test/e2e/interactive_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build interactive
```
jessicalins, line 1: we need to remove this since obs API uses go 1.16

test/e2e/services.go
```diff
@@ -1,3 +1,4 @@
+//go:build integration || interactive
```
jessicalins, line 1: same here

api/traces/v1/http.go
jessicalins: so we don't have control if this tag gets updated in Jaeger UI right? I think then the response won't get properly modified?

test/e2e/interactive_test.go
esnible: I keep making this mistake.  Thank you for your patience.  My editor uses `gofmt` on save.  My new workflow is to hand-edit with `vi` before pushing, but I forgot.  I have fixed this in the latest push.

Note that the official guidance from Go in the [1.17 release notes](https://go.dev/doc/go1.17) says "... all Go files should be updated to have both forms with the same meaning."  Consider allowing both forms in Observatorium code, as long as both have the same meaning.

proxy/proxy.go
matej-g: I might have misunderstood. I only meant if we need have that constant accessible from other packages (i.e. the constant name is capitalized), if we don't use the constant outside of here.

main.go
esnible: By not doing RBAC for the UI resources we allow the UI to load and make a failing API query for traces.

If I remove RBAC permission for traces the UI for http://localhost:8080/api/traces/v1/test-oidc/search/ looks like this:

no-trace-rbac

I thought it was cute to allow it to work this way.  I am happy to secure the UI resources behind RBAC permission to read traces if that is your request.

matej-g: I'm just trying to think what's the most reasonable solution here, but I'm open to other opinions. I'm currently leaning more towards authorizing on all endpoints, since if the tenants is not authorized to read / write trace signals, there's no reason for the tenant to even be accessing the UI.

test/e2e/interactive_test.go
jessicalins: I think we can also update obs/api to go 1.17 - as far as I understand there should be no problem with that. I opened an issue here: https://github.com/observatorium/api/issues/262 
Regarding your flow/editor, it may be possible to configure it to use go v1.16 per project, but something to look into

main.go
jessicalins: I think it makes sense what @matej-g proposes - also it may avoid unnecessary loading of the UI + trial to fetch traces. So maybe is better to consistently to add authorization on all endpoints

api/traces/v1/http.go
pavolloffay: > @pavolloffay , Does the Jaeger community want to pass the tenant to the UI as a segment in the path or in a header?

There was no decision on this so far. But I think it will be a header to avoid changing the current APIs.


> What Observatorium does with this PR is offer the endpoint {host}/api/traces/v1/{tenant}/{suffix} and proxy adding header x-tenant: ... to {jaeger}/{suffix}. All queries are sent to the same --traces.read.endpoint={host:port}

As you mentioned this is the final goal but it does not work with the Jaeger at the moment as there is no (soft) multitenancy support in Jaeger. As a temporary solution the API service could use tenant name/id as part (prefix) of the endpoint for the read endpoint.

jessicalins (dismissed):
LGTM!

api/traces/v1/http.go
```diff
@@ -0,0 +1,293 @@
...
+
+func (n nopInstrumentHandler) NewHandler(labels prometheus.Labels, handler http.Handler) http.HandlerFunc {
+	return handler.ServeHTTP
+}
+
+// NewV2APIHandler creates a trace query handler for Jaeger V2 HTTP queries
+func NewV2APIHandler(read *url.URL, opts ...HandlerOption) http.Handler {
+	if read == nil {
+		panic("missing Jaeger read url")
```
matej-g, line 97: We should never really hit this, since we're checking if `read` is nil in the previous method in the chain, isn't it? I'm wondering whether we need the check here.

main.go
```diff
@@ -607,6 +608,66 @@ func main() {
...
 			}
+
+			// Traces.
+			if cfg.traces.enabled && cfg.traces.readEndpoint != nil {
+				r.Group(func(r chi.Router) {
+					r.Use(authentication.WithTenantMiddlewares(pm.Middlewares))
+					r.Use(authentication.WithTenantHeader(cfg.traces.tenantHeader, tenantIDs))
+
+					// V2 query API, protected by RBAC
+					r.Mount("/api/traces/v1/{tenant}/api",
```
matej-g, line 619: After looking closely, I'm wondering why we're departing here from how the paths are mounted for metrics and traces. If I'm looking at the code correctly, we're mounting these paths, but we're handling everything on them via `/*`. Why aren't we handling only specific paths?

api/traces/v1/http.go
esnible: The check is here because due to another bug (which was never committed) the caller passed `nil`.  This function passed that nil to `proxy.MiddlewareSetUpstream()`, which created a closure containing the nil.  The closure was later invoked in the scope of an HTTP interceptor, leading to a stack trace with no hint where the nil came from.

It has been my experience it is far better to catch invalid parameters near the call stack where they are made.

Your comment is correct.  Do you want me to remove the check?

main.go
esnible: Thanks for this comment.  I have restricted things.
Now the API handler only proxies GET for /api/traces, /api/dependencies, and /api/services.  The UI and static asset handlers still take /* but only proxy GET.

I am a bit confused about the code that I needed to do this.  I am doing `stripTenantPrefix("/api/traces/v1"()` then
```
		r.Get("/traces*", c.instrument.NewHandler(
			prometheus.Labels{"group": "tracesv1api", "handler": "api"},
			proxyRead))
```

I do not understand why it works when I use `r.Get("/traces*"...)`.  I had expected to be doing `r.Get("/api/traces*"...)

esnible:
Let me give testing instructions.

There are two ways to test this.  One way is to run `make test-interactive`.  This allows the API to be tested but not the UI (at least on the Mac).  (The reason is that Observatorium expects OIDC to have the same auth URL from the point of view of the browser and the point of view of Observatorium.  From the browser the auth URL needs to be something like "localhost:55000/login" and from Observatorium it needs to be something like "interactive-test-dex:5555/login".)

The fun, interactive way that I test is by running three programs in three windows:

```
docker run --rm --name dex --publish 5556:5556 --volume /tmp/dex.yaml:/tmp/dex.yaml --volume /tmp/dex.pem:/tmp/dex.pem --volume /tmp/dex.key:/tmp/dex.key dexidp/dex:v2.30.0 dex serve /tmp/dex.yaml
```

```
cd ~/go/src/github.com/jaegertracing/jaeger
ADMIN_HTTP_HOST_PORT=localhost:14269 HTTP_SERVER_HOST_PORT=localhost:5778 COLLECTOR_GRPC_SERVER_HOST_PORT=localhost:14250 COLLECTOR_HTTP_SERVER_HOST_PORT=localhost:14268 QUERY_HTTP_SERVER_HOST_PORT=localhost:16686 QUERY_GRPC_SERVER_HOST_PORT=localhost:16685 PROCESSOR_JAEGER_BINARY_SERVER_HOST_PORT=localhost:9999 QUERY_UI_CONFIG=/tmp/ui.json make run-all-in-one
```

```
cd ~/go/src/github.com/observatorium/api 
go run main.go -tenants.config=/tmp/http-dex-tenants-email-claim.yaml --rbac.config=/tmp/rbac.yaml -traces.write.endpoint localhost:4317 -grpc.listen localhost:8317 -traces.read.endpoint http://localhost:16686 -logs.write.endpoint localhost:8080 -web.internal.listen localhost:8081 -web.listen localhost:8080 -log.level debug
```

(The config files needed for the above can be copied out of a `make test-interactive` _e2e_ folder.)

api/traces/v1/http.go
matej-g: Not a must from my side, I'm fine with it :+1:

matej-g (approved):
This is looking good to me now, stellar job @esnible! :star:

PR #270, opened 2022-03-31
Add OpenAPI spec and client

### LINKED ISSUE #139: Add OpenAPI Specification
It would be super useful to have these specifications to generate SDKs and client codes
https://swagger.io/specification/

saswatamcode:
This PR adds an OpenAPI spec for the Observatorium API (only metrics for now) as well as a generated client.

As discussed in https://github.com/observatorium/obsctl/pull/2. this repo is better suited to house this client, as other users might also want to use the spec to generate tooling of their own against the Observatorium API.

Fixes #139.


PhilipGough:
@saswatamcode thanks for the lift and shift. One thought that comes to mind is that perhaps we don't even need to generate the client and commit that here. `obsctl` could take care of that since it needs the code, but I have no strong opinions on it, so let's see what others think.

periklis:
@saswatamcode Any limitations I can help out on why not expand this to the logs API in this PR too?

saswatamcode:
@periklis sure! I'm not too familiar with the logging bits of the API, so would appreciate the help! We can extend this to obsctl too! 🙂

@PhilipGough, I'm not sure how good the support is for importing specs from external repos in https://github.com/deepmap/oapi-codegen, which is why I generated the client here. But if it's possible, I have no strong opinion on this either. 🙂

PhilipGough:
Closes #139

saswatamcode:
@periklis how about we add the Logs API spec/client in a separate parallel PR, as this one is already kind of large? 🙂

client/spec.yaml
```diff
@@ -0,0 +1,286 @@
...
+  license:
+    name: Apache 2.0
+    url: "http://www.apache.org/licenses/LICENSE-2.0.html"
+tags:
+  - name: observatoriumv1
+    description: Calls related to Observatorium
+# Keeping all the API paths here, but likely need to separate it out to another file,
+# once paths for other signals are also added.
+paths:
+  /api/metrics/v1/{tenant}/api/v1/rules/raw:
```
esnible, line 15: I do not understand the meaning of the word "raw" here.

Do we have two formats for rules?  Under which circumstance would a client desire raw vs non-raw?

saswatamcode: We refer to the YAML format of the Prometheus Rules as "raw", as this is, in a sense, unprocessed. 

When these Rules are configured on a component like Thanos Ruler, they are "processed", i.e, some health data like "lastEvaluatedTime" or "lastError" is added to it which we can retrieve in JSON format via API call. Also, in the case of alerting Rules, we can get the alerts fired from the Rule and there are also some templated labels that are parsed! 

The circumstances under which this distinction is important for a client is when using the Rules API.
The client can configure their YAML Rules via the /rules/raw endpoint by sending a PUT request to Observatorium API and can see what is configured by sending a GET request to the same endpoint.

They can then retrieve health data + alerts or "processed" Rules, that they configured, in JSON format by sending a GET request to the /rules endpoint (Observatorium API just proxies request to Thanos Querier in this case, which gets this information from Thanos Ruler). 

Hope that makes it a bit clear! 🙂

PhilipGough:
@saswatamcode I agree keeping the scope narrow here and adding logs spec in a follow-up is likely a good idea.
@saswatamcode looking good overall, what do you think about adding the codegen to the `make generate` target so we don't end up with drift that is not checked by CI?

saswatamcode:
@PhilipGough sure le me add that!

PhilipGough (approved):
lgtm

client/spec.yaml
```diff
@@ -0,0 +1,286 @@
...
+    query:
+      $ref: ./parameters.yaml#/components/parameters/query
+    queryTimeout:
+      $ref: ./parameters.yaml#/components/parameters/queryTimeout
+  schemas:
+    # Responses from API.
+    SeriesResponse:
+      $ref: ./responses/responses.yaml#/components/schemas/SeriesResponse
+    QueryResponse:
+      $ref: ./responses/responses.yaml#/components/schemas/QueryResponse
```
squat, line 249: right now, the path definitions have refs that point down here and these schemas have refs that point elsewhere.
why are we doing double refs here instead of putting all of these refs directly above where they are actually used?

saswatamcode: Good catch! Let me try and clean this up a bit! :)

saswatamcode: Thanks for pointing this out @squat! Most of the imports weren't required or could be done directly, so I cleaned it up and simplified it further in https://github.com/observatorium/api/pull/270/commits/e696f2bff00ed8880766382774a7dbce9c14074c. 🙂

saswatamcode:
Would love to move this forward, as we plan to use it in obsctl https://github.com/observatorium/obsctl/pull/2 🙂

periklis (commented):
One question from my side, in general LGTM. Do we miss the `403`s in the spec.yaml ?

saswatamcode:
@periklis adding the response code like 403, doesn't make a difference in the generated client as such, plus few already have covered under `4XX`. But let me add it where it is missing! :)

periklis:
> @periklis adding the response code like 403, doesn't make a difference in the generated client as such, plus few already have covered under `4XX`. But let me add it where it is missing! :)

Sure I got it in the places where `4XX` was used, but others listed only 401 and 500. This was more a nit suggestion rather than a blocker. The observatorium's api sole purpose is to provide solid and uniform 401 and 403 behavior in-front of Thanos/Loki/Jaeger. Thus being explicit in the spec never hurts :smile_cat:

onprem (approved):
LGTM as well!

Issue #268, opened 2022-03-30
Trace Read tenancy architecture and implementation


esnible:
Now that https://github.com/observatorium/api/pull/252 has merged we can query traces through Observatorium.

Unfortunately, we don't have a full story for queries because Jaeger doesn't support tenancy yet.  (For writing traces we handle tenancy at the OTel collector level, see https://github.com/observatorium/observatorium/pull/461 ).

In order to make https://github.com/observatorium/api/issues/267 meaningful we need to decide how to handle tenancy for queries.  I see three approaches.  I would like to start discussion for the correct approach.

- Do nothing, and wait for the Jaeger community to implement https://github.com/jaegertracing/jaeger/issues/3427
- Allow templates beyond simple hostnames in `-trace.read.endpoint`.  Using this approach we could say `-trace.read.endpoint="jaeger-{tenant}:14250", and Observatorium would send traces appropriately to multiple Jaeger instances.
- Create an additional reverse proxy which forwards the same way, perhaps by generating an Envoy or Nginx config from JSONNET that would take the role for queries that OTel takes for posting traces.

@pavolloffay has suggested the template endpoint approach would be best.  (When Jaeger gets multi-tenancy we would simply stop supplying templates to `-trace.read.endpoint`.)

I'd like to get this in before https://github.com/observatorium/api/issues/267 so that I can test properly.

matej-g:
Given the current situation without native multi-tenancy features and considering the fact that we'll need to have a Jaeger instance per tenant either way, the second option also seems like the simplest way to go for me.

Issue #267, opened 2022-03-30
Expose flag traces.read.endpoint in jsonnet


pavolloffay:
Follow up https://github.com/observatorium/api/pull/252 and similar to https://github.com/observatorium/api/pull/247/files#diff-d79e05fae1c17a80c15dfd8e05409c355f8eb20c179667763c26ef6351c18fbfR135

The flag `--traces.read.endpoint` needs to be exposed in jsonnet to accordingly configure API service.

pavolloffay:
@esnible would you like to take a look at this?

pavolloffay:
After the change is implemented the API jsonnet in https://github.com/observatorium/observatorium needs to be upgraded and k8s resources regenerated with the API read endpoint.

PR #271, opened 2022-03-31
Pseudo-tenancy for trace queries

esnible:
Signed-off-by: Ed Snible <snible@us.ibm.com>

Resolves https://github.com/observatorium/api/issues/267 allowing Jaeger query hosts to be specified in Jsonnet.

Resolves https://github.com/observatorium/api/issues/268 by allowing the `-traces.read.endpoint` to describe multiple single-tenant Jaeger instances.  For example, we can say `-traces.read.endpoint=http://jaeger-{tenant}-query:16686`.  (This uses the same convention we use in the configuration of the OTel processor.)

With this implementation, Observatorium can provide all tenants with their own Jaeger instance.  In the future, when Jaeger itself supports tenancy, we simply change the read endpoint to point to the multi-tenant Jaeger.

cc @pavolloffay

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,8 @@ Usage of ./observatorium-api:
     	The percentage of mutex contention events that are reported in the mutex profile. (default 10)
   -debug.name string
     	A name to add as a prefix to log lines. (default "observatorium")
+  -experimental.traces.read.endpoint-template string
+    	A template replacing --read.traces.endpoint, such as http://jaeger-{tenant}-query:16686
   -grpc.listen string
     	The address on which the public gRPC server listens.
   -internal.tracing.endpoint string
--- a/api/traces/v1/http.go
+++ b/api/traces/v1/http.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"path"
 	"strings"
 	"time"
 
@@ -20,6 +21,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 
+	"github.com/observatorium/api/authentication"
 	"github.com/observatorium/api/proxy"
 )
 
@@ -95,8 +97,8 @@ func (n nopInstrumentHandler) NewHandler(labels prometheus.Labels, handler http.
 // The web UI handler is able to rewrite
 // HTML to change the <base> attribute so that it works with the Observatorium-style
 // "/api/v1/traces/{tenant}/" URLs.
-func NewV2Handler(read *url.URL, opts ...HandlerOption) http.Handler {
-	if read == nil {
+func NewV2Handler(read string, opts ...HandlerOption) http.Handler {
+	if read == "" {
 		panic("missing Jaeger read url")
 	}
 
@@ -116,7 +118,7 @@ func NewV2Handler(read *url.URL, opts ...HandlerOption) http.Handler {
 	{
 		level.Debug(c.logger).Log("msg", "Configuring upstream Jaeger", "queryv2", read)
 		middlewares := proxy.Middlewares(
-			proxy.MiddlewareSetUpstream(read),
+			middlewareSetTemplatedUpstream(c.logger, read),
 			proxy.MiddlewareSetPrefixHeader(),
 			proxy.MiddlewareLogger(c.logger),
 			proxy.MiddlewareMetrics(c.registry, prometheus.Labels{"proxy": "tracesv1-read"}),
@@ -161,6 +163,55 @@ func NewV2Handler(read *url.URL, opts ...HandlerOption) http.Handler {
 	})
 
 	return r
+}
+
+// Parse a URL; if the URL includes `{tenant}` that portion will be replaced by the tenant
+func ExpandTemplatedUpstream(templateUpstream, tenant string) (*url.URL, error) {
+	rawTracesReadEndpoint := strings.Replace(templateUpstream, "{tenant}", tenant, 1)
+	return url.ParseRequestURI(rawTracesReadEndpoint)
+}
+
+// middlewareSetTemplatedUpstream is a variation of proxy.MiddlewareSetUpstream()
+// with additional processing if the upstream includes "{tenant}"
+func middlewareSetTemplatedUpstream(logger log.Logger, upstreamTemplate string) proxy.Middleware {
+	// If the upstream is a URL, parse it once
+	upstream, err := url.ParseRequestURI(upstreamTemplate)
+	if err == nil {
+		return func(r *http.Request) {
+			r.URL.Scheme = upstream.Scheme
+			r.URL.Host = upstream.Host
+			r.URL.Path = path.Join(upstream.Path, r.URL.Path)
+		}
+	}
+
+	// The upstream is not a URL.  We checked at startup that `-traces.read.endpoint`
+	// was either a URL or a templated URL, so we must have a templated URL.
+
+	// Cache upstream URLs to avoid re-parse on every read
+	templateToURL := map[string]*url.URL{}
+
+	return func(r *http.Request) {
+		tenant, ok := authentication.GetTenant(r.Context())
+		if !ok {
+			// At this point another middleware must have put the tenant into the context
+			level.Debug(logger).Log("msg", "Internal error; expected tenant in request context")
+		}
+
+		upstream, ok := templateToURL[tenant]
+		if !ok {
+			upstream, err = ExpandTemplatedUpstream(upstreamTemplate, tenant)
+			if err != nil {
+				// Log if the tenant label includes characters that can't appear in a hostname (such as punctuation)
+				level.Debug(logger).Log("msg", "Internal error; tenant contains characters that cannot appear in hostname")
+			}
+
+			templateToURL[tenant] = upstream
+		}
+
+		r.URL.Scheme = upstream.Scheme
+		r.URL.Host = upstream.Host
+		r.URL.Path = path.Join(upstream.Path, r.URL.Path)
+	}
 }
 
 func jaegerUIResponseModifier(response *http.Response) error {
--- a/api/traces/v1/http.go
+++ b/api/traces/v1/http.go
@@ -190,7 +190,8 @@ func middlewareSetUpstream(logger log.Logger, read *url.URL, readTemplate string
 
 		upstream, ok := templateToURL[tenant]
 		if !ok {
-			upstream, err := ExpandTemplatedUpstream(readTemplate, tenant)
+			var err error
+			upstream, err = ExpandTemplatedUpstream(readTemplate, tenant)
 			if err != nil {
 				// Log if the tenant label includes characters that can't appear in a hostname (such as punctuation)
 				level.Debug(logger).Log("msg", "Internal error; tenant contains characters that cannot appear in hostname")
--- a/api/traces/v1/http.go
+++ b/api/traces/v1/http.go
@@ -118,7 +118,7 @@ func NewV2Handler(read *url.URL, readTemplate string, opts ...HandlerOption) htt
 	{
 		level.Debug(c.logger).Log("msg", "Configuring upstream Jaeger", "queryv2", read)
 		middlewares := proxy.Middlewares(
-			middlewareSetUpstream(c.logger, read, readTemplate),
+			middlewareSetTemplatedUpstream(c.logger, read, readTemplate),
 			proxy.MiddlewareSetPrefixHeader(),
 			proxy.MiddlewareLogger(c.logger),
 			proxy.MiddlewareMetrics(c.registry, prometheus.Labels{"proxy": "tracesv1-read"}),
@@ -165,26 +165,26 @@ func NewV2Handler(read *url.URL, readTemplate string, opts ...HandlerOption) htt
 	return r
 }
 
-// Parse a URL; if the URL includes `{tenant}` that portion will be replaced by the tenant
+// Parse a URL; if the URL includes `{tenant}` that portion will be replaced by the tenant.
 func ExpandTemplatedUpstream(templateUpstream, tenant string) (*url.URL, error) {
 	rawTracesReadEndpoint := strings.Replace(templateUpstream, "{tenant}", tenant, 1)
 	return url.ParseRequestURI(rawTracesReadEndpoint)
 }
 
 // middlewareSetTemplatedUpstream is a variation of proxy.MiddlewareSetUpstream()
-// with additional processing if the upstream includes "{tenant}"
-func middlewareSetUpstream(logger log.Logger, read *url.URL, readTemplate string) proxy.Middleware {
+// with additional processing if the upstream includes "{tenant}".
+func middlewareSetTemplatedUpstream(logger log.Logger, read *url.URL, readTemplate string) proxy.Middleware {
 	if read != nil {
 		return proxy.MiddlewareSetUpstream(read)
 	}
 
-	// Cache upstream URLs to avoid re-parse on every read
+	// Cache upstream URLs to avoid re-parse on every read.
 	templateToURL := map[string]*url.URL{}
 
 	return func(r *http.Request) {
 		tenant, ok := authentication.GetTenant(r.Context())
 		if !ok {
-			// At this point another middleware must have put the tenant into the context
+			// At this point another middleware must have put the tenant into the context.
 			level.Debug(logger).Log("msg", "Internal error; expected tenant in request context")
 		}
 
@@ -193,7 +193,7 @@ func middlewareSetUpstream(logger log.Logger, read *url.URL, readTemplate string
 			var err error
 			upstream, err = ExpandTemplatedUpstream(readTemplate, tenant)
 			if err != nil {
-				// Log if the tenant label includes characters that can't appear in a hostname (such as punctuation)
+				// Log if the tenant label includes characters that can't appear in a hostname (such as punctuation).
 				level.Debug(logger).Log("msg", "Internal error; tenant contains characters that cannot appear in hostname")
 			}
 
--- a/api/traces/v1/http.go
+++ b/api/traces/v1/http.go
@@ -117,8 +117,16 @@ func NewV2Handler(read *url.URL, readTemplate string, opts ...HandlerOption) htt
 	var proxyRead http.Handler
 	{
 		level.Debug(c.logger).Log("msg", "Configuring upstream Jaeger", "queryv2", read)
+
+		var upstreamMiddleware proxy.Middleware
+		if read != nil {
+			upstreamMiddleware = proxy.MiddlewareSetUpstream(read)
+		} else {
+			upstreamMiddleware = middlewareSetTemplatedUpstream(c.logger, readTemplate)
+		}
+
 		middlewares := proxy.Middlewares(
-			middlewareSetTemplatedUpstream(c.logger, read, readTemplate),
+			upstreamMiddleware,
 			proxy.MiddlewareSetPrefixHeader(),
 			proxy.MiddlewareLogger(c.logger),
 			proxy.MiddlewareMetrics(c.registry, prometheus.Labels{"proxy": "tracesv1-read"}),
@@ -173,11 +181,7 @@ func ExpandTemplatedUpstream(templateUpstream, tenant string) (*url.URL, error) 
 
 // middlewareSetTemplatedUpstream is a variation of proxy.MiddlewareSetUpstream()
 // with additional processing if the upstream includes "{tenant}".
-func middlewareSetTemplatedUpstream(logger log.Logger, read *url.URL, readTemplate string) proxy.Middleware {
-	if read != nil {
-		return proxy.MiddlewareSetUpstream(read)
-	}
-
+func middlewareSetTemplatedUpstream(logger log.Logger, readTemplate string) proxy.Middleware {
 	// Cache upstream URLs to avoid re-parse on every read.
 	templateToURL := map[string]*url.URL{}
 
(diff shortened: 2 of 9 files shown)


pavolloffay:
Should we document the tenant template in the read endpoint (flag) help message?

pavolloffay (dismissed):

api/traces/v1/http.go
```diff
@@ -163,6 +165,55 @@ func NewV2Handler(read *url.URL, opts ...HandlerOption) http.Handler {
...
 
+// Parse a URL; if the URL includes `{tenant}` that portion will be replaced by the tenant
+func ExpandTemplatedUpstream(templateUpstream, tenant string) (*url.URL, error) {
+	rawTracesReadEndpoint := strings.Replace(templateUpstream, "{tenant}", tenant, 1)
+	return url.ParseRequestURI(rawTracesReadEndpoint)
+}
+
+// middlewareSetTemplatedUpstream is a variation of proxy.MiddlewareSetUpstream()
+// with additional processing if the upstream includes "{tenant}"
+func middlewareSetTemplatedUpstream(logger log.Logger, upstreamTemplate string) proxy.Middleware {
```
squat, line 176: I'm a bit worried about the future of this code. The hope is that eventually Jaeger will natively support tenancy and we will be able to stop using templated strings in the `--endpoint` flag but this code will still live here, providing a kind of secret functionality without maintenance. I wonder if this is a special piece of Jaeger-specific stop-gap business logic that should be handled somewhere else like (sigh) a special-purpose proxy that we can later throw away. Actually, you could probably even configure nginx to do this transformation with minimal config.

pavolloffay: This PR is a temporal workaround because Jaeger does not support multitenancy natively. The plan is to add multitenancy to Jaeger, we are already working on it. 

We could slightly change this PR to add a new flag that would allow tenant template in the endpoint. It will clearly allow us to enable/disable the feature and deprecate once multitenancy is in place.

squat: Yes, I'm aware that this is temporary and that's exactly what makes me so worried. I'm worried that this temporary code will secretly live here forever because it's a forgotten hack.

I agree that adding an explicitly UX-fiendish flag that is tied to this hack would help isolate the code and improve the likelihood of getting this it removed once the feature is available in Jaeger. An example could be `--experimental.traces.read.endpoint-template`

esnible: @squat I have moved the template stuff to an experimental flag as recommended by you and @pavolloffay .

Your point about using a proxy to keep this code out of Observatorium-api has merit.  I tried to start the discussion in https://github.com/observatorium/api/issues/268 .  Architecturally I agree with you.  I did it here because my feeling is that creating Jsonnet to configure a reverse proxy creates technical debt than the Golang to implement the template.

README.md
```diff
@@ -75,6 +75,8 @@ Usage of ./observatorium-api:
     	The percentage of mutex contention events that are reported in the mutex profile. (default 10)
   -debug.name string
     	A name to add as a prefix to log lines. (default "observatorium")
+  -experimental.traces.read.endpoint-template string
+    	A template replacing --read.traces.endpoint, such as http://jaeger-{tenant}-query:16686
```
pavolloffay, line 79: How does this flag relate to ` -traces.read.endpoint` what is the precedence if both are defined?

api/traces/v1/http.go
```diff
@@ -163,6 +165,46 @@ func NewV2Handler(read *url.URL, opts ...HandlerOption) http.Handler {
...
+	return func(r *http.Request) {
+		tenant, ok := authentication.GetTenant(r.Context())
+		if !ok {
+			// At this point another middleware must have put the tenant into the context
+			level.Debug(logger).Log("msg", "Internal error; expected tenant in request context")
+		}
+
+		upstream, ok := templateToURL[tenant]
+		if !ok {
+			upstream, err := ExpandTemplatedUpstream(readTemplate, tenant)
```
PhilipGough, line 193: What is stopping this being called with tenant being an empty string and does it matter?

api/traces/v1/http.go
```diff
@@ -163,6 +165,46 @@ func NewV2Handler(read *url.URL, opts ...HandlerOption) http.Handler {
...
+			if err != nil {
+				// Log if the tenant label includes characters that can't appear in a hostname (such as punctuation)
+				level.Debug(logger).Log("msg", "Internal error; tenant contains characters that cannot appear in hostname")
+			}
+
+			templateToURL[tenant] = upstream
+		}
+
+		r.URL.Scheme = upstream.Scheme
+		r.URL.Host = upstream.Host
```
PhilipGough, line 203: We continue to proceed from the earlier error check so won't this panic if the URL is nil?

api/traces/v1/http.go
esnible: - _main.go_ has `r.Use(authentication.WithTenantHeader(cfg.traces.tenantHeader, tenantIDs))` which prevents tenant from being empty
- It doesn't matter; suppose the template is `http://jaeger-{tenant}-query:16686`, it will fail with 502 BAD GATEWAY unless there is a host jaeger--query visible to DNS.

api/traces/v1/http.go
esnible: I missed a `:=` instead of `=` when refactoring.  Fixed; I have added an e2e test for the template parameter.

pavolloffay:
E2e test failed :/

esnible:
> E2e test failed :/

@pavolloffay I ignored the failure as it involved code I didn't write -- but it turned out to be my failure after all.  I was locally testing the old and new tests individually.  Both tracing tests used the same environment name.  When CircleCI ran both in parallel one failed.  Fixed.

api/traces/v1/http.go
bwplotka: Yea, I see both points, but to the Ed point I think it make sense to keep in in API if there is no need to reuse it in Jaeger. Let's make sure it is separate package, documented accordingly. Seems a bit better than jsonnet + another proxy yes.

README.md
jessicalins: I think would be nice to document the difference between the flags in the `Backends` section from the readme https://github.com/observatorium/api#backends

jsonnet/lib/observatorium-api.libsonnet
```diff
@@ -134,7 +134,10 @@ function(params) {
                   [
                     '--traces.write.endpoint=' + api.config.traces.writeEndpoint,
                     '--grpc.listen=0.0.0.0:' + api.config.ports['grpc-public'],
-                  ] else []
+                  ] + (if std.objectHas(api.config.traces, 'readEndpoint') then [
+                         '--traces.read.endpoint=' + api.config.traces.readEndpoint,
+                       ] else [])
+                else []
```
jessicalins, line 140: do we also need another if for `--experimental.traces.read.endpoint-template` ?

main.go
```diff
@@ -1028,6 +1034,20 @@ func parseFlags() (config, error) {
 		cfg.traces.writeEndpoint = rawTracesWriteEndpoint
 	}
 
+	if cfg.traces.readTemplateEndpoint != "" {
+		if cfg.traces.readEndpoint != nil {
```
jessicalins, line 1038: do we also need to check for `readTemplateEndpoint` [here](https://github.com/observatorium/api/blob/main/main.go#L1009-L1018) ? (to avoid having 2 flags defined as it is done here)

test/e2e/configs.go
```diff
@@ -22,6 +22,7 @@ const (
 	rules       testType = "rules"
 	logs        testType = "logs"
 	traces      testType = "traces"
+	tracetempls testType = "tracetempls"
```
jessicalins, line 25:
```suggestion
	tracesTemplate testType = "tracesTemplate"
```

I think something like this would be easier to read

test/e2e/configs.go
```diff
@@ -36,6 +37,7 @@ const (
 	envRulesAPIName = "e2e_rules_api"
 	envLogsName     = "e2e_logs_read_write_tail"
 	envTracesName   = "e2e_traces_read_export"
+	envTTemplsName  = "e2e_traces2_templ_query"
```
jessicalins, line 40:
```suggestion
	envTracesTemplateName  = "e2e_traces_template_query"
```

test/e2e/traces_test.go
```diff
@@ -257,3 +257,115 @@ func requestWithRetry(t *testing.T, testLabel string, client *http.Client, reque
 	testutil.Assert(t, false, fmt.Sprintf("%s: HTTP %d response not received within time limit", testLabel, expectedResponse))
 	return "", -1
 }
+
+func TestTraceTemplateQuery(t *testing.T) {
```
jessicalins, line 261:
```suggestion
func TestTracesTemplateQuery(t *testing.T) {
```

test/e2e/traces_test.go
```diff
@@ -257,3 +257,115 @@ func requestWithRetry(t *testing.T, testLabel string, client *http.Client, reque
...
+
+		// This test doesn't actually write logs, but we need
+		// this because Observatorium currently MUST see a logs or metrics endpoints
+		withLogsEndpoints("http://localhost:8080"),
+	)
+	testutil.Ok(t, err)
+	testutil.Ok(t, e2e.StartAndWaitReady(api))
+
+	t.Run("write-then-query-single-trace", func(t *testing.T) {
+
```
jessicalins, line 287: I think this extra line can be removed

test/e2e/traces_test.go
```diff
@@ -257,3 +257,115 @@ func requestWithRetry(t *testing.T, testLabel string, client *http.Client, reque
...
+		createOtelForwardingCollectorConfigYAML(t, e,
+			api.InternalEndpoint("grpc"),
+			token)
+
+		otel := e.Runnable("otel-fwd-coll").
```
jessicalins, line 292:
```suggestion
		otel := e.Runnable("otel-fwd-collector").
```

test/e2e/traces_test.go
jessicalins, line 261: I think would be nice to include unit tests for when defining the `read`/`readTemplate` flags (e.g. testing when both are defined, when the `readTemplate` flag has a wrong format, etc) - what do you think?

jessicalins (commented):
thanks @esnible, nice work :) I left some suggestions/questions

api/traces/v1/http.go
```diff
@@ -163,6 +165,47 @@ func NewV2Handler(read *url.URL, opts ...HandlerOption) http.Handler {
 	return r
 }
 
+// Parse a URL; if the URL includes `{tenant}` that portion will be replaced by the tenant
+func ExpandTemplatedUpstream(templateUpstream, tenant string) (*url.URL, error) {
+	rawTracesReadEndpoint := strings.Replace(templateUpstream, "{tenant}", tenant, 1)
```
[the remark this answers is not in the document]
esnible: Nothing; and this is by design, mostly for testing.

I thought about validating that the parameter `--experimental.traces.read.endpoint-template` includes `{tenant}`, as the user will not get the behavior without including it.

However, if I allow the template to be a constant, I can test the code without DNS manipulation.

So I would like to keep the behavior of not requiring a template.  Possibly we could add a warning message, if you think it would help the deployer.

api/traces/v1/http.go
```diff
@@ -163,6 +165,47 @@ func NewV2Handler(read *url.URL, opts ...HandlerOption) http.Handler {
...
+	return url.ParseRequestURI(rawTracesReadEndpoint)
+}
+
+// middlewareSetTemplatedUpstream is a variation of proxy.MiddlewareSetUpstream()
+// with additional processing if the upstream includes "{tenant}"
+func middlewareSetUpstream(logger log.Logger, read *url.URL, readTemplate string) proxy.Middleware {
+	if read != nil {
```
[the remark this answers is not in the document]
esnible: Yes.

From the point of view of Observatorium users this is not a problem; the flags are validated and Observatorium exists with "only one of --traces.read.endpoint and --experimental.traces.read.endpoint-template allowed" if the user supplied both.

If someone tried to import Observatorium-API as a library in another program it might be unclear that this function ignores the template if URL is supplied.  What do you think?

api/traces/v1/http.go
```diff
@@ -163,6 +165,47 @@ func NewV2Handler(read *url.URL, opts ...HandlerOption) http.Handler {
...
 
+// Parse a URL; if the URL includes `{tenant}` that portion will be replaced by the tenant
+func ExpandTemplatedUpstream(templateUpstream, tenant string) (*url.URL, error) {
+	rawTracesReadEndpoint := strings.Replace(templateUpstream, "{tenant}", tenant, 1)
+	return url.ParseRequestURI(rawTracesReadEndpoint)
+}
+
+// middlewareSetTemplatedUpstream is a variation of proxy.MiddlewareSetUpstream()
+// with additional processing if the upstream includes "{tenant}"
+func middlewareSetUpstream(logger log.Logger, read *url.URL, readTemplate string) proxy.Middleware {
```
bwplotka, line 176:
```suggestion
func middlewareSetTemplatedUpstream(logger log.Logger, read *url.URL, readTemplate string) proxy.Middleware {
```

api/traces/v1/http.go
```diff
@@ -163,6 +165,47 @@ func NewV2Handler(read *url.URL, opts ...HandlerOption) http.Handler {
...
 }
 
+// Parse a URL; if the URL includes `{tenant}` that portion will be replaced by the tenant
+func ExpandTemplatedUpstream(templateUpstream, tenant string) (*url.URL, error) {
+	rawTracesReadEndpoint := strings.Replace(templateUpstream, "{tenant}", tenant, 1)
+	return url.ParseRequestURI(rawTracesReadEndpoint)
+}
+
+// middlewareSetTemplatedUpstream is a variation of proxy.MiddlewareSetUpstream()
+// with additional processing if the upstream includes "{tenant}"
```
bwplotka, line 175:
```suggestion
// with additional processing if the upstream includes "{tenant}".
```

api/traces/v1/http.go
```diff
@@ -163,6 +165,47 @@ func NewV2Handler(read *url.URL, opts ...HandlerOption) http.Handler {
 	return r
 }
 
+// Parse a URL; if the URL includes `{tenant}` that portion will be replaced by the tenant
```
bwplotka, line 168:
```suggestion
// Parse a URL; if the URL includes `{tenant}` that portion will be replaced by the tenant.
```

api/traces/v1/http.go
```diff
@@ -163,6 +165,47 @@ func NewV2Handler(read *url.URL, opts ...HandlerOption) http.Handler {
...
+}
+
+// middlewareSetTemplatedUpstream is a variation of proxy.MiddlewareSetUpstream()
+// with additional processing if the upstream includes "{tenant}"
+func middlewareSetUpstream(logger log.Logger, read *url.URL, readTemplate string) proxy.Middleware {
+	if read != nil {
+		return proxy.MiddlewareSetUpstream(read)
+	}
+
+	// Cache upstream URLs to avoid re-parse on every read
```
bwplotka, line 181:
```suggestion
	// Cache upstream URLs to avoid re-parse on every read.
```

api/traces/v1/http.go
```diff
@@ -163,6 +165,47 @@ func NewV2Handler(read *url.URL, opts ...HandlerOption) http.Handler {
...
+	templateToURL := map[string]*url.URL{}
+
+	return func(r *http.Request) {
+		tenant, ok := authentication.GetTenant(r.Context())
+		if !ok {
+			// At this point another middleware must have put the tenant into the context
```
bwplotka, line 187:
```suggestion
			// At this point another middleware must have put the tenant into the context.
```

api/traces/v1/http.go
bwplotka: Nice

main.go
```diff
@@ -140,10 +140,13 @@ type logsConfig struct {
 }
 
 type tracesConfig struct {
+	// readTemplateEndpoint is of the form "http://jaeger-{tenant}-query:16686"
```
bwplotka, line 143:
```suggestion
	// readTemplateEndpoint is of the form "http://jaeger-{tenant}-query:16686".
```

test/e2e/services.go
```diff
@@ -313,6 +313,9 @@ type apiOptions struct {
 	tracesWriteEndpoint  string
 	gRPCListenEndpoint   string
 	jaegerQueryEndpoint  string
+
+	// "experimental.traces.read.endpoint-template" value
```
bwplotka, line 317:
```suggestion
	// "experimental.traces.read.endpoint-template" value.
```

test/e2e/traces_test.go
```diff
@@ -257,3 +257,115 @@ func requestWithRetry(t *testing.T, testLabel string, client *http.Client, reque
...
+	t.Cleanup(e.Close)
+
+	prepareConfigsAndCerts(t, tracetempls, e)
+	_, token, _ := startBaseServices(t, e, tracetempls)
+	internalOtlpEndpoint, httpExternalQueryEndpoint, httpInternalQueryEndpoint := startServicesForTraces(t, e)
+
+	api, err := newObservatoriumAPIService(
+		e,
+		withGRPCListenEndpoint(":8317"),
+		// Note that we don't include `{tenant}`, because we can't easily do this with DNS on Docker
```
bwplotka, line 275:
```suggestion
		// Note that we don't include `{tenant}`, because we can't easily do this with DNS on Docker.
```

test/e2e/traces_test.go
```diff
@@ -257,3 +257,115 @@ func requestWithRetry(t *testing.T, testLabel string, client *http.Client, reque
...
+		withOtelTraceEndpoint(internalOtlpEndpoint),
+		withExperimentalJaegerTemplateEndpoint("http://"+httpInternalQueryEndpoint),
+
+		// This test doesn't actually write logs, but we need
```
bwplotka, line 279: Missing trailing period at the end of all comments below too

bwplotka (dismissed):
Thanks Ed! This looks solid. Some nits only.

main.go
esnible: The initial version of the PR had only one flag.  The PR was simpler.  I broke the behavior into flags because of the suggestion of @squat https://github.com/observatorium/api/pull/271#discussion_r840353683 who wanted it to be easy to remove this behavior in the future.

test/e2e/traces_test.go
esnible: I like to test parameter validation in unit tests rather than e2e tests.  It feels like too much CircleCI effort to spin up a copy of the Observatorium pod and verify that it refuses to start when both flags are there.

There are currently no unit tests for `main()`.  If we need flag validation, let's do it in a separate PR.

api/traces/v1/http.go
esnible: I seem to make this mistake a lot.  Perhaps Observatorium could add a linter for it? https://github.com/tetafro/godot

api/traces/v1/http.go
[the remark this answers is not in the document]
jessicalins: Lets add a warning message then :+1:

api/traces/v1/http.go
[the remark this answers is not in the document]
jessicalins: Sorry, I'm still confused with the behavior of the 2 flags. If `-experimental.traces.read.endpoint-template` is provided but `--traces.read.endpoint` is provided too, `--traces.read.endpoint` will be used here? Wouldn't this be caught [here](https://github.com/observatorium/api/blob/099ad6d12754959be92647e620ebb62b55dc5d66/main.go#L1037-L1038) already?

main.go
jessicalins: ah, I mean if we also need to check for `readTemplateEndpoint` at https://github.com/observatorium/api/blob/main/main.go#L1009-L1018 and throw `only one of --traces.read.endpoint and --experimental.traces.read.endpoint-template allowed` in case `readTemplateEndpoint` it is defined too

jessicalins (commented):
just a couple of questions for my own understanding 
unfortunately e2e tests are failing again :(

test/e2e/traces_test.go
jessicalins: yeah, I meant to add an unit test for that. I think since the flag validation is not happening, then I agree to add it in case flag validation is needed in the future

esnible:
@jessicalins I found a race.  The flaky trace test should be fixed.

_main.go_ 's `parseFlags()` will fail and Observatorium will not start if both flags are defined.

_api/traces/v1/http.go_ 's `middlewareSetTemplatedUpstream()` was doing double-duty.  I have refactored to remove that confusing behavior.

main.go
```diff
@@ -1039,6 +1039,12 @@ func parseFlags() (config, error) {
 			return cfg, fmt.Errorf("only one of --traces.read.endpoint and --experimental.traces.read.endpoint-template allowed")
 		}
 
+		if !strings.Contains(cfg.traces.readTemplateEndpoint, "{tenant}") {
+			fmt.Fprintf(os.Stderr,
```
jessicalins, line 1043: I would change this to `level.Warn(logger).Log(...)`

test/e2e/traces_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build integration
```
jessicalins, line 1: the tags again =/

main.go
jessicalins: ping @esnible

test/e2e/traces_test.go
esnible: I will fix.  **It's just a comment**.  Sure, it is a comment the Go team wants us to use, but it is only a comment.

main.go
esnible: The problem here is that the problem is detected in parseFlags(), but until parseFlags() is completed we can't create the logger because don't know the level.

I can probably move the check outside of parseFlags completely if you think that will help.

jessicalins: ah you are right. I think then makes sense to leave inside `parseFlags(..)` :+1:

jessicalins (approved):
LGTM, thanks @esnible! :rocket:

Issue #266, opened 2022-03-28
OIDC logout handler


esnible:
There is no way to log out of Observatorium.

I propose a logout handler `/oidc/{tenant}/logout` to be added near https://github.com/observatorium/api/blob/main/authentication/oidc.go#L160 

This handler would set the tenant OIDC cookie to "" with expiration at 1-1-1970, and then redirect to some Observatorium path, triggering the OIDC login flow to let the user log in as someone else.

esnible:
Notes from April 6 2021 Observatorium Community Meeting

- Separation of UI and tenancy metadata (UI does not need to compute path)
- /login /logout
  - Each tenant has a different cookie
  - Putting tenants in the path will be then easier
- Prem proposal: Easier for Jaeger UI - relative UI /logout/{tenant} that proxies to proper endpoint in UI.
  - Better namespaces to not have top-level matchers
  - /login/{tenant}
- Every tenant has a specific URL

esnible:
Notes from June 1 Observatorium Community Meeting

I am experiencing this problem again.  I have installed Observatorium and am using Keycloak OIDC, but misconfigured tenants.yaml or rbac.yaml.

The symptom is that Keycloak believes I am logged in, but the login user has no permissions to see Observatorium resources.  / returns valid JSON, the real pages correctly return HTTP Unauthorized.

The problem is that /oidc/<tenant>/login redirects to Keycloak, which believes I am logged in, and redirects back without an opportunity to log in as a different user.  The only way to change users is by deleting Keycloak cookies on the browser.

PR #274, opened 2022-04-06
Login endpoint at /api/traces/v1/{tenent}/login

esnible:
Resolves https://github.com/observatorium/api/issues/266

This creates the endpoint /api/traces/v1/{tenent}/login which redirects to that tenant's provider-specific login endpoint.

This allows the trace UI to log out using a relative path, e.g. `./logout`, rather than requiring knowledge of the authentication provider specified in _tenants.yaml_.

Currently only exposed for tracing, but all the plumbing is there other UIs.

PR is created as a **draft** so please comment on the direction/idea.  If the approach is good I will begin work on tests.

### DIFF
--- a/authentication/oidc.go
+++ b/authentication/oidc.go
@@ -434,3 +434,7 @@ func (a oidcAuthenticator) checkAuth(ctx context.Context, token string) (context
 
 	return ctx, "", http.StatusOK, codes.OK
 }
+
+func (a oidcAuthenticator) LoginPath(tenant string) string {
+	return strings.ReplaceAll("/oidc/{tenant}/login", "{tenant}", tenant)
+}
--- a/authentication/authentication.go
+++ b/authentication/authentication.go
@@ -42,6 +42,7 @@ type Provider interface {
 	Middleware() Middleware
 	GRPCMiddleware() grpc.StreamServerInterceptor
 	Handler() (string, http.Handler)
+	LoginPath(tenant string) string
 }
 
 type tenantHandlers map[string]http.Handler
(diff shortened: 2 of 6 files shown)


authentication/oidc.go
```diff
@@ -434,3 +434,7 @@ func (a oidcAuthenticator) checkAuth(ctx context.Context, token string) (context
 
 	return ctx, "", http.StatusOK, codes.OK
 }
+
+func (a oidcAuthenticator) LoginPath(tenant string) string {
+	return strings.ReplaceAll("/oidc/{tenant}/login", "{tenant}", tenant)
```
saswatamcode, line 439: Maybe `path.Join("/oidc", tenant, "login")` would be better here?

saswatamcode, line 439: Maybe `path.Join("/oidc", tenant, "login")` would be better here?

esnible: I am happy to make this change.

We currently have `handlerPrefix = "/oidc/{tenant}"` and `return "/oidc/{tenant}", a.handler`.  My thought was that it was nice to keep a similar appearance in related code.  Eventually those two fragments and my new piece should be merged.

Similar code exists in _authentication/openshift.go_.  Let me know what you think and I will make it happen.

frzifus:
Hi, whats the status here?

matej-g (commented):
Hey @esnible is this still draft? I think the solution looks good :+1:

esnible:
@matej-g I marked this ready for review six days ago.

authentication/oidc.go
matej-g: I'd still say `path.Join` is more common in our code base as well as easier on the eyes, it's a minor point though.

matej-g (approved):
This is looking good to me, thanks @esnible :+1:

squat (changes requested):
I'd like to see any changes / additions to the API accompanied by a proposal with concrete plans about how to standardize this behavior and a discussion of alternatives. Otherwise I'm afraid we'll perpetually lock ourselves into one-off behavior.

PR #284, opened 2022-04-25
Update LogQL parser language, tests

sasagarw:
This PR:
- updates parser language to correctly parse `unwrap` and `offset` expr.
- add tests for `unwrap` expr.
- add tests for `__line__` function.
- add tests for `offset` expr.

/cc @periklis


periklis:
@sasagarw Can you please doublecheck that we support also the `offsetExpr` by adding a test if non available yet?

periklis (commented):
One last gate check before approving this. Do we have sufficient parser test coverage for all the places in LogQL where `unwrap` and `offset` can be put? Did you double check that we sufficiently cover the same test cases as in [upstream parser_test.go](https://github.com/grafana/loki/blob/main/pkg/logql/syntax/parser_test.go) for `unwrap` and `offset`?

sasagarw:
It's able to parse all the expressions now @periklis 
Added tests for trivial ones.

periklis (approved):
LGTM

PR #292, opened 2022-05-17
Support reading access token from header

xperimental:
Grafana can be paired with [oauth-proxy](https://github.com/openshift/oauth-proxy) in an OpenShift cluster for proper authentication with the cluster authentication service. While the proxy can be configured to forward the access token, it uses a custom header (`X-Forwarded-Access-Token`) which is currently not evaluated by observatorium-api.

This PR adds support for the custom header by picking it up in a middleware and passing it on to the authentication/authorization part if the `Authorization` header is empty.

The authentication code of this service uses code from `k8s.io/apiserver`. As I have not found a way to customize the bahaviour of picking up the access token from the `Authorization` header, the current implementation in this PR will add a new `Authorization` header with the provided token for the authentication code to use.

### DIFF
--- a/authentication/http.go
+++ b/authentication/http.go
@@ -2,6 +2,7 @@ package authentication
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


authentication/http.go
```diff
@@ -59,7 +60,17 @@ func WithAccessToken() Middleware {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			rawToken := r.Header.Get("Authorization")
-			token := rawToken[strings.LastIndex(rawToken, " ")+1:]
+			token := strings.TrimPrefix(rawToken, "Bearer ")
```
xperimental, line 63: This is a change to the behavior in the code. It seemed safer than the original code, if `Bearer token` is the only variant of `Authorization` header that is supported.

The new code is more strict in that it only removes a pre-defined prefix. The old version would accept a header in the form `Bearer firsttoken secondtoken` and read `secondtoken` from the header, which might lead to exploits if another component only evaluates the `firsttoken`.

PR #293, opened 2022-05-19
Stop OIDC flow when remote writing with bad auth header

douglascamata:
When handling remote write requests, if the Authorization
header isn't present, we don't want to redirect users back
to the identity provider for login.

The reason behind this is that tenants can overload the
identity provider in case they have bad configuration.
We want to be friendly and help prevent this issue from
happening.


main.go
```diff
@@ -520,6 +520,10 @@ func main() {
 
 			r.Use(authentication.WithTenant)
 			r.Use(authentication.WithTenantID(tenantIDs))
+			// We want to enforce that all remote write requests have an access token present in the Authorization
+			// header to avoid repeatedly redirecting clients that might be unaware of OIDC flow to the identity
+			// provider, possibly overloading it.
+			r.Use(authentication.EnforceAccessTokenPresent("/api/v1/receive"))
```
matej-g, line 526: I'm wondering if we could make this somehow more universal, since right now this will only apply to metrics / remote write, but similar situation could occur with logs and traces as well right?

matej-g (commented):
Thanks for the initiative :heart: This approach is looking great to me, I just have few minor points. You would also get extra coolness points if we had a test for this :sunglasses:.

authentication/http.go
```diff
@@ -54,6 +54,30 @@ func WithTenantID(tenantIDs map[string]string) Middleware {
...
+
+			rawToken := r.Header.Get("Authorization")
+			if rawToken == "" {
+				http.Error(w, "couldn't find the authorization header", http.StatusBadRequest)
+				return
+			}
+			if !strings.Contains(rawToken, "Bearer ") {
+				http.Error(w, "malformed authorization header", http.StatusBadRequest)
+				return
+			}
```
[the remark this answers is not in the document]
PhilipGough: I tend to agree with @matej-g here, but just out of interest, does the presence of an `Authorization` header alone avoid the server sending the client into this redirect hotloop or does the situation also occur wit an invalid/expired token?

main.go
douglascamata: @matej-g I don't know much how writing logs/traces work, but I suspect they are susceptible to the same kind of situation. I'm happy to use this approach for these other telemetry signals in follow up PRs. :)

authentication/http.go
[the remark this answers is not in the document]
douglascamata: We could leave it out, at the cost of decreasing protection from tenants "DoSing" the identity provider with bad configuration. 

The idea of doing this very simple check here is exactly to hard-stop a request **before** it reaches the OIDC middleware, because this kind of failure there means a redirect to the identity provider's login page.

What do you think?

main.go
PhilipGough: We know the write endpoints upfront right? So perhaps renaming the function to something like `EnsureAccessTokenPresentOnWrite()` or similar and have a check against all endpoints?

authentication/http.go
```diff
@@ -54,6 +54,30 @@ func WithTenantID(tenantIDs map[string]string) Middleware {
...
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if !strings.HasSuffix(r.URL.Path, urlPath) {
+				next.ServeHTTP(w, r)
+				return
+			}
+
+			rawToken := r.Header.Get("Authorization")
+			if rawToken == "" {
+				http.Error(w, "couldn't find the authorization header", http.StatusBadRequest)
```
PhilipGough, line 69: Is there potentially any other flows we could break here? cookies etc?

authentication/http.go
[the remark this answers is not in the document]
PhilipGough: what about situations other than OIDC, I see support for mTLS and OpenShift authenticator for example? 🤔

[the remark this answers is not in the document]
douglascamata: I don't think Prometheus remote write supports OpenShift authenticator, so this discards the OpenShift authenticator at least, correct?

I am investigating how can we make this not break the mTLS authentication. Sadly there aren't tests for this. :(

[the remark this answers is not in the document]
matej-g: > what about situations other than OIDC, I see support for mTLS and OpenShift authenticator for example? thinking

That's a good point actually, should this be then part of the OIDC middleware rather, as it applies only to tenants with OIDC authentication providers?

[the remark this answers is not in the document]
douglascamata: Thanks for the comments too, @PhilipGough and @esnible. I am taking your suggestions into account. 👍

douglascamata:
Summarizing the recent changes after applying review's feedback:

* Check on content of the `Authorization` header removed.
* No redirect to login in both Prometheus remote write **and** Loki log push endpoints. Traces already don't have the login redirect (it runs on the gRPC middleware stack).
* The core logic is separate from the OIDC authenticator middleware to avoid injecting knowledge about what's behind authentication into the authenticator.

douglascamata:
@matej-g @esnible @PhilipGough looks like the Github feature for re-requesting a PR view is a bit messed up. This is ready for another look from you.

douglascamata:
By the way, I can give a try to adding a test for this in another PR. I am afraid it will take too long and add too much modification to a single PR though, @matej-g

authentication/oidc.go
```diff
@@ -434,3 +434,46 @@ func (a oidcAuthenticator) checkAuth(ctx context.Context, token string) (context
...
 }
+
+// EnforceAccessTokenPresentOnSignalWrite enforces that the Authorization header is present in the incoming request
+// for the given list of tenants. Otherwise, it returns an error.
+// It protects the Prometheus remote write and Loki push endpoints. The tracing endpoint is not protected because
+// it goes through the gRPC middleware stack, which behaves differently from the HTTP one.
+func EnforceAccessTokenPresentOnSignalWrite(oidcTenants map[string]struct{}) Middleware {
+	protectedSignalWritePath := []string{
+		// Prometheus remote write endpoint
+		"/api/v1/receive",
```
squat, line 445: This looks to me like we are crossing abstraction boundaries. I'm worried about hardcoding API endpoints for data into authentication middleware. The auth middleware should really not know anything about this. Imagine we one day change an endpoint or add a new one, now we have to somehow remember to update this list as well. I would worry this is asking for vulnerability to be introduced.

Maybe we could get around this by having a middleware factory that accepts a list of endpoints and we can produce this middleware in the main.go.

douglascamata: > This looks to me like we are crossing abstraction boundaries.
I had a similar feeling about it, @squat. I tried to think of a way to prepend (not append, because we need this to run before the auth middleware, as it will do the redirect) this middleware to the stack of each sub-router (for metrics, logs, and traces... in their respective blocks in `main.go`). I didn't find a good way of doing this though. I'll investigate this more.

I was even wondering whether we could "cut the problem at its root" and remove the login redirect in the OIDC authentication flow and let it a responsibility of the clients. What do you think?

jessicalins:
thanks for adding this!

> By the way, I can give a try to adding a test for this in another PR. I am afraid it will take too long and add too much modification to a single PR though, @matej-g
^ I'd vote for adding a test on this PR too

douglascamata:
To be able to write tests for this using the e2e framework, we need https://github.com/observatorium/up/pull/44 to be merged.

authentication/http.go
```diff
@@ -158,3 +158,42 @@ func WithTenantMiddlewares(mwFns ...MiddlewareFunc) Middleware {
...
+			if _, found := oidcTenants[tenant]; !found {
+				next.ServeHTTP(w, r)
+				return
+			}
+
+			// We aren't interested in blocking requests from tenants not using OIDC.
+			if _, found := oidcTenants[tenant]; !found {
+				next.ServeHTTP(w, r)
+				return
+			}
```
matej-g, line 187: Looks like this is an erroneous duplicate with lines 178-181?

matej-g (commented):
Great job @douglascamata and thanks for adding tests! I'm totally happy with current state, minus removing the duplicate code lines. After that all :green_circle:.

authentication/http.go
douglascamata: Totally! It must have passed through some merge conflict or something. Fixed in a3de8f4. Thanks!

matej-g (approved):

matej-g (dismissed):

main.go
```diff
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"github.com/metalmatze/signal/internalserver"
```
PhilipGough, line 11: tiny nit: This is an unnecessary diff

PhilipGough (approved):
looks good @douglascamata great job!!

api/metrics/legacy/http.go
```diff
@@ -71,6 +71,14 @@ func WithUIMiddleware(m func(http.Handler) http.Handler) HandlerOption {
 	}
 }
 
+// WithGlobalMiddleware adds a middleware for all operations.
+func WithGlobalMiddleware(m ...func(http.Handler) http.Handler) HandlerOption {
+	return func(h *handlerConfiguration) {
+		h.uiMiddlewares = append(h.uiMiddlewares, m...)
+		h.queryMiddlewares = append(h.queryMiddlewares, m...)
+	}
```
PhilipGough, line 79: do we need to update the legacy path?

douglascamata: The legacy path doesn't seem to accept remote writes, so we don't need to update it. Serving just UI and query it's "safe".

main.go
douglascamata: Auto import 😅

douglascamata: Fixed in [4c521e0](https://github.com/observatorium/api/pull/293/commits/4c521e0d51cbae0f67e1ed6343414bc359a142d3)

douglascamata:
@PhilipGough: can you please re-review the PR? I fixed your last comment. :)

PhilipGough (approved):

PR #295, opened 2022-05-27
Add per tenant HTTP metrics

douglascamata:
This pull request adds the tenant ID to the HTTP metrics, so that HTTP traffic metrics can be drilled down to each tenant. 

This should help identifying issues happening to specific tenants and also allow dashboards to be built to track tenants' behavior.


squat (approved):

matej-g (approved):
*chef's kiss*

PR #299, opened 2022-05-30
Makefile: Fix clashing `go mod tidy` and `go mod download all`

matej-g:

Currently, doing `go mod tidy` and `go mod download all` is giving a different resulting `go.sum` file, causing a clash, which shows up e.g. in Dependabot PRs, which are failing due to diff in `go.sum`.

We added the `all` argument to `go mod download` in our `make build` target with move to Go 1.17 and I can trace this change in behavior to that Go release (see release notes: https://go.dev/doc/go1.17#go-command, and related change https://go-review.googlesource.com/c/go/+/318629). AFAIU, for our build command, we should be fine with having just `go mod download` to download dependency for the main module.


douglascamata:
@matej-g on a similar note, WDYT if we remove the `-a` (forces rebuilding of packages) option we use in the `go build` command to compile the binary in the Makefile? Without it the compilation is much faster, as the cache is used when applicable. I don't know historic context on why we need that `-a` flag there.

douglascamata (approved):

matej-g:
> @matej-g on a similar note, WDYT if we remove the `-a` (forces rebuilding of packages) option we use in the `go build` command to compile the binary in the Makefile? Without it the compilation is much faster, as the cache is used when applicable. I don't know historic context on why we need that `-a` flag there.

I'm not sure as well whether there are any historical reasons for that, as it predates me, but the suggestion makes sense.

PR #314, opened 2022-06-13
Add logo within spec for API docs

saswatamcode:
This PR adds a logo in the spec, which would appear in API docs here: https://github.com/observatorium/observatorium/pull/483.


matej-g (approved):
image

PR #316, opened 2022-06-13
Supress "TLS handshake error.. connection reset by peer" logs 

aminesnow:
**Context:** When enabling TLS with Route reencrypt, the gateway gets frequently healthcheck pinged (every 5s). These pings don't finished the TLS handshake and thus cause the container to log a TLS handshake error.

This PR adds a custom error logger to the gateway server in order to suppress the flood of TLS handshake error logs.


squat:
This is from healthchecking ourselves, right? Not from another system doing the healthchecking, e.g. k8s?

aminesnow:
> This is from healthchecking ourselves, right? Not from another system doing the healthchecking, e.g. k8s?

No, this from router pods health check probes.

aminesnow:
> > No, this from router pods health check probes.
> 
> Ah, in that case I think we can unfortunately not make this proposed change. What if Observatiorium is getting _real_ TLS Handshake failures from a client? We'll be erroneously suppressing the errors logs.

Understood. Is it possible to make it a flag on/off? (supress-tls-errors for ex)

PR #318, opened 2022-06-17
Do not continue with a request if label enforcement fails

matej-g:

I have noticed strange behavior in responses to the API, when the label enforcement fails. The request got still propagated through the middleware chain all the way to the metrics backend. This was causing a strange double error responses (mix of JSON / plain text), e.g.:
```
could not enforce labels: parse expr error: 1:13: parse error: unexpected end of input inside braces
{"status":"error","errorType":"bad_data","error":"1:13: parse error: unexpected end of input inside braces"}
```

After some investigation, I realized we are actually continuing in the middleware chain even if the label enforcement fails, but this should not be the case. The provided fix ensures we return from the middleware chain once the enforcement fails.


squat (approved):
Nice catch

Issue #319, opened 2022-06-17
Consider adopting the Prometheus JSON format for error responses


matej-g:
Referring to: https://prometheus.io/docs/prometheus/latest/querying/api/#format-overview

We chatted about this with @saswatamcode (off the back of https://github.com/observatorium/api/pull/318) and we are finding the current approach with mixed JSON (which is returned in case of `2xx` responses) and plain-text (returned for `4xx` and `5xx`) responses not satisfactory.

Ideally, we should enhance the API spec to specify that the error responses should be also JSON and adapt the API to return the Prometheus JSON format for errors. This would have added benefit of using the well-known Prometheus format, which would make user experience nicer, especially in the downstream web UI. For example, right now Thanos UI will fail to show an error response, since it expects JSON and cannot parse a plain-text response.

squat:
Nice, we recently did the same in prom label proxy: https://github.com/prometheus-community/prom-label-proxy/pull/108

saswatamcode:
Closing as this has been implemented! 🙂

PR #326, opened 2022-06-22
Return errors in Prometheus HTTP API format

saswatamcode:
This PR addresses #319 and ensures that all `4xx`/`5xx` responses from Observatorium API are sent in Prometheus API JSON format instead of plaintext, by taking inspiration from https://github.com/prometheus-community/prom-label-proxy/pull/108. This makes debugging via downstream UI/clients much easier.

Thus for a faulty query,

Response before, 
```
could not enforce labels: parse expr error: 1:12: parse error: unterminated quoted string
```
Response after,
```json
{"error":"could not enforce labels: parse expr error: 1:12: parse error: unterminated quoted string","errorType":"observatorium-api","status":"error"}
```

Loki is also [compatible](https://grafana.com/docs/loki/latest/api/) with Prometheus API result format, so we can use the same `PrometheusAPIError` method in `api/logs/v1/labels_enforcer.go` too. :)

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -1,0 +1,19 @@
+package utils
+
+import (
+	"encoding/json"
+	"log"
+	"net/http"
+)
+
+// Adapted from https://github.com/prometheus-community/prom-label-proxy/blob/02d43edb82b7d139f4a4e41912ad903bff46d5c4/injectproxy/utils.go#L22
+func PrometheusAPIError(w http.ResponseWriter, errorMessage string, code int) {
+	w.Header().Set("Content-Type", "application/json; charset=utf-8")
+	w.WriteHeader(code)
+
+	res := map[string]string{"status": "error", "errorType": "observatorium-api", "error": errorMessage}
+
+	if err := json.NewEncoder(w).Encode(res); err != nil {
+		log.Printf("failed to encode json: %v", err)
+	}
+}
--- a/api/logs/v1/labels_enforcer.go
+++ b/api/logs/v1/labels_enforcer.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/observatorium/api/authorization"
 	logqlv2 "github.com/observatorium/api/logql/v2"
+	"github.com/observatorium/api/utils"
 	"github.com/prometheus/prometheus/pkg/labels"
 )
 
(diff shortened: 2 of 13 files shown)


utils/utils.go
```diff
@@ -0,0 +1,19 @@
+package utils
```
matej-g, line 1: What do you think about naming it `error` or something similar? If we have only error handling here, I prefer to be more specific with package names.

utils/utils.go
```diff
@@ -0,0 +1,19 @@
...
+	"log"
+	"net/http"
+)
+
+// Adapted from https://github.com/prometheus-community/prom-label-proxy/blob/02d43edb82b7d139f4a4e41912ad903bff46d5c4/injectproxy/utils.go#L22
+func PrometheusAPIError(w http.ResponseWriter, errorMessage string, code int) {
+	w.Header().Set("Content-Type", "application/json; charset=utf-8")
+	w.WriteHeader(code)
+
+	res := map[string]string{"status": "error", "errorType": "observatorium-api", "error": errorMessage}
```
matej-g, line 14: I wonder if we should establish in the API spec what error types there can be, whether there should always be just one etc., but either way we should have this documented in the spec. For starters, this is fine for me though :+1:.

utils/utils.go
```diff
@@ -0,0 +1,19 @@
...
+
+	if err := json.NewEncoder(w).Encode(res); err != nil {
```
matej-g, line 16: Optional: My personal preference only - encode into JSON and then use `http.Error` to set code, headers etc.; that way we can 'outsource' it to the `http` package and not do it ourselves by hand

matej-g (commented):
This is looking nice! :clap: I have few small suggestions. 

It would also be cool to update the API spec and think about what `error_type`s we could have (or if everything should be single error type) and update accordingly.

utils/utils.go
bwplotka: Yes - the `utils` package is an anti-pattern, we, Go developers should never create. 

It's like a mantra =D 

NO `UTILS`. Maybe `common` (🙃 sometimes when drunk). 

Otherwise, let's aim always for some single ~functionality purpose package like `httperr` in this case. Otherwise,  it can grow into some nice meaningless garbage package.

bwplotka (commented):
Nice!

utils/utils.go
saswatamcode: Discussed with @matej-g, so using http.Error isn't suitable here as it always sends a [plaintext response](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/net/http/server.go;l=2093). So resolving this! 🙂

squat (approved):
great work @saswatamcode

PR #327, opened 2022-06-22
authentication: clean up naming

squat:
This PR includes two commits aimed at cleaning up the naming in the authentication/authentication.go file:
- authentication: clarify naming
This commit makes the naming within the `InitializeProvider` func follow go conventions of naming to clarify the intent of the named identifier.
- authentication: use consistent naming
This commit tries to standardize the naming of the `authenticator` variable and renames it internally within the `authentication` package as a `provider`.

### DIFF
--- a/authentication/authentication.go
+++ b/authentication/authentication.go
@@ -69,38 +69,38 @@ func NewProviderManager(l log.Logger, registrationRetryCount *prometheus.Counter
 
 // InitializeProvider initializes an authenticator provider and register the created
 // authentication middleware and handler.
-func (ah *ProviderManager) InitializeProvider(config map[string]interface{},
+func (pm *ProviderManager) InitializeProvider(config map[string]interface{},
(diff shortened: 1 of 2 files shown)


authentication/authentication.go
```diff
@@ -74,24 +74,24 @@ func (ah *ProviderManager) InitializeProvider(config map[string]interface{},
...
 			level.Error(ah.logger).Log("msg", err, "tenant", tenant, "authenticator", authenticatorType)
 			authCh <- nil
 			return
 		}
 
 		ah.mtx.Lock()
-		ah.middlewares[tenant] = authenticator.Middleware()
-		ah.gRPCInterceptors[tenant] = authenticator.GRPCMiddleware()
-		pattern, handler := authenticator.Handler()
+		ah.middlewares[tenant] = provider.Middleware()
```
jessicalins, line 92: I think we could also rename `ah` -> `p` or `pm` - what do you think?

jessicalins (commented):
much better :rocket:

jessicalins (approved):
:rocket:

PR #335, opened 2022-07-06
[spike] Add native routing for logs routes

periklis:
### Background
The observatorium-api is currently used as the default reverse proxy for RHOBS as well as by the Loki-Operator for the LokiStack resources. Currently it supports a path-based prefix-only model to proxy API path `/logs/api/v1/{tenant}/loki/api/v1/...` to `loki/api/v1/...`. This model serves well enough for environments where we control the client configuration or at least have impact on the controller side.

### Problem statement
With the re-use of the observatorium-api in the Loki-Operator as the default reverse-proxy for the LokiStack CR, we started to get user reports on why they can't configure their Grafana or Log collector to simply read/push logs via the official Loki API, i.e. `/loki/api/v1/...` ([Official Loki API](https://grafana.com/docs/loki/latest/api/)). This repeatedly creates a demand to create bespoke documentation for each client on how to connect via observatorium api (See [Loki Operator Forwarding Logs Docs](https://loki-operator.dev/docs/forwarding_logs_to_gateway.md/)). Since we don't control all clients and will never be, the documentation effort becomes unmaintainable and user experience remains bad.

### Proposed solution
The following PR adds support for exposing natively the Loki routes w/o sacrificing any functionality of the observatorium api (authn/authz/rate-limiter). This requires to opt-in via a CLI flag and in turn keep the old behavior (not exposing the native routes) as the prevailing default.

WDYT?

(cc @xperimental @alanconway)

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,8 @@ Usage of ./observatorium-api:
     	The log format to use. Options: 'logfmt', 'json'. (default "logfmt")
   -log.level string
     	The log filtering level. Options: 'error', 'warn', 'info', 'debug'. (default "info")
+  -logs.native.routes
+    	Enable native /loki/api/v1 routes for opaque proxying to Loki. Requires passing X-Scope-OrgID header.
   -logs.read.endpoint string
     	The endpoint against which to make read requests for logs.
   -logs.tail.endpoint string
--- a/main.go
+++ b/main.go
@@ -901,7 +901,7 @@ func parseFlags() (config, error) {
 	flag.StringVar(&cfg.server.healthcheckURL, "web.healthchecks.url", "http://localhost:8080",
 		"The URL against which to run healthchecks.")
 	flag.BoolVar(&cfg.logs.enableNativeRoutes, "logs.native.routes", false,
-		"Enable native `/loki/api/v1` routes for opaque proxying to Loki. Requires passing `X-Scope-OrgID` header.")
+		"Enable native /loki/api/v1 routes for opaque proxying to Loki. Requires passing X-Scope-OrgID header.")
 	flag.StringVar(&rawLogsTailEndpoint, "logs.tail.endpoint", "",
 		"The endpoint against which to make tail read requests for logs.")
 	flag.StringVar(&rawLogsReadEndpoint, "logs.read.endpoint", "",


main.go
```diff
@@ -894,6 +900,8 @@ func parseFlags() (config, error) {
 		"The address on which the internal server listens.")
 	flag.StringVar(&cfg.server.healthcheckURL, "web.healthchecks.url", "http://localhost:8080",
 		"The URL against which to run healthchecks.")
+	flag.BoolVar(&cfg.logs.enableNativeRoutes, "logs.native.routes", false,
+		"Enable native /loki/api/v1 routes for opaque proxying to Loki. Requires passing X-Scope-OrgID header.")
```
bwplotka, line 904:
```suggestion
		"Enable '/loki/api/v1'  prefix instead of '/logs/api/v1/{tenant}/loki/api/v1` prefix for routes to allow opaque proxying to Loki. Requires passing X-Scope-OrgID header.")
```

bwplotka (changes requested):
Makes sense to me, LGTM 👍🏽 
Small nit only.

bwplotka (approved):
Misclicked above. Approved

authentication/http.go
```diff
@@ -42,6 +42,23 @@ func WithTenant(next http.Handler) http.Handler {
 	})
 }
 
+// WithTenantFromHeader finds the tenant from the HTTP headers and adds it to the request context.
+func WithTenantFromHeader(tenantHeader string) Middleware {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			tenant := r.Header.Get(tenantHeader)
+			if tenant == "" {
+				httperr.PrometheusAPIError(w, "tenant not found", http.StatusInternalServerError)
```
matej-g, line 51: I think we can make this `400` (if user does not provide the header, it's bad request)

matej-g (commented):
The solution makes sense to me :+1:, one small suggestion

README.md
```diff
@@ -91,6 +91,8 @@ Usage of ./observatorium-api:
     	The log format to use. Options: 'logfmt', 'json'. (default "logfmt")
   -log.level string
     	The log filtering level. Options: 'error', 'warn', 'info', 'debug'. (default "info")
+  -logs.native.routes
```
xperimental, line 94: Shouldn't it be `logs.native-routes`?

PR #336, opened 2022-07-06
Add flags for setting maximum TLS version and client authentication

xperimental:
We currently encounter an issue where the health-check of HAProxy causes a periodic error message to be printed by this service:

```
2022/07/06 16:26:52 http: TLS handshake error from haproxy-ip:43002: write tcp service-ip:8080->haproxy-ip:43002: write: connection reset by peer
```

This is because observability-api tries to request a client-certificate from all connections by default, which fails in this case because HAProxy does not do any TLS exchange, but immediately closes the connection again.

I have also added the `max-version` setting because it was helpful during debugging this case and looked useful for the future.

This PR replaces #316 .

Refs: https://issues.redhat.com/browse/LOG-2750

### DIFF
--- a/tls/config.go
+++ b/tls/config.go
@@ -10,7 +10,7 @@ import (
 )
 
 // NewServerConfig provides new server TLS configuration.
-func NewServerConfig(logger log.Logger, certFile, keyFile, minVersion string, cipherSuites []string) (*tls.Config, error) {
+func NewServerConfig(logger log.Logger, certFile, keyFile, minVersion, maxVersion, clientAuthType string, cipherSuites []string) (*tls.Config, error) {
 	if certFile == "" && keyFile == "" {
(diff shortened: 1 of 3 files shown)


periklis (approved):

tls/config.go
```diff
@@ -24,25 +24,57 @@ func NewServerConfig(logger log.Logger, certFile, keyFile, minVersion string, ci
...
+	}
+
 	cipherSuiteIDs, err := flag.TLSCipherSuites(cipherSuites)
 	if err != nil {
 		return nil, fmt.Errorf("TLS cipher suite name to ID conversion: %v", err)
 	}
 
+	tlsClientAuthType, err := parseClientAuthType(clientAuthType)
+	if err != nil {
+		return nil, fmt.Errorf("can not parse TLS Client authentication policy: %w", err)
```
squat, line 48: nit:
s/can not/cannot/

PR #342, opened 2022-07-14
Running as non-root in Kubernetes

Red-GV:
This PR defines the USER in the Dockerfile as 10000. When running the image as nonroot in Kubernetes, the resulting pod(s) will have a CreateContainerConfigError. This is because the USER in the image is not defined and assumed to be root.

The work around for this in Kubernetes is to define the user to be some non-zero user. However, as the image does not require any elevated privileges, it would be nice to make it work with the non-root user specification. This is relevant to the Kubernetes Pod Security Admission changes, which require a number of changes to be compliant with the `restricted` policy.

ref: https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,8 @@ ARG VERSION
 ARG VCS_REF
 ARG DOCKERFILE_PATH
 
+USER 10000:10000
+
 LABEL vendor="Observatorium" \
     name="observatorium/api" \
     description="Observatorium API" \


squat (dismissed):
This sounds good to me. We should ideally standardize this across all of our container images, as none of them require root as well as out observatorium/observatorium conffiguration repo so that this change is coupled to a corresponding configuration change.

Red-GV:
Just as a note, it doesn't matter what number the user is as long as it isn't 0. I don't know if there is some specific value that would like to be used instead.

Dockerfile
```diff
@@ -17,6 +17,8 @@ ARG VERSION
 ARG VCS_REF
 ARG DOCKERFILE_PATH
 
+USER 10000
```
squat, line 20: Let's make sure that the group is also non-zero:
```suggestion
USER 10000:10000
```

periklis (approved):

PR #349, opened 2022-07-21
Update golang to v1.18

Red-GV:
This PR updates the project to go1.18

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/observatorium/api
 
-go 1.17
```
squat, line 3: This should not be needed as this declares the minimum version required by the project, however we don't use any go 1.18-specific features.

periklis (approved):

PR #365, opened 2022-08-23
Remove usage of io/ioutil

saswatamcode:
This PR removes the usage of `ioutil` package, which has been deprecated: https://go.dev/doc/go1.16#ioutil.

`ioutil` is still however used in generated files (rules.go, client,go).


matej-g (approved):
Thanks @saswatamcode, great clean-up job :broom:

PR #370, opened 2022-09-05
Add mTLS support for upstream servers

periklis:
The following PR adds support to reverse proxy calls to metrics/logs/traces endpoints using mTLS. The purpose for this to support for example a Loki cluster that is secured via mTLS on an isolated environment. In detail, Loki does not support any native authentication methods on the HTTP layer.

### DIFF
--- a/api/logs/v1/http.go
+++ b/api/logs/v1/http.go
@@ -2,8 +2,7 @@
 package http
 
 import (
-	"crypto/tls"
-	"crypto/x509"
+	stdtls "crypto/tls"
 	"net"
 	"net/http"
 	"net/http/httputil"
@@ -16,6 +15,7 @@ import (
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 
 	"github.com/observatorium/api/proxy"
+	"github.com/observatorium/api/tls"
 )
 
 const (
@@ -95,7 +95,7 @@ func (n nopInstrumentHandler) NewHandler(labels prometheus.Labels, handler http.
 	return handler.ServeHTTP
 }
 
-func NewHandler(read, tail, write *url.URL, upstreamCA []byte, opts ...HandlerOption) http.Handler {
+func NewHandler(read, tail, write *url.URL, upstreamCA []byte, upstreamCert *stdtls.Certificate, opts ...HandlerOption) http.Handler {
 	c := &handlerConfiguration{
 		logger:     log.NewNopLogger(),
 		registry:   prometheus.NewRegistry(),
@@ -122,13 +122,7 @@ func NewHandler(read, tail, write *url.URL, upstreamCA []byte, opts ...HandlerOp
 				DialContext: (&net.Dialer{
 					Timeout: ReadTimeout,
 				}).DialContext,
-			}
-
-			if len(upstreamCA) != 0 {
-				t.TLSClientConfig = &tls.Config{
-					RootCAs: x509.NewCertPool(),
-				}
-				t.TLSClientConfig.RootCAs.AppendCertsFromPEM(upstreamCA)
+				TLSClientConfig: tls.NewClientConfig(upstreamCA, upstreamCert),
 			}
 
 			proxyRead = &httputil.ReverseProxy{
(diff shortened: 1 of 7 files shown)


api/logs/v1/http.go
```diff
@@ -131,6 +131,10 @@ func NewHandler(read, tail, write *url.URL, upstreamCA []byte, opts ...HandlerOp
 				t.TLSClientConfig.RootCAs.AppendCertsFromPEM(upstreamCA)
 			}
 
+			if upstreamCert != nil {
+				t.TLSClientConfig.Certificates = append(t.TLSClientConfig.Certificates, *upstreamCert)
```
xperimental, line 135: If `upstreamCA` is empty then `TLSClientConfig` is still `nil` here. In that case it will panic here.

api/logs/v1/http.go
```diff
@@ -229,6 +233,10 @@ func NewHandler(read, tail, write *url.URL, upstreamCA []byte, opts ...HandlerOp
...
```
xperimental, line 237: `TLSClientConfig` can be `nil`.

api/logs/v1/http.go
```diff
@@ -275,6 +283,10 @@ func NewHandler(read, tail, write *url.URL, upstreamCA []byte, opts ...HandlerOp
...
```
xperimental, line 287: `TLSClientConfig` can be `nil`.

api/metrics/legacy/http.go
```diff
@@ -120,9 +120,14 @@ func NewHandler(url *url.URL, upstreamCA []byte, opts ...HandlerOption) http.Han
 			t.TLSClientConfig = &tls.Config{
 				RootCAs: x509.NewCertPool(),
 			}
+
 			t.TLSClientConfig.RootCAs.AppendCertsFromPEM(upstreamCA)
 		}
 
+		if upstreamCert != nil {
+			t.TLSClientConfig.Certificates = append(t.TLSClientConfig.Certificates, *upstreamCert)
```
xperimental, line 128: `TLSClientConfig` can be `nil`.

api/metrics/v1/http.go
```diff
@@ -205,6 +205,10 @@ func NewHandler(read, write, rulesEndpoint *url.URL, upstreamCA []byte, opts ...
 				t.TLSClientConfig.RootCAs.AppendCertsFromPEM(upstreamCA)
 			}
 
+			if upstreamCert != nil {
+				t.TLSClientConfig.Certificates = append(t.TLSClientConfig.Certificates, *upstreamCert)
```
xperimental, line 209: `TLSClientConfig` can be `nil`.

api/metrics/v1/http.go
```diff
@@ -250,6 +254,10 @@ func NewHandler(read, write, rulesEndpoint *url.URL, upstreamCA []byte, opts ...
...
```
xperimental, line 258: `TLSClientConfig` can be `nil`. I also noticed that this seems to be the only handler where `t` is initialized using `http.DefaultTransport` instead of creating a new one, not part of this PR though.

xperimental (changes requested):
Just one inline comment (repeated though).

Would it make sense to extract the `TLSClientConfig` initialization (and maybe also the CLI-parameter parsing) to a separate function that can be re-used across the handlers?

api/traces/v1/http.go
```diff
@@ -137,6 +139,17 @@ func NewV2Handler(read *url.URL, readTemplate string, opts ...HandlerOption) htt
...
 
+		if len(upstreamCA) != 0 {
+			t.TLSClientConfig = &tls.Config{
+				RootCAs: x509.NewCertPool(),
+			}
+			t.TLSClientConfig.RootCAs.AppendCertsFromPEM(upstreamCA)
+		}
+
+		if upstreamCert != nil {
+			t.TLSClientConfig.Certificates = append(t.TLSClientConfig.Certificates, *upstreamCert)
```
xperimental, line 150: `TLSClientConfig` can be `nil`.

periklis:
> Just one inline comment (repeated though).
> 
> Would it make sense to extract the `TLSClientConfig` initialization (and maybe also the CLI-parameter parsing) to a separate function that can be re-used across the handlers?

Good catch according to our CLI flags we enable upstream TLS only if we provide an upstream CA, thus I wrapped the cert inside the first if block.

PR #380, opened 2022-09-28
[WIP] Add support for storing Loki rules via rules-objstore

periklis:
## Summary

The following PR adds support for serving Loki rules via the [observatorium/rules-objstore](/bservatorium/rules-objstore) service. The underlying design supports two mutual exclusive use cases with this addition:

### Main Case: RHOBS Unified Rules management

Use the Loki Ruler API **only** for read-only access on rules. Provide the `/api/v1/rules/raw` API to create/update rules through the  [observatorium/rules-objstore](/bservatorium/rules-objstore). 

#### Required Further work

To support a unified rules management on RHOBS, the [observatorium/rules-objstore](/bservatorium/rules-objstore) requires to store per tenant and per signal source.

### Secondary Case: Kubernetes-Native Rules Management

Use the Loki Ruler API for fetching/creating/updating rules for native usage (See cli flags `logs.rules.native`). This case applies to using the observatorium-api in the Loki-Operator combined with the CLI flag `logs.rules.read-only` where rules are managed via [AlertingRule](https://loki-operator.dev/docs/api.md/#alertingrule) and/or [RecordingRule](https://loki-operator.dev/docs/api.md/#recordingrule) custom resources.

#### Required Further Work

For example for any kubernetes-native application (e.g. the Loki Operator), it is required to pass the appropriate CLI flags to the observatorium-api, i.e. `--logs.rules.native=loki-ruler.example.com` and `--logs.rules.read-only=true`. For example:

* grafana/loki#7254

PR #381, opened 2022-09-29
Seggregate read and write access to logs rules routes based on flag

periklis:
The following PR is a spin-off from the discussion on observatorium/api#380 and observatorium/rules-objstore#11 to support Loki rules management via the observatorium-api. In particular it simplifies the original design by removing the need to run the `rules-objstore` as a middle-man for Loki rules. In addition it continues to support kubernetes-native rules management via custom resources by adding the option to have a read-only access to Loki rules via the observatorium-api. In detail:

### RHOBS Rules Management

For managing Loki Rules the observatorium-api allows to proxy read and write requests to the Loki Ruler endpoints (i.e. `--logs.rules.endpont=http://loki-ruler:3100` and `--logs.rules.read-only=false`). The Loki Ruler is managing all CRUD operations on a shared object storage bucket on a distinct path, e.g. `/logs/rules`. The bucket is shared with the `rules-objstore` that manages exclusively the metrics rules on separate path, e.g. `/metrics/rules`.

### Kubernetes native Rules Management

The read and write access to Loki Rules is separated here between the observatorium-api and the kubernetes operator that provides custom resources for alerting and recording rules (See Loki Operator). The observatorium-api enables only a read access to the rules. On the other hand the custom resources for alerting and recording rules are translated into Loki Rules YAML stored in a ConfigMap.  Thus the cli flags look like `--logs.rules.endpont=http://loki-ruler.ns.svc:3100` and `--logs.rules.read-only=true`

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,8 @@ Usage of ./observatorium-api:
     	The endpoint against which to make read requests for logs.
   -logs.rules.endpoint string
     	The endpoint against which to make rules requests for logs.
+  -logs.rules.read-only
+    	Allow only read-only rule requests for logs.
   -logs.tail.endpoint string
     	The endpoint against which to make tail read requests for logs.
   -logs.tenant-header string
--- a/api/logs/v1/http.go
+++ b/api/logs/v1/http.go
@@ -122,7 +122,7 @@ func (n nopInstrumentHandler) NewHandler(labels prometheus.Labels, handler http.
 	return handler.ServeHTTP
 }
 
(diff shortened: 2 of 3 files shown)


saswatamcode (approved):
I think it makes sense to drop /api/v1/rules/raw for other signals, as no system relies on that endpoint. 
Maybe in future PR the client spec could be updated with these new endpoints?

periklis:
> I think it makes sense to drop /api/v1/rules/raw for other signals, as no system relies on that endpoint. Maybe in future PR the client spec could be updated with these new endpoints?

Yes I want to update the spec in a separate PR and start adding support to `obsctl` ASAP. This way we will get SD using it.

PR #385, opened 2022-10-05
Add logs rules client api spec

periklis:
The following PR extends the client API spec with all remaining rules APIs for retrieving/manipulating Logs-based rules.

### DIFF
--- a/client/spec.yaml
+++ b/client/spec.yaml
@@ -31,6 +31,8 @@ x-tagGroups:
       - logs/labelsv1
       - logs/labelvaluesv1
       - logs/queryv1
+      - logs/rulesrawv1
+      - logs/rulesv1
       - logs/seriesv1
       - logs/tailV1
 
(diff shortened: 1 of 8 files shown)


client/spec.yaml
```diff
@@ -495,4 +495,178 @@ paths:
...
+    get:
+      tags:
+        - logs/rulesv1
+      summary: Get Logs Rules for a tenant in JSON form
+      operationId: getAllLogsRules
+      description: |
+        You can get Logs rules for a tenant with health data
+      responses:
+        "2XX":
+          description: rule for a tenant
```
jessicalins, line 510:
```suggestion
          description: All log rules for a tenant
```
I think we could change the description to something on that direction, since `getAllLogsRules` returns more than one rule (I think the same suggestion applies to the other endpoints that return more than one rule)

client/spec.yaml
```diff
@@ -495,4 +495,178 @@ paths:
...
+        "4XX":
+          description: Bad request
+        "5XX":
+          description: Server side error
+    delete:
+      tags:
+        - logs/rulesv1
+      summary: Delte Logs Rules for a tenant in JSON form inside a namespace
```
jessicalins, line 583:
```suggestion
      summary: Delete Logs Rules for a tenant in JSON form inside a namespace
```

client/spec.yaml
```diff
@@ -495,4 +495,178 @@ paths:
...
+              schema:
+                $ref: "./responses/responses.yaml#/components/schemas/AllLogRulesResponse"
+        "4XX":
+          description: Bad request
+        "5XX":
+          description: Server side error
+  /api/logs/v1/{tenant}/loki/api/v1/rules/{namespace}:
+    parameters:
+      - $ref: ./parameters/parameters.yaml#/components/parameters/tenant
+      - $ref: ./parameters/parameters.yaml#/components/parameters/logRulesNamespace
```
jessicalins, line 522: I'm not sure how `$ref` works but does `logRulesNamespace` need to be named as `namespace` so that it can be replaced in `/api/logs/v1/{tenant}/loki/api/v1/rules/{namespace}`? or naming does not matter and it just refers sequentially to the params in the path (in this case, `tenant` -> `tenant`, `namespace` -> `logRulesNamespace`?

client/spec.yaml
jessicalins, line 583:
```suggestion
      summary: Delete Logs Rules for a tenant in JSON form inside a namespace
```

client/spec.yaml
```diff
@@ -495,4 +495,178 @@ paths:
...
+      operationId: getLogsPromAlerts
+      description: |
+        Prometheus-compatible rules endpoint to list all active alerts.
+      responses:
+        "2XX":
+          description: rule for a tenant
+          content:
+            application/json:
+              schema:
+                $ref: "./responses/responses.yaml#/components/schemas/LogRulesPrometheusResponse"
```
jessicalins, line 668: would we need to refer here to the alert rules schema instead? https://github.com/observatorium/api/blob/16e21d560de0f56477a533d4701427e3ede6584a/client/models/models.yaml#L117

jessicalins (commented):
looking good, thanks for this! :sparkle: left some suggestions/questions

client/spec.yaml
periklis: @jessicalins AFAIK the path name `namespace` needs to map into the parameter's name definition `logRulesNamespace.name`, i.e.:
```
    logRulesNamespace:
      in: path
      name: namespace
      description: namespace of the log rule
      required: true
      schema:
        type: string
```

client/spec.yaml
periklis: The chain is `LogRulesPrometheusResponse` -> `Rules` -> `RuleGroupEvaluated` and the last one has:
```
        rules:
          type: array
          items:
            oneOf:
              - $ref: "#/components/schemas/RecordingRuleEvaluated"
              - $ref: "#/components/schemas/AlertingRuleEvaluated"
```

client/spec.yaml
jessicalins: ah I see! makes sense :+1:

client/spec.yaml
```diff
@@ -495,4 +495,178 @@ paths:
...
+    get:
+      tags:
+        - logs/rulesv1
+      summary: Get Logs Rules for a tenant in JSON form inside a namespace
+      operationId: getLogsRules
+      description: |
+        You can get Logs rules for a tenant in a namespace with health data
+      responses:
+        "2XX":
+          description: rule for a tenant
```
jessicalins, line 532: I think this description and the remaining ones in this file can be updated too to fit more than 1 rule

PR #395, opened 2022-10-18
Update Golang to 1.19

periklis:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/observatorium/api
 
-go 1.17
+go 1.19
```
douglascamata, line 3: We like to keep the Go version here as low as possible. So ideally if we don't need any Go 1.19 features it should be unchanged.

squat (approved):

PR #396, opened 2022-10-20
Remove component-base for manual implementation

Red-GV:
This PR removes the `k8s.io/component-base` dependency in favor of a manual lookup for the TLS variables. `component-base` and `apiserver` rely on old versions of `otel`. This causes problems when trying to upgrade the k8s dependencies. Since `component-base` is only required to look up the TLS variables, it makes sense to replace it to reduce the number of dependencies which rely on this old library version.

### DIFF
--- a/tls/config.go
+++ b/tls/config.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
-	"k8s.io/component-base/cli/flag"
 )
 
 // NewClientConfig returns a tls config for the reverse proxy handling if an upstream CA is given.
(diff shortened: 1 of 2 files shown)


tls/config.go
```diff
@@ -81,6 +80,49 @@ func NewServerConfig(logger log.Logger, certFile, keyFile, minVersion, maxVersio
...
+
+func mapCipherNamesToIDs(rawTLSCipherSuites []string) ([]uint16, error) {
+	cipherSuites := []uint16{}
+	allCipherSuites := tlsCipherSuites()
+
+	for _, name := range cipherSuiteNames {
+		id, ok := allCipherSuites[name]
+		if !ok {
+			return nil, fmt.Errorf("unknown TLSCipherSuite: %s", rawTLSCipherSuite)
+		}
```
saswatamcode, line 119: Build fails due to this I think?
```suggestion
	for _, name := range rawTLSCipherSuites {
		id, ok := allCipherSuites[name]
		if !ok {
			return nil, fmt.Errorf("unknown TLSCipherSuite: %s", name)
		}
```

periklis (approved):

matej-g (approved):
Thanks @Red-GV

PR #403, opened 2022-11-01
Fix default TLS cipher configuration when none are specified

Red-GV:
This is a follow-up patch to #396. When there is no cipher suite set, an empty array would be set in for the `TLSConfig`. Instead, this should be nil so that the default list of cipher suites are used instead.


xperimental (approved):

saswatamcode (approved):
Thanks!

PR #409, opened 2022-11-22
*: Use 1.17 compatibility for go mod

matej-g:

It seems we're hitting an issue with updating dependencies in https://github.com/observatorium/api/pull/408 due to module graph pruning and changes between 1.16 and 1.17 (https://go.dev/ref/mod#graph-pruning). Relevant output:

```
github.com/observatorium/api imports
	github.com/ghodss/yaml imports
	gopkg.in/yaml.v2 tested by
	gopkg.in/yaml.v2.test imports
	gopkg.in/check.v1 loaded from gopkg.in/check.v1@v1.0.0-20200227125254-8fa46927fb4f,
	but go 1.16 would select v1.0.0-20201130134442-10cb98267c6c

To upgrade to the versions selected by go 1.16:
	go mod tidy -go=1.16 && go mod tidy -go=1.17
If reproducibility with go 1.16 is not needed:
	go mod tidy -compat=1.17
For other options, see:
	https://golang.org/doc/modules/pruning
```

I believe we're fine with dropping the reproducibility for 1.16 and enforcing the compatibility with 1.17 mode.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ build: $(BIN_NAME)
 
 .PHONY: deps
 deps: go.mod go.sum
-	go mod tidy
+	go mod tidy -compat=1.17
 	go mod download
 	go mod verify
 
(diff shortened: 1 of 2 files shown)


jessicalins (approved):
looks good, thanks!

PR #415, opened 2022-12-09
Align gRPC middleware with HTTP to send tenant ID upstream 

pavolloffay:
Signed-off-by: Pavol Loffay <p.loffay@gmail.com>

### DIFF
--- a/authentication/grpc.go
+++ b/authentication/grpc.go
@@ -49,6 +49,10 @@ func WithGRPCTenantHeader(header string, tenantIDs map[string]string, logger log
 			// codes.Unauthenticated and not explain about the tenant if this is a concern.)
 			return status.Error(codes.InvalidArgument, "unknown tenant")
 		}
+		// The tenant header contains the tenant name.
+		// It needs to be overridden to send the ID to match the HTTP functionality from openshift.WithTenantHeader.
+		md.Set(header, id)
+		ctx = metadata.NewIncomingContext(ctx, md)
 		ctx = context.WithValue(ctx, tenantIDKey, id)
 
 		wrapped := grpc_middleware.WrapServerStream(ss)


authentication/grpc.go
```diff
@@ -49,6 +49,10 @@ func WithGRPCTenantHeader(header string, tenantIDs map[string]string, logger log
 			// codes.Unauthenticated and not explain about the tenant if this is a concern.)
 			return status.Error(codes.InvalidArgument, "unknown tenant")
 		}
+		// The tenant header contains the tenant name.
+		// It needs to be overridden to send the ID to match the HTTP functionality from openshift.WithTenantHeader.
+		md.Set(header, id)
+		ctx = metadata.NewIncomingContext(ctx, md)
```
pavolloffay, line 55: @periklis could you please explain how the tenant header is sent to in Loki? 

For tempo the ingestion clients (OTELcol) are required to send:
*  `X-Scope-OrgID` to identify the tenant via name e.g. `X-Scope-OrgID: test-oidc` 
* `Authorization` header with the service account token

Both headers are validated, the tenant must be defined in the `tenants.yaml` and the token is validated via token review API.

The `openshift.WithTenantHeader` sends upstream the tenant ID (UID) and not the name. Therefore Tempo will use the UID as the tenant name. @periklis is Loki using UIDs as well?

periklis: Yes IIRC we use the tenant ID for sending to an upstream Loki. For example in RHOBS tenant ID is not the same as tenant name, thus RHOBS' Loki will have the UUID stored. In Loki-Operator we aligned the IDs with the name for openshift-logging as per having only three unique and simple to separate tenants (application, infrastructure, audit).

periklis (approved):

saswatamcode (approved):
LGTM

PR #421, opened 2022-12-13
Ensure rate limiting middleware called before any paths are stripped

### LINKED ISSUE #420: Breaking change in rate limiting endpoint matching
It appears that #293 introduced a breaking change to rate limiting functionality. 

Prior to this change, the entire path was passed to the rate limiting middleware for the `/api/metrics/v1/{tenant}` route 
before

After, the path has been stripped from the request object passed to the middleware which has the potential to cause pre-existing endpoint regular expressions to no longer match and thus, bypass the rate limiting functionality.

  
after

Although the `RequestURI` is correct, the middleware behaviour which uses `URL.Path` is affected.

PhilipGough:
Fixes #420

### DIFF
--- a/main.go
+++ b/main.go
@@ -584,17 +584,21 @@ func main() {
 
 			// Metrics.
 			if cfg.metrics.enabled {
-				var metricsMiddlewares []func(http.Handler) http.Handler
-				tenantMiddlewares := authentication.WithTenantMiddlewares(pm.Middlewares)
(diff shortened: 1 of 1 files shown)


saswatamcode (approved):
LGTM! Thanks for fixing this! 🌟

moadz:
I have an e2e test to reproduce this on my local copy, will see if I can get it raised before I leave for christmas.

PR #422, opened 2022-12-14
Fix parsing LogQL aliased filter expressions

periklis:
Parsing aliased filter expression like this:
```
{ kubernetes_pod_name="logger-5d89f8d985-4pxm4" } | json | level=~"critical|emerg|fatal|alert|crit" 
```
resulted into the wrong expression missing the `level=~` alias:
```
{ kubernetes_pod_name="logger-5d89f8d985-4pxm4" } | json | "critical|emerg|fatal|alert|crit" 
```

Ref [LOG-3397](https://issues.redhat.com/browse/LOG-3397)

cc @xperimental

### DIFF
--- a/logql/v2/parser_test.go
+++ b/logql/v2/parser_test.go
@@ -106,11 +106,13 @@ func TestParseExpr(t *testing.T) {
 			},
 		},
 		{
-			input: `{ first = "value" }|logfmt|addr>=ip("1.1.1.1")`,
+			input: `{first="value"} | logfmt | addr>=ip("1.1.1.1")`,
 			expr: &LogQueryExpr{
 				filter: LogPipelineExpr{
 					{
-						parser: "logfmt",
+						expr: &LogParserExpr{
+							parser: "logfmt",
+						},
 					},
 					{
 						stages: LogFiltersExpr{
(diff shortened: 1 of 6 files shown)


logql/v2/parser_test.go
```diff
@@ -440,7 +480,7 @@ func TestParseExpr(t *testing.T) {
 		},
 		// log query expressions with format expressions
 		{
-			input: `{app="first"} |= "value" | json | line_format "loop{{ .first }}blop {{.status_code}}" | label_format first=value,status_code="blop{{.value}}"`, //nolint:lll
+			input: `{app="first"} |= "value" | json | line_format "loop{{ .first }}blop {{.status_code}}" | label_format first="value",status_code="blop{{.value}}"`, //nolint:lll
```
xperimental, line 483: Per [LogQL docs](https://grafana.com/docs/loki/latest/logql/log_queries/#labels-format-expression) there's a difference between `first=value` and `first="value"`, the former referring to a `value` label and the latter producing `value` as literal value.

xperimental (approved):
Works fine with the log-queries I threw at it.

Some metric-queries did not work, but I'm not sure if they should already be working, so approving anyway.

Issue #423, opened 2022-12-15
Make e2e tests arm64 friendly


douglascamata:
They do not work on arm64 right now. 

Will possibly need an upgrade on github.com/efficientgo dependencies, which includes breaking changes. 

Potentially will require to add multi-arch images to other github.com/observatorium projects first.

PR #424, opened 2022-12-15
Fix rate limiter window duration

douglascamata:
Gubernator's duration parameter should be in milliseconds, according to their own documentation: https://github.com/mailgun/gubernator/blob/v1.0.1-rc.1/README.md?plain=1#L41.

Unfortunately I cannot add an e2e test for this right now because:

* This is is kind of urgent. 
* The e2e tests aren't (yet) arm64 friendly and I work on arm64. 
* To make them arm64 friendly we need to upgrade github.com/efficientgo dependencies (there are breaking changes) and potentially create multi-arch images for other projects in this org. I created #423 to track this.

I found the problem and solution by spinning up Gubernator and manually calling the gRPC service.


PhilipGough (approved):
lgtm

squat (approved):

PR #430, opened 2022-12-20
Metrics middleware reorg

PhilipGough:
There is a problem in the existing setup of the handlers where if a middleware causes the HTTP response to return early, the Prometheus metrics for the HTTP handlers never get to see the request/response.

This PR attempts to reorganise the structure in a backwards compatible way by moving the instrumentation middleware to the top of the handlers stack enabling us to count all response codes in case of failures (rate limiting, auth, etc).

There also appears to be an issue where some of the middlewares will have been called multiple times as they were appended in both `main.go` and the specific handlers. I have adjusted these so they will get called in their respective handler. 

`stripTenantPrefix` now moves to middleware called in each handler since we want the middlewares to receive the correct path and strip it after the fact.

! Needs to be followed up with some additional tests and if we get agreement, take care of similar for logs and traces

### DIFF
--- a/api/metrics/legacy/http.go
+++ b/api/metrics/legacy/http.go
@@ -14,10 +14,14 @@ import (
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 
 	"github.com/observatorium/api/proxy"
+	"github.com/observatorium/api/server"
 	"github.com/observatorium/api/tls"
 )
 
 const (
+	QueryRoute      = "/api/v1/query"
+	QueryRangeRoute = "/api/v1/query_range"
+
 	readTimeout = 15 * time.Minute
 )
 
@@ -28,6 +32,7 @@ type handlerConfiguration struct {
 	spanRoutePrefix  string
 	queryMiddlewares []func(http.Handler) http.Handler
 	uiMiddlewares    []func(http.Handler) http.Handler
+	labelParser      func(r *http.Request) prometheus.Labels
 }
 
 type HandlerOption func(h *handlerConfiguration)
@@ -79,6 +84,14 @@ func WithGlobalMiddleware(m ...func(http.Handler) http.Handler) HandlerOption {
 	}
 }
 
+// WithLabelParser adds a custom label parser to the handler.
+// The label parser is used to parse prometheus.Labels from the request.
+func WithLabelParser(labelParser func(r *http.Request) prometheus.Labels) HandlerOption {
+	return func(h *handlerConfiguration) {
+		h.labelParser = labelParser
+	}
+}
+
 type handlerInstrumenter interface {
 	NewHandler(labels prometheus.Labels, handler http.Handler) http.HandlerFunc
 }
@@ -101,6 +114,10 @@ func NewHandler(url *url.URL, upstreamCA []byte, upstreamCert *stdtls.Certificat
 	}
 
 	r := chi.NewRouter()
+	r.Use(server.InstrumentationMiddleware(c.labelParser))
+	r.Use(func(handler http.Handler) http.Handler {
+		return c.instrument.NewHandler(nil, handler)
+	})
 
 	var legacyProxy http.Handler
 	{
@@ -126,19 +143,8 @@ func NewHandler(url *url.URL, upstreamCA []byte, upstreamCert *stdtls.Certificat
 
 	r.Group(func(r chi.Router) {
 		r.Use(c.queryMiddlewares...)
-		const (
-			queryRoute      = "/api/v1/query"
-			queryRangeRoute = "/api/v1/query_range"
-		)
-
-		r.Handle(queryRoute, c.instrument.NewHandler(
-			prometheus.Labels{"group": "metricslegacy", "handler": "query"},
-			otelhttp.WithRouteTag(c.spanRoutePrefix+queryRoute, legacyProxy),
-		))
-		r.Handle(queryRangeRoute, c.instrument.NewHandler(
-			prometheus.Labels{"group": "metricslegacy", "handler": "query_range"},
-			otelhttp.WithRouteTag(c.spanRoutePrefix+queryRangeRoute, legacyProxy),
-		))
+		r.Handle(QueryRoute, otelhttp.WithRouteTag(c.spanRoutePrefix+QueryRoute, legacyProxy))
+		r.Handle(QueryRangeRoute, otelhttp.WithRouteTag(c.spanRoutePrefix+QueryRangeRoute, legacyProxy))
 	})
 
 	r.Group(func(r chi.Router) {
--- a/api/metrics/v1/http.go
+++ b/api/metrics/v1/http.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/observatorium/api/proxy"
 	"github.com/observatorium/api/rules"
+	"github.com/observatorium/api/server"
 	"github.com/observatorium/api/tls"
 )
 
@@ -25,15 +26,15 @@ const (
 )
 
 const (
-	uiRoute          = "/"
-	queryRoute       = "/api/v1/query"
-	queryRangeRoute  = "/api/v1/query_range"
-	seriesRoute      = "/api/v1/series"
-	labelNamesRoute  = "/api/v1/labels"
-	labelValuesRoute = "/api/v1/label/{label_name}/values"
-	receiveRoute     = "/api/v1/receive"
-	rulesRoute       = "/api/v1/rules"
-	rulesRawRoute    = "/api/v1/rules/raw"
+	UIRoute          = "/"
+	QueryRoute       = "/api/v1/query"
+	QueryRangeRoute  = "/api/v1/query_range"
+	SeriesRoute      = "/api/v1/series"
+	LabelNamesRoute  = "/api/v1/labels"
+	LabelValuesRoute = "/api/v1/label/{label_name}/values"
+	ReceiveRoute     = "/api/v1/receive"
+	RulesRoute       = "/api/v1/rules"
+	RulesRawRoute    = "/api/v1/rules/raw"
 )
 
 type handlerConfiguration struct {
@@ -46,6 +47,7 @@ type handlerConfiguration struct {
 	readMiddlewares  []func(http.Handler) http.Handler
 	uiMiddlewares    []func(http.Handler) http.Handler
 	writeMiddlewares []func(http.Handler) http.Handler
+	labelParser      func(r *http.Request) prometheus.Labels
 }
 
 // HandlerOption modifies the handler's configuration.
@@ -124,6 +126,14 @@ func WithGlobalMiddleware(m ...func(http.Handler) http.Handler) HandlerOption {
 	}
 }
 
+// WithLabelParser adds a custom label parser to the handler.
+// The label parser is used to parse prometheus.Labels from the request.
+func WithLabelParser(labelParser func(r *http.Request) prometheus.Labels) HandlerOption {
+	return func(h *handlerConfiguration) {
+		h.labelParser = labelParser
+	}
+}
+
 type handlerInstrumenter interface {
 	NewHandler(labels prometheus.Labels, handler http.Handler) http.HandlerFunc
 }
@@ -141,6 +151,9 @@ func NewHandler(read, write, rulesEndpoint *url.URL, upstreamCA []byte, upstream
 		logger:     log.NewNopLogger(),
 		registry:   prometheus.NewRegistry(),
 		instrument: nopInstrumentHandler{},
+		labelParser: func(r *http.Request) prometheus.Labels {
+			return nil
+		},
 	}
 
 	for _, o := range opts {
@@ -148,6 +161,10 @@ func NewHandler(read, write, rulesEndpoint *url.URL, upstreamCA []byte, upstream
 	}
 
 	r := chi.NewRouter()
+	r.Use(server.InstrumentationMiddleware(c.labelParser))
+	r.Use(func(handler http.Handler) http.Handler {
+		return c.instrument.NewHandler(nil, handler)
+	})
 
 	if read != nil {
 		var proxyQuery http.Handler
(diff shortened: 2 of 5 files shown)


api/metrics/v1/http.go
```diff
@@ -153,6 +153,7 @@ func NewHandler(read, write, rulesEndpoint *url.URL, upstreamCA []byte, upstream
 	r.Use(func(handler http.Handler) http.Handler {
 		return c.instrument.NewHandler(nil, handler)
 	})
+	r.Use(server.StripTenantPrefix("/api/metrics/v1"))
```
PhilipGough, line 156: if this happens here, any middleware which expects the original path to be untouched (rate limiter for example) will be broken. Hence why it was applied directly after the middleware in each case

instrumentation/http_instrumentation.go
```diff
@@ -102,24 +102,70 @@ func (m instrumentedHandlerFactory) NewHandler(extraLabels prometheus.Labels, ne
 }
 
 // NewInstrumentedHandlerFactory creates a new instrumentedHandlerFactory.
-func NewInstrumentedHandlerFactory(req *prometheus.Registry, hardcodedLabels []string) instrumentedHandlerFactory {
+func NewInstrumentedHandlerFactory(req *prometheus.Registry) instrumentedHandlerFactory {
```
PhilipGough, line 105: this helps organise the code better, but its a breaking change for anyone that relies on this package.

instrumentation/http_instrumentation.go
```diff
@@ -62,15 +65,12 @@ type instrumentedHandlerFactory struct {
...
 func (m instrumentedHandlerFactory) NewHandler(extraLabels prometheus.Labels, next http.Handler) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		// if no extra labels are provided check for them in the context
 		if len(extraLabels) == 0 {
-			if labels := r.Context().Value(ExtraLabelContextKey); labels != nil {
-				ctxLabels, ok := labels.(prometheus.Labels)
-				if ok {
-					extraLabels = ctxLabels
-				}
+			if labels := chiLabelParser(r); len(labels) != 0 {
```
PhilipGough, line 72: something to keep in mind is that moving this from being injectable to being hardcoded here makes the package unusable with any other routing framework

main.go
```diff
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"github.com/observatorium/api/instrumentation"
```
PhilipGough, line 11: a nit but for this and others, we might want to retain the existing organisation of imports

instrumentation/http_instrumentation.go
douglascamata: I fixed all of usages 🤘

main.go
douglascamata: Ah, the order of imports. Let me fix my IDE config. 😂

api/metrics/v1/http.go
douglascamata: Hummm, I wasn't sure about it. I will spend some time thinking hard about the need for this `StripTenantPrefix` thing. I want to get rid of it, because modifying the path like this is kinda confusing and annoying for all the subsequent middlewares.

PhilipGough: it i there because without it, we would need to add a handler for each path for each tenant no?

api/logs/v1/http.go
```diff
@@ -150,6 +151,7 @@ func NewHandler(read, tail, write, rules *url.URL, rulesReadOnly bool, upstreamC
 	}
 
 	r := chi.NewRouter()
+	r.Use(server.StripTenantPrefix("/api/v1/logs"))
```
saswatamcode, line 154: Maybe we can do logs and traces separately as a follow-up PR?

api/metrics/v1/http.go
saswatamcode: +1 to what @PhilipGough mentioned

instrumentation/http_instrumentation.go
```diff
@@ -19,36 +24,36 @@ type httpMetricsCollector struct {
 }
 
 // newHTTPMetricsCollector creates a new httpMetricsCollector.
-func newHTTPMetricsCollector(reg *prometheus.Registry, hardcodedLabels []string) httpMetricsCollector {
+func newHTTPMetricsCollector(reg *prometheus.Registry) httpMetricsCollector {
```
saswatamcode, line 27: Maybe let's not remove this functionality and break API? For any other users of this package, `hardcodedLabels` would have made a lot of sense and they can instrument different things as they want to.

instrumentation/http_instrumentation.go
```diff
@@ -19,36 +24,36 @@ type httpMetricsCollector struct {
...
 	m := httpMetricsCollector{
 		requestCounter: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
 			Name: "http_requests_total",
 			Help: "Counter of HTTP requests.",
 		},
-			append(hardcodedLabels, "code", "method", "tenant"),
+			[]string{"group", "handler", "code", "method", "tenant"},
```
saswatamcode, line 33: In fact, I think the rest of the labels can also be added to `hardcodedLabels`?
```suggestion
			hardcodedLabels,
```

instrumentation/http_instrumentation.go
```diff
@@ -91,10 +102,67 @@ func (m instrumentedHandlerFactory) NewHandler(extraLabels prometheus.Labels, ne
...
+	return extraLabels
+}
+
+type groupHandler struct {
+	group   string
+	handler string
+}
+
+var legacyMetricsGroup = map[string]groupHandler{
+	metricslegacy.QueryRoute:      {"metricslegacy", "query"},
```
saswatamcode, line 152: Is it necessary to move these here and import `metricslegacy` or `metricsv1` in `instrumentation`? I fear that at some point this would lead to cyclic dependencies between things, when someone tries to import instrumentation stuff in the metrics packages. 

Maybe we can just add these maps to `main.go` or even in the `metrics` package itself?

instrumentation/http_instrumentation.go
```diff
@@ -91,10 +102,67 @@ func (m instrumentedHandlerFactory) NewHandler(extraLabels prometheus.Labels, ne
...
 // NewInstrumentedHandlerFactory creates a new instrumentedHandlerFactory.
-func NewInstrumentedHandlerFactory(req *prometheus.Registry, hardcodedLabels []string) instrumentedHandlerFactory {
+func NewInstrumentedHandlerFactory(req *prometheus.Registry) instrumentedHandlerFactory {
 	return instrumentedHandlerFactory{
-		metricsCollector: newHTTPMetricsCollector(req, hardcodedLabels),
+		metricsCollector: newHTTPMetricsCollector(req),
+	}
+}
+
+func httpHandlerLabels(r *http.Request) prometheus.Labels {
```
saswatamcode, line 111: Ok, so this seems very rigid and hardcoded. Why not keep it injectable as @PhilipGough mentioned and did previously here? 
> something to keep in mind is that moving this from being injectable to being hardcoded here makes the package unusable with any other routing framework

Is there some bug/issue you noticed with keeping it injectable @douglascamata?

instrumentation/http_instrumentation.go
```diff
@@ -1,13 +1,18 @@
-package server
+package instrumentation
```
saswatamcode, line 1: Just a small nit, but we do have separate instrumentation code than this like `logger.NewLogger`. So this might lead to some confusion for contributors  (for eg which to use `instrumentation.Logger` or `logger.NewLogger`?).

Maybe we can rename this package to `http_instrumention` instead?

saswatamcode (commented):
Thanks for awesome work @PhilipGough and @douglascamata. 🌟 

I think the initial approach of injecting this labelParser functionality onto handlers works really well! Also, let's do logging and tracing in separate PR, as they might have different requirements of metrics.

Some small nits! 🙂

instrumentation/http_instrumentation.go
douglascamata: I fixed all usages of this package.

instrumentation/http_instrumentation.go
douglascamata: Not exactly. The meaning of hardcoded labels is that they would always have the same values for each handler where they are used. The others will depend on the request and thus aren't hardcoded labels.

instrumentation/http_instrumentation.go
douglascamata: I think we should solve this from a different perspective. First, adding yet more things in `main.go` is becoming unsustainable: it has around 1k lines already and becomes a bucket where we throw all kinds of logic. Second, this map is meant only for **instrumentation** purposes: the group and handler labels. So I think it makes total sense to move them here, together with the rest of the instrumentation code.

In the `api/metrics` package we have only handlers being set up.

instrumentation/http_instrumentation.go
douglascamata: There's no reason to keep the added complexity if in reality we will not migrate to any different routing framework any time soon. We need something simple that works with what we have. Besides, how is this more rigid and hardcoded than the previous solution?

instrumentation/http_instrumentation.go
douglascamata: It shouldn't lead to confusion: `instrumentation.Logger` has documentation explaining it's a logger middleware. We can guess, because of the lack of documentation, that `logger.NewLogger` creates a logger because of its name and the package where it lives.

instrumentation/http_instrumentation.go
douglascamata: In the end, the handlers themselves are using `Chi`, so they pretty much have a hard dependency on the exact router system we are using.

instrumentation/http_instrumentation.go
douglascamata: Plus this is an unexported function.

api/logs/v1/http.go
douglascamata: I prefer a single one, as this started, to make it easier to deploy/revert.

instrumentation/http_instrumentation.go
PhilipGough: > I fixed all the usages

This is not possible since we don't own all the code that can depend on this. We have fixed it for our usages within this project, but it is a breaking change.

douglascamata: The introduction of `hardcodedLabels` here in this function was a breaking change on its own and it happened early this year. I don't think we have to worry about this also because Observatorium API didn't reach v1 yet and thus there's no guarantee of backwards compatibility in the terms of SemVer... but there isn't also a mention whether the project adheres to SemVer or not.

PhilipGough: 👍  yeah, lets wait and see what maintainers think

saswatamcode: I see. Sure, maybe breaking the API won't really affect anyone practically, but this still is reducing functionality, no? So in the future, if we ever needed hardcoded labels, we need to add them to every HTTP metric again, instead of just appending to slice.

I'm not entirely sure I get what the motivation behind this refactor is and why on this middleware reorg PR. 😅 
Maybe we can iterate on this in a separate PR @douglascamata?

main.go
saswatamcode: Yup, goimports is needed. Raised relevant PR: https://github.com/observatorium/api/pull/431

instrumentation/http_instrumentation.go
douglascamata: The refactor is needed to fix issues with the HTTP instrumentation middleware that doesn't monitor any HTTP interaction that happens before this middleware runs. And there are plenty of other middlewares that are mounted before the HTTP instrumentation that could fail, like rate limiting.

It's a big chains of problems and solutions that cannot be solved in multiple PRs, because everything depends on each other. We would love to have a minimal PR to fix this, but even after talking a thinking a lot about this we couldn't come up with a better approach. 

I already have commits prepared to reintroduce the labeling on a per handler basis.

douglascamata: We can try again to separate this, but it's hard to move a piece of the domino here without all the others falling. 😅

saswatamcode: Sorry if I wasn't clear. 😅 

I agree that this middleware refactor PR is necessary, but I didn't understand the reason behind removing an arg in `NewInstrumentedHandlerFactory` in this commit: https://github.com/observatorium/api/pull/430/commits/f087e9cd3b559892487476aa28c98486aae4012d?

douglascamata:
FYI I separated the fix for the position of the HTTP instrumentation middleware in the stack and the required changes for the `handler` and `group` labels, scoped only to the v1 metrics handler in #432.

main.go
```diff
@@ -1320,3 +1317,60 @@ func newGRPCServer(cfg *config, tenantHeader string, tenantIDs map[string]string
...
+	metricsv1.UIRoute:          {"metricsv1", "ui"},
+	metricsv1.QueryRoute:       {"metricsv1", "query"},
+	metricsv1.QueryRangeRoute:  {"metricsv1", "query_range"},
+	metricsv1.SeriesRoute:      {"metricsv1", "series"},
+	metricsv1.LabelNamesRoute:  {"metricsv1", "labels"},
+	metricsv1.LabelValuesRoute: {"metricsv1", "labelvalues"},
+	metricsv1.ReceiveRoute:     {"metricsv1", "receive"},
+	metricsv1.RulesRoute:       {"metricsv1", "rules"},
+	metricsv1.RulesRawRoute:    {"metricsv1", "rules"},
+}
```
saswatamcode, line 1341: Maybe we can put the `groupHandler` struct in `server` package and the `metricsV1Group` and `legaceMetricsGroup` maps in the `legacy` and `v1` package? Then import them here to use in `chiLabelParser`?

api/metrics/legacy/http.go
```diff
@@ -79,6 +84,13 @@ func WithGlobalMiddleware(m ...func(http.Handler) http.Handler) HandlerOption {
 	}
 }
 
+// WithLabelParser adds a custom label parser to the handler.
```
saswatamcode, line 87: Maybe a line on what the label parser aims to do here would be great?

saswatamcode (dismissed):
Thanks for awesome work!

LGTM! Just a couple of comments

api/metrics/legacy/http.go
PhilipGough: done

main.go
PhilipGough: I agree with the sentiment but the contents are tied to the input in this package/file on https://github.com/observatorium/api/blob/main/main.go#L509 so if its not a blocker, I suggest we leave as is for now.

I do have some ideas for restructuring though to make it cleaner overall that I can add in a follow on PR

saswatamcode: Ack! We can iterate on this then.

saswatamcode (approved):
LGTM!

We can iterate on refactors. 🙂

PR #431, opened 2022-12-23
*: Make formatting and linting consistent on CI

saswatamcode:
Currently, linting and formatting do not work fully and lint CI fails with `golangci-lint` panics too, but are always green ([example](https://app.circleci.com/pipelines/github/observatorium/api/1065/workflows/75308e0a-b93e-422e-89e1-51f995c92359/jobs/5390)).

This PR ensures that we lint and format correctly and fail on CI if there are changes to what is committed.
This PR,
- Adds faillint, misspell and goimports. Also, bumps golangci-lint to latest ver.
- Updates Makefile `format` and `lint` directive and adds function to check git status
- Updates golangci config to be similar to current ones used in ecosystem ([Thanos](https://github.com/thanos-io/thanos/blob/main/.golangci.yml))
- Updates go files to follow linting/formatting
- Updates metrics `observatorium_api_tenants_failed_registrations` and `observatorium_api_tenants_skipped_invalid_configuration` to `observatorium_api_tenants_failed_registrations_total` and `observatorium_api_tenants_skipped_invalid_configuration_total` to fix promlint issue

Feel free to suggest any other linters/tools that might be needed here. 🙂


danielmellado:
@bwplotka @brancz PTAL ;)

bwplotka (approved):

.bingo/Variables.mk
```diff
@@ -21,95 +21,113 @@ BINGO := $(GOBIN)/bingo-v0.3.0
 $(BINGO): $(BINGO_DIR)/bingo.mod
 	@# Install binary/ries using Go 1.14+ build command. This is using bwplotka/bingo-controlled, separate go module with pinned dependencies.
 	@echo "(re)installing $(GOBIN)/bingo-v0.3.0"
-	@cd $(BINGO_DIR) && $(GO) build -mod=mod -modfile=bingo.mod -o=$(GOBIN)/bingo-v0.3.0 "github.com/bwplotka/bingo"
+	@cd $(BINGO_DIR) && GOWORK=off $(GO) build -mod=mod -modfile=bingo.mod -o=$(GOBIN)/bingo-v0.3.0 "github.com/bwplotka/bingo"
```
periklis, line 24: You are using to bingo v0.7 to render this file but the definition here is still stuck to v0.3.0.

.golangci.yml
```diff
@@ -5,6 +5,42 @@ run:
...
+  print-issued-lines: true
+
+  # print linter name in the end of issue text, default is true
+  print-linter-name: true
+
+linters:
+  enable:
+    # Sorted alphabetically.
+    - errcheck
+    - gochecknoglobals
```
matej-g, line 28: I would not even mind also removing `gochecknoglobals` since rely on globals in some parts of code and we're fine living with it, but that's optionally.

matej-g (approved):
Even though it's only January, this is what I call spring cleaning 🧹 , nicely done @saswatamcode!

Just FYI, any upstream repo's should also reflect metrics changes in order no to break any alerting rules 😉

authentication/metrics.go
```diff
@@ -9,7 +9,7 @@ func RegisterTenantsFailingMetric(reg prometheus.Registerer) *prometheus.Counter
 	return promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
 		Namespace: "observatorium",
 		Subsystem: "api",
-		Name:      "tenants_failed_registrations",
+		Name:      "tenants_failed_registrations_total",
```
periklis, line 12: WoW this is a dealbreaker of prom naming conventions we are supposed to be an observability team :dagger:

periklis (approved):
Only nitpicks, go ahead. Thanks a lot for resolving this :clap:

Makefile
```diff
@@ -70,16 +92,37 @@ deps: go.mod go.sum
...
 
 .PHONY: shellcheck
 shellcheck: $(SHELLCHECK)
 	$(SHELLCHECK) $(shell find . -type f -name "*.sh" -not -path "*vendor*" -not -path "*tmp*")
 
 .PHONY: lint
-lint: $(GOLANGCI_LINT) deps shellcheck jsonnet-fmt
-	$(GOLANGCI_LINT) run -v --enable-all -c .golangci.yml
+lint: ## Runs various static analysis against our code.
+lint: $(FAILLINT) $(GOLANGCI_LINT) $(MISSPELL) build docs check-git deps shellcheck jsonnet-fmt format
```
douglascamata, line 115: @saswatamcode Could we not make `lint` depend on `check-git`? This basically forces people to commit to be able to run the `lint` target and it creates annoying ergonomics, i.e. forcing people to use git flows they might not be used to or simply coming to the Makefile and removing the `check-git` code (which I literally do 100% of the time).

.bingo/golangci-lint.mod
```diff
@@ -2,4 +2,4 @@ module _ // Auto generated by https://github.com/bwplotka/bingo. DO NOT EDIT
 
 go 1.15
```
PhilipGough, line 3: Why does this not end up aligning with the other files at 1.19?

Makefile
douglascamata: Also seems weird to me that we run all these linters but still have to run the actual `build` and then after building we run `deps and then `format`.... it seems too much and completely out of order.

.golangci.yml
saswatamcode: Ack, let me remove! 
Thanks for your hint about downstream, I did have that in mind! 🙂

.bingo/Variables.mk
saswatamcode: Will update!

authentication/metrics.go
saswatamcode: 😅

Makefile
saswatamcode: I agree about the ordering, it seems like it isn't accurate as I took it from obsctl, I will update this to be specific.

About `check-git`, that only checks if you have git installed. 
I think you meant `require_clean_work_tree`, which is pretty standard and used across many of our upstream projects like Thanos. The flow is made for CI too and is ideally something like generate+format+build -> check if that changes anything in tree -> lint -> check if linters change anything in tree. If it fails at any step, the next ones aren't run and you get notified by CI.

It doesn't force anything though, and it is just convenient to commit code and then lint to detect issues and fix them in one go? Or just push and fix if and when CI fails?
And we squash merge as it is, so extraneous commits are not problematic.

Wdyt? 🙂

douglascamata: I meant the `require_clean_work_tree` indeed. If CI needs a clean work tree check after linting I believe it should call a separate `make check-work-tree` target after linting. Why? Because we're mixing things. I absolutely do not need a clean work tree to run linters, it's an arbitrary requirement that creates an obstacle to linting the project. 

Some might even say linters aren't supposed to format code, this is responsibility of formatters, hinting that they should be separate independent Makefile targets. But this is not the point now. 

> It doesn't force anything though, and it is just convenient to commit code and then lint to detect issues and fix them in one go? Or just push and fix if and when CI fails?

It's forcing people to commit before running linters, **unless** they remove it in their local copies. Some people might enjoy working with committing all the time and squashing, ammending, doing fixups, push `--force-with-lease`, etc... but others might not. For example, I prefer to code, lint the project with a dirty work tree, fix the issues, then commit. This is why you often see commits where I accidentally comment out the `require_clean_work_tree` check.

PR #432, opened 2022-12-23
Proposal: Fix HTTP instrumentation for metricsv1 handler

douglascamata:

## Motivation

This is an idea of how to separate the work from #430 into smaller parts.

## Problem

Currently the instrumentation middleware is not catching all requests. There are plenty of middlewares being set up before instrumentation (i.e. the timeout middleware, rate limit middleware) that can finish a request early.

## Changes

- HTTP instrumentation middleware moved to as high as possible on the middlestack, before anything that could finish the request early before it's instrumented.
  - Changed the hardcoded labels per handler mechanism to accomodate this. Note This was never meant to be a flexible mechanism to add arbitrary labels to the HTTP instrumentation metrics, as such labels have to be defined in the metrics first.
  - Created two small middlewares to add the "group" and "handler" labels
    - Used them everywhere in the "metricsv1" group as an example.

### DIFF
--- a/api/metrics/v1/http.go
+++ b/api/metrics/v1/http.go
@@ -8,6 +8,8 @@ import (
 	"net/url"
 	"time"
 
+	"github.com/observatorium/api/server"
+
 	"github.com/go-chi/chi"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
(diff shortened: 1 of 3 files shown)


douglascamata:
Closing this for now. Talked about it with @PhilipGough and we decided to favor #430 in its original and later decide about any follow up refactor.

PR #436, opened 2023-01-13
logql: Fix parsing nested label filter expressions

periklis:
The following PR implements parsing nested label filters as: 
`{label="value"} | json | level="info" level="error"`
`{label="value"} | json | level="info" and level="error"`
`{label="value"} | json | level="info" or level="error"`
`{label="value"} | json | level="info",  level="error"`

In addition it simplifies the internal AST for pipeline stages by removing unnecessary nesting structs (e.g. `LogFiltersExpr` and `LogPipelineStageExpr`). Every pipeline stage implementation requires to implement the `LogStageExpr` interface.

Fixes: [LOG-3397](https://issues.redhat.com/browse/LOG-3397) and [LOG-3441](https://issues.redhat.com/browse/LOG-3441)

/cc @xperimental

### DIFF
--- a/logql/v2/expr.y
+++ b/logql/v2/expr.y
@@ -10,14 +10,14 @@ import (
 %union{
   Expr                 Expr
   LogBinaryOpExpr      LogBinaryOpExpr
-  LogFilterExpr        LogFilterExpr
-  LogFiltersExpr       LogFiltersExpr
-  LogFormatExpr        *LogFormatExpr
+  LogFilterExpr        LogStageExpr
+  LogLabelFilterExpr   LogStageExpr
+  LogFormatExpr        LogStageExpr
   LogQueryExpr         LogSelectorExpr
   LogMetricExpr        LogMetricSampleExpr
   LogNumberExpr        LogNumberExpr
   LogPipelineExpr      LogPipelineExpr
-  LogPipelineStageExpr LogPipelineStageExpr
+  LogStageExpr         LogStageExpr
   LogRangeQueryExpr    LogSelectorExpr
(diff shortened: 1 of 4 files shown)


logql/v2/expr.y
```diff
@@ -104,42 +104,62 @@ convOp:
...
+                  $$.right = $3
+                  $$.right.isNested = true
+                }
+        |       logLabelFilterExpr COMMA logLabelFilterExpr
+                {
+                  $$ = $1
+                  $$.chainOp = ","
+                  $$.right = $3
+                  $$.right.isNested = true
+                }
```
xperimental, line 150: `$1` always seems to be the "first filter", so filters having multiple terms will be truncated to the first and last:

```
{first="value"} | json | level="info" or level="notice" or level="warn"
```
becomes
```
{first="value"} | json | level="info" or level="warn"
```

I'm not sure if this might be "by design" in yacc and we should switch to an "append-like" structure with filters as well. I think a `$1.chain($3)` method call that traverses the single-linked list of filters and appends the new one to it should be a fix for this issue.

xperimental (commented):
Tested it on a cluster, it works with the queries provided in the issues, but I found a new one that does not work correctly.

logql/v2/expr.y
periklis: You are particularly right about yacc it is the nature of descending the structure when you enter a type that `$1` sticks to the original entry point. My mistake here is that I think recursion is the solution for this, although yacc does not work this way.

I changed this to an append like structure that is used elsewhere too (e.g. `LogPipelineExpr`)

xperimental (approved):

PR #440, opened 2023-01-18
Allow multiple matchers for NetObserv multi-tenancy (NETOBSERV-171)

jotak:
JIRA: https://issues.redhat.com/browse/NETOBSERV-171

In short, without this PR, only one matcher injection for multi-tenancy is allowed: e.g. `namespace=foo`
NetObserv is an observability solution dedicated to the network, that stores flow logs in Loki. Implementing multi-tenancy for NetObserv requires injecting more complex matchers for handling flow logs sources and destinations: e.g.
`src-namespace=foo OR dst-namespace=foo` (as a single query injection).

Changes:

- Updated result data from OPA Authz (**OPA breaking change**)
- Query injection needs to handle several matchers, and logical OR
- Fix string escaping issue in Loki reencoder
- Add a few tests
- As an aside, allow building with podman

Related PRs:
- https://github.com/observatorium/opa-openshift/pull/15
- https://github.com/grafana/loki/pull/8192

### DIFF
--- a/api/logs/v1/labels_enforcer.go
+++ b/api/logs/v1/labels_enforcer.go
@@ -14,7 +14,7 @@ import (
 
 type AuthzResponseData struct {
 	Matchers  []*labels.Matcher `json:"matchers,omitempty"`
-	LogicalOp string            `json:"logicalOp,omitempty"`
+	MatcherOp string            `json:"matcherOp,omitempty"`
 }
 
 const logicalOr = "or"
@@ -71,7 +71,7 @@ func enforceValues(mInfo AuthzResponseData, v url.Values) (values string, err er
 		return "", fmt.Errorf("failed parsing LogQL expression: %w", err)
 	}
 
-	switch mInfo.LogicalOp {
+	switch mInfo.MatcherOp {
 	case logicalOr:
 		// Logical "OR" to combine multiple matchers needs to be done via LogQueryExpr > LogPipelineExpr
 		expr.Walk(func(expr interface{}) {
--- a/api/logs/v1/labels_enforcer.go
+++ b/api/logs/v1/labels_enforcer.go
@@ -71,18 +71,27 @@ func enforceValues(mInfo AuthzResponseData, v url.Values) (values string, err er
 		return "", fmt.Errorf("failed parsing LogQL expression: %w", err)
 	}
 
-	expr.Walk(func(expr interface{}) {
-		switch le := expr.(type) {
-		case *logqlv2.StreamMatcherExpr:
-			if mInfo.LogicalOp == logicalOr {
-				le.AppendORMatchers(mInfo.Matchers)
-			} else {
-				le.AppendMatchers(mInfo.Matchers)
+	switch mInfo.LogicalOp {
+	case logicalOr:
+		// Logical "OR" to combine multiple matchers needs to be done via LogQueryExpr > LogPipelineExpr
+		expr.Walk(func(expr interface{}) {
+			switch le := expr.(type) {
+			case *logqlv2.LogQueryExpr:
+				le.AppendPipelineMatchers(mInfo.Matchers, logicalOr)
+			default:
+				// Do nothing
 			}
-		default:
-			// Do nothing
-		}
-	})
+		})
+	default:
+		expr.Walk(func(expr interface{}) {
+			switch le := expr.(type) {
+			case *logqlv2.StreamMatcherExpr:
+				le.AppendMatchers(mInfo.Matchers)
+			default:
+				// Do nothing
+			}
+		})
+	}
 
 	v.Set(queryParam, expr.String())
 
--- a/api/logs/v1/labels_enforcer.go
+++ b/api/logs/v1/labels_enforcer.go
@@ -12,6 +12,14 @@ import (
 	"github.com/prometheus/prometheus/model/labels"
 )
 
+// TODO: share with OPA?
+type MatchersInfo struct {
+	Matchers  []*labels.Matcher `json:"matchers,omitempty"`
+	LogicalOp string            `json:"logicalOp,omitempty"`
+}
+
+const logicalOr = "or"
+
 // WithEnforceAuthorizationLabels return a middleware that ensures every query
 // has a set of labels returned by the OPA authorizer enforced.
 func WithEnforceAuthorizationLabels() func(http.Handler) http.Handler {
@@ -32,14 +40,14 @@ func WithEnforceAuthorizationLabels() func(http.Handler) http.Handler {
 				return
 			}
 
-			var lm []*labels.Matcher
-			if err := json.Unmarshal([]byte(data), &lm); err != nil {
+			var matchersInfo MatchersInfo
+			if err := json.Unmarshal([]byte(data), &matchersInfo); err != nil {
 				httperr.PrometheusAPIError(w, "error parsing authorization label matchers", http.StatusInternalServerError)
 
 				return
 			}
 
-			q, err := enforceValues(lm, r.URL.Query())
+			q, err := enforceValues(matchersInfo, r.URL.Query())
 			if err != nil {
 				httperr.PrometheusAPIError(w, fmt.Sprintf("could not enforce authorization label matchers: %v", err), http.StatusInternalServerError)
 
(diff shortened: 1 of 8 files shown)


api/logs/v1/labels_enforcer.go
```diff
@@ -67,7 +75,11 @@ func enforceValues(lm []*labels.Matcher, v url.Values) (values string, err error
 	expr.Walk(func(expr interface{}) {
 		switch le := expr.(type) {
 		case *logqlv2.StreamMatcherExpr:
```
periklis, line 77: FWIW considering the observatorium/opa-openshift#15 draft PR it looks like the `opa.matcher` (considering we can only do `vector or vector` with stream label matcher in LogQL) is overloaded. The `label.Matchers` should maybe include a field that tells if they are going to be appended as label selectors in `StreamMatcherExpr` or as line filters., e.g.:
```go
	expr.Walk(func(expr interface{}) {
		switch le := expr.(type) {
		case *logqlv2.StreamMatcherExpr:
			le.AppendMatchers(lm)
		case *logqlv2.LineFilterExpr: // I believe our syntax does not support this case yet
			if mInfo.LogicalOp == logicalOr {
				le.AppendORMatchers(mInfo.Matchers)
			} else {
				le.AppendMatchers(mInfo.Matchers)
			}
		default:
			// Do nothing
		}
	})
```

jotak: yes, or we do line filters anyway, regardless if it's vectors or not.

logql/v2/ast_test.go
```diff
@@ -136,3 +136,78 @@ func Test_AstWalker_AppendMatcher(t *testing.T) {
 		}
 	}
 }
+
+func Test_AstWalker_AppendORMatcher(t *testing.T) {
```
jotak, line 140: FYI this test passes, and the produced LogQL is valid wrt upstream Loki

logql/v2/parser_test.go
```diff
@@ -1392,6 +1392,44 @@ func TestParseExpr(t *testing.T) {
 				},
 			},
 		},
+		// OR Log filters
```
jotak, line 1395: this test fails:

```
unexpected err: syntax error: unexpected IDENTIFIER: 1:35
```

The "or" operator doesn't seem to be recognized

jotak:
@periklis this one is ready for review

logql/v2/parser_test.go
jotak: (fixed)

api/logs/v1/labels_enforcer.go
periklis: Looking on the `StreamMatchersExpr.String()` implementation below I see that you circuit break with a closing curly brace in case of an or-Matcher and move forward with a line filter. This looks very to me. In the above sample I propose to clearly render or-Matchers as line filters, e.g. by appending a new line filter with or-s at the end of the query.

PS: I mention Line Filter above, but I am not sure if it the right expression. IIRC logical expressions on matchers are supported only on label filters and not line filters. Please consider the parser tests for the right expression.

periklis (changes requested):
In general the approach looks good to me. However I suggest we keep rendering rendering stream matchers independently from label filters  with logical expressions.

api/logs/v1/labels_enforcer.go
jotak: I tried your suggested approach at first but I hit an issue: if I remember correctly, Line filters "walker" is only invoked if there are actually line filters in the initial expression, it isn't invoked at all if there are none. Or maybe I didn't proceed like what you were thinking about... I can retry and let you see the  issue if you want

jotak: I think the approach you suggest would work but requires a little more refactoring of the traversal pattern

logql/v2/ast.go
```diff
@@ -61,17 +61,15 @@ func (s *StreamMatcherExpr) Walk(fn WalkFn) {
 func (s *StreamMatcherExpr) String() string {
 	var sb strings.Builder
 
-	sb.WriteString("{")
-
+	sb.WriteRune('{')
```
periklis, line 64: What is the value of this? I am just curious.

logql/v2/ast.go
```diff
@@ -114,9 +112,7 @@ func (l *LogFilterExpr) String() string {
 		sb.WriteString(`"`)
 		sb.WriteString(")")
 	} else {
-		sb.WriteString(`"`)
-		sb.WriteString(l.value)
-		sb.WriteString(`"`)
+		sb.WriteString(strconv.Quote(l.value))
```
periklis, line 115: We can apply the same improvement to L110-L112, right?

logql/v2/parser_test.go
```diff
@@ -1477,3 +1528,13 @@ func trimOutput(s string) string {
...
+
+func TestQuotesEncode(t *testing.T) {
+	expr, err := ParseExpr(`{app="test"}`)
+	require.NoError(t, err)
+	require.Equal(t, `{app="test"}`, expr.String())
+
+	expr, err = ParseExpr("{app=\"test\"}|~`key\":\"val\"`")
+	require.NoError(t, err)
+	require.Equal(t, `{app="test"} |~ "key\":\"val\""`, expr.String())
+}
```
periklis, line 1540: Can we eliminate testify usage here? This repo is under shared maintenance with many teams and we don't have any vote on changing the way we write tests.

logql/v2/parser_test.go
```diff
@@ -451,6 +482,26 @@ func TestParseExpr(t *testing.T) {
...
+					matchers: []*labels.Matcher{
+						{
+							Type:  labels.MatchEqual,
+							Name:  "first",
+							Value: "value",
+						},
+					},
+				},
+			},
+		},
```
periklis, line 504: What is the value added by this test case not covered by any of the existing?

logql/v2/parser_test.go
```diff
@@ -385,6 +386,36 @@ func TestParseExpr(t *testing.T) {
...
```
periklis, line 418: Isn't this test case already covered by the next one? Or is this all about testing the `strconv.Quote`? If yes, on the last question then we can drop this test as per `strconv.Quote` having enough tests in the Golang std library.

periklis (commented):
In general looks good to me. I will add my approve once I have tested this in companion with observatorium-api on our Loki Operator release branches.

logql/v2/ast.go
jotak: I assume it's supposed to be more optimized but I admin I have no certitude I could not find a clear statement on that while googling, and doing a quick benchmark isn't conclusive. I can revert then ...

logql/v2/ast.go
jotak: I guess so .. just been conservative to make sure nothing unexpected could break; but doing the same on L110-112, all tests still pass, so I can do that as well

logql/v2/parser_test.go
jotak: I think I wrote this test because it was the closest I had from the actual netobserv queries with tenancy matchers. So it helped me while doing the implementation. But you're right that it overlaps with other tests. I can remove it.

logql/v2/parser_test.go
jotak: It's indeed similar to `TestQuotesEncode` below except this one tests double-quotes whereas the other one tests backquotes. I can move it to the `TestQuotesEncode` test, maybe it will be more clear & consistent

logql/v2/parser_test.go
jotak: btw: I tend to write flat tests rather than test cases arrays (aka table-driven tests), that's a personal preference, but since the norm here seems to be test cases arrays, I can definitely comply with it.

logql/v2/ast.go
periklis: Let's keep it the old way

logql/v2/ast.go
periklis: Yes please let's use for both case `strconv.Quote` for consistency.

periklis (commented):
Still looks good to me

PR #447, opened 2023-01-31
Update k8s dependencies to latest (v0.26.1)

Red-GV:
This PR updates the k8s dependencies. In order to fix the breaking changes from this, the `prometheus`, `prometheus-community`, and `certificate-transparency-go` libraries also had to be updated.


api/metrics/v1/labels_enforcer.go
```diff
@@ -34,7 +34,7 @@ func WithEnforceTenancyOnQuery(label string) func(http.Handler) http.Handler {
 				return
 			}
 
-			e := injectproxy.NewEnforcer([]*labels.Matcher{{
+			e := injectproxy.NewEnforcer(false, []*labels.Matcher{{
```
periklis, line 37: What is this new parameter set to `false` here? Can you elaborate a bit.

authentication/openshift/options.go
```diff
@@ -138,7 +138,7 @@ func (s *DelegatingAuthenticationOptions) getRequestHeader() (*RequestHeaderAuth
 	return nil, fmt.Errorf("no request header config")
 }
 
-func (s *DelegatingAuthenticationOptions) newTokenAccessReview() (authenticationclient.TokenReviewInterface, error) {
+func (s *DelegatingAuthenticationOptions) newAuthenticationClient() (authenticationclient.AuthenticationV1Interface, error) {
```
periklis, line 141: Can you link a k8s PR that explains why we switch the interface here?

go.mod
```diff
@@ -26,10 +26,10 @@ require (
 	github.com/oklog/run v1.1.0
 	github.com/open-policy-agent/opa v0.39.0
 	github.com/pkg/errors v0.9.1
-	github.com/prometheus-community/prom-label-proxy v0.3.1-0.20210623095334-9d425172d7bb
+	github.com/prometheus-community/prom-label-proxy v0.6.0
```
periklis, line 29: Anything else we need to be aware by jumping so many versions? Anything particular that your eye catched scanning the changelog?

authentication/openshift/options.go
Red-GV: v0.21.1: https://pkg.go.dev/k8s.io/apiserver@v0.21.1/pkg/authentication/authenticatorfactory#DelegatingAuthenticatorConfig

v0.26.1: https://pkg.go.dev/k8s.io/apiserver/pkg/authentication/authenticatorfactory#DelegatingAuthenticatorConfig

api/metrics/v1/labels_enforcer.go
Red-GV: This is parameter is for `errorOnReplace` to allow the enforcer to return an error when [`EnforceMatchers`](https://pkg.go.dev/github.com/prometheus-community/prom-label-proxy@v0.4.0/injectproxy#Enforcer.EnforceMatchers) is called. When false, this method would not report a method. However - as far as I could tell - we do not use this method as of yet.

go.mod
Red-GV: These ones kinda popped out from the [changelog](https://github.com/prometheus-community/prom-label-proxy/blob/main/CHANGELOG.md):

```
v0.4.0
- [ENHANCEMENT] Support HTTP POST for /api/v1/labels endpoint.
- [ENHANCEMENT] Add label enforce support for the new query_exemplars API.

v0.5.0
- [ENHANCEMENT] Add /healthz endpoint for (Kubernetes) probes.

v0.6.0
- [FEATURE] Add the --header-name flag to pass the label value via HTTP header.
- [FEATURE] Add the --internal-listen-address flag to expose Prometheus metrics. 
```

authentication/openshift/options.go
periklis: :+1:

go.mod
periklis: :+1:

douglascamata:
@Red-GV some e2e tests are failing, could you have a look at them, please?

api/metrics/v1/labels_enforcer.go
douglascamata: Does this preserve backwards compatibility on how this function behaves? 🤔

Red-GV:
@douglascamata They passed yesterday before I did a rebase to rewrite some commit messages. Is there a way to just re-run that one?

douglascamata:
@Red-GV I think the only way is to push an empty commit: `git commit --allow-empty -s -m "Rerun CI"`

api/metrics/v1/labels_enforcer.go
Red-GV: As far as I know. It seems like the default behavior for this is `false`. However, if I'm not mistaken, this was added in v0.4.0 as a new feature.

Red-GV:
@douglascamata All the tests have passed.

api/metrics/v1/labels_enforcer.go
saswatamcode: Hmm, we use `EnforceNode` which indirectly calls `EnforceMatchers`.
So `errOnReplace=false` basically allows us to silently replace the matcher in a PromQL expression (either in query or in rules), if it already has the matcher we are enforcing, but with a different value. 

For e.g if we get a query `http_requests_total{tenant_id="foo"}`, and we enforce `tenant_id=bar`, the resultant query getting processed would be `http_requests_total{tenant_id="foo"}`

This was the default before so we haven't changed anything! 🙂

saswatamcode (approved):
LGTM!

go.mod
simonpasquier: Feel free to tag me in the future as I'm maintaining prom-label-proxy. Just to confirm: the bump should be safe.

PR #463, opened 2023-02-20
Fix rules-raw handler instrumentation

saswatamcode:
This fixes the rules-raw handler instrumentation. Missed from #430 review.


PhilipGough (approved):
lgtm

matej-g (approved):

PR #481, opened 2023-03-17
logsv1: Enfore authz matcher on label values

periklis:
**What this PR does / why we need it**:
The current state of the logsv1 API enforces only authorization matchers on instant and range queries. This allows to limit users to search within a tenant by rewriting the LogQL stream selectors with label matchers provided by the authorization process (e.g. limit to accessible kubernetes namespaces). However the users are still able to access all label values within a tenant (e.g. all namespace names) although they have via RBAC access only to a limited set. This issue can be characterized simply as information intrusion w/o breaking the tenancy model per se.

The following PR adds support to apply authorization label matchers to the Loki Label Values API as effective filters. It applies the label matchers to the query parameter provided by:
- https://github.com/grafana/loki/pull/8824

In particular the Loki Labels/Label-Values API works in two ways:
- For queries answered by the ingester (live unflushed data) the active streams are queried in memory for labels keys or label values for a given label key. The additional authorization label matchers are used as a second pass filter to eliminate streams that don't match.
- For queries on historical data the querier will execute queries only on the index files. Authorization label matchers are used in an and expression to k-set operation against each index.

**Special notes for your reviewer**:

Container images for testing:
- Loki: v2.8.0
- Observatorium-API: [quay.io/periklis/observatorium-api:authz-matcher-on-labels-values-2023-03-17-v0.1.2-355-g4151849d](quay.io/periklis/observatorium-api:authz-matcher-on-labels-values-2023-03-17-v0.1.2-355-g4151849d)

cc @xperimental

### DIFF
--- a/api/logs/v1/labels_enforcer.go
+++ b/api/logs/v1/labels_enforcer.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strings"
 
 	"github.com/observatorium/api/authorization"
 	"github.com/observatorium/api/httperr"
@@ -46,7 +47,7 @@ func WithEnforceAuthorizationLabels() func(http.Handler) http.Handler {
 				return
 			}
 
-			q, err := enforceValues(matchersInfo, r.URL.Query())
+			q, err := enforceValues(matchersInfo, r.URL)
 			if err != nil {
 				httperr.PrometheusAPIError(w, fmt.Sprintf("could not enforce authorization label matchers: %v", err), http.StatusInternalServerError)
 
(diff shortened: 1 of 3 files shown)


api/logs/v1/labels_enforcer.go
```diff
@@ -61,20 +62,56 @@ func WithEnforceAuthorizationLabels() func(http.Handler) http.Handler {
...
+		case *logqlv2.LogQueryExpr:
+			matchers := combineLabelMatchers(le.Matchers(), lm)
+			le.SetMatchers(matchers)
+		default:
+			// Do nothing
+		}
+	default:
+		expr = &logqlv2.StreamMatcherExpr{}
+		expr.(*logqlv2.StreamMatcherExpr).SetMatchers(lm)
+	}
```
aminesnow, line 100: An alternative way to avoid nested switch cases:
```golang
        var expr logqlv2.Expr = &logqlv2.StreamMatcherExpr{}
	expr.(*logqlv2.StreamMatcherExpr).SetMatchers(lm)

	q := v.Get(queryParam)
	if q != "" {
		expr, err = logqlv2.ParseExpr(q)
		if err != nil {
			return "", fmt.Errorf("failed parsing LogQL expression: %w", err)
		}

		switch le := expr.(type) {
		case *logqlv2.LogQueryExpr:
			matchers := combineLabelMatchers(le.Matchers(), lm)
			le.SetMatchers(matchers)
		default:
			// Do nothing
		}
	}
```

api/logs/v1/labels_enforcer.go
```diff
@@ -131,3 +164,17 @@ func combineLabelMatchers(queryMatchers, authzMatchers []*labels.Matcher) []*lab
 
 	return matchers
 }
+
+func initAuthzMatchers(lm []*labels.Matcher) ([]*labels.Matcher, error) {
```
xperimental, line 168: Took me a while to understand what this function is doing, why it's recreating the same type. It seems like the answer is that this type was not meant to be serialized to a JSON representation and so is not properly initialized when de-serialized.

This is not really something for this PR, but I think introducing a custom type that makes it clear that the deserialized type can not actually be used as a Matcher should be cleaner.

xperimental (commented):
Tested it on my test-cluster, seems to work fine with Grafana.

Most of my comments are "optional", the only thing I really want you to take into consideration is the one about "if to emit an error in the do-nothing-case".

api/logs/v1/labels_enforcer.go
```diff
@@ -61,20 +62,52 @@ func WithEnforceAuthorizationLabels() func(http.Handler) http.Handler {
...
 		}
 
-		lm[i] = nm
+		switch le := expr.(type) {
+		case *logqlv2.LogQueryExpr:
+			matchers := combineLabelMatchers(le.Matchers(), lm)
+			le.SetMatchers(matchers)
+		default:
+			// Do nothing
+		}
```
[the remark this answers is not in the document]
periklis: I agree with the suggestion. Regarding the default case it was there to highlight that you can have literal expression in Loki like `1+1` which will never match any logs.

api/logs/v1/labels_enforcer.go
```diff
@@ -61,20 +62,52 @@ func WithEnforceAuthorizationLabels() func(http.Handler) http.Handler {
...
+	default:
+		return enforceValuesOnQuery(mInfo, u.Query())
 	}
+}
 
-	lm := mInfo.Matchers
-	// Fix label matchers to include a non nil FastRegexMatcher for regex types.
-	for i, m := range lm {
-		nm, err := labels.NewMatcher(m.Type, m.Name, m.Value)
+func enforceValuesOnLabelValues(mInfo AuthzResponseData, v url.Values) (values string, err error) {
```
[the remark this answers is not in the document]
periklis: Keeping `enforceValues` to stay naming-wise with the other signals.

api/logs/v1/labels_enforcer.go
periklis: I agree but this requires refining the initial proposal (see ) and rework maybe opa-openshift and opa-ams sidecars:
https://github.com/observatorium/observatorium/blob/f174afa24b856b51b3575d45d8797503f3f22072/docs/proposals/drafts/20210709-opa-authorizer-labels-observatorium-api.md

PR #493, opened 2023-04-04
Split request timeouts per signal handler

periklis:
### Problem statement
Previously the per signal handlers used arbitrary chosen timeout values to express how long a proxied request to the corresponding upstream should take. However due to conceptual misunderstandings (e.g. request read timeout set as dialer timeout) or global middleware configuration (e.g. `middleware.Timeout(2 * time.Minute)`) the effective timeout was set to **2 Minutes** (configured in the `http.Server.WriteTimeout`). 

In general the current behavior treats all signal handlers equally and in turn unfairly. For example range queries to Loki can in general take longer than those to Thanos, because processing logs (e.g. parsing, filtering, aggregating) might take longer. I

### Proposed solution
The following implementation splits the timeouts per signal handler group (logs, metrics, traces) and keeps for sanity some global server timeouts to prevent misuse by faulty HTTP clients. I.e.:
1. The global HTTP header timeout is set to `1 second`
2. The global HTTP request read timeout (incl. processing the HTTP request body) is set to `5 seconds`
3. The global HTTP request write timeout (incl. writing the HTTP response body) is set to `12 minutes` as per slowest signal handler is the range query handler for Loki (See below)

For each signal group a request context timeout is set and passed to the upstream HTTP client as below:
1. **Metrics**: `2 minutes` (Note: Keeping the old value and leaving this for further investigation by the metrics API maintainers cc @douglascamata @PhilipGough @saswatamcode)
2. **Logs**: `10 minutes`
3. **Traces**: `2 minutes` (Note: Keeping the old value and leaving this for further investigation by the traces API maintainers cc @pavolloffay) 

#### Further notes
The previously called `readTimeout` in each handler implementation used as Dialer Timeout is renamed as `dialTimeout` and set to `30 seconds` for every signal group (logs, metrics, traces). This value seems good enough as it is usually set when using the `http.DefaultTransport` used by `http.DefaultClient` for terminating TCP connects.

### Open Questions

- [ ]  Should we enable users to configure the timeouts per CLI arguments? (AFAICS this was previously possible but got removed)
- [ ] Do we have enough logging in place for request context timeouts? This needs further investigation to improve debugging slow queries.

### DIFF
--- a/api/logs/v1/http.go
+++ b/api/logs/v1/http.go
@@ -19,8 +19,7 @@ import (
 )
 
 const (
-	ReadTimeout  = 15 * time.Minute
-	WriteTimeout = time.Minute
+	dialTimeout = 30 * time.Second // Set as in http.DefaultTransport
 )
 
 const (
@@ -163,7 +162,7 @@ func NewHandler(read, tail, write, rules *url.URL, rulesReadOnly bool, upstreamC
 
 			t := &http.Transport{
 				DialContext: (&net.Dialer{
-					Timeout: ReadTimeout,
+					Timeout: dialTimeout,
 				}).DialContext,
 				TLSClientConfig: tls.NewClientConfig(upstreamCA, upstreamCert),
 			}
@@ -231,7 +230,7 @@ func NewHandler(read, tail, write, rules *url.URL, rulesReadOnly bool, upstreamC
 
 			t := &http.Transport{
 				DialContext: (&net.Dialer{
-					Timeout: ReadTimeout,
+					Timeout: dialTimeout,
 				}).DialContext,
 				TLSClientConfig: tls.NewClientConfig(upstreamCA, upstreamCert),
 			}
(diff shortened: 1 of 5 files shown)


saswatamcode (approved):
Thanks a lot for taking care of this! That changes make sense to me overall!

>  Should we enable users to configure the timeouts per CLI arguments? (AFAICS this was previously possible but got removed)

I think we can have flags for this, with the defaults set in there. That would allow users to override it if needed. But I'm fine either way. 🙂 

> Do we have enough logging in place for request context timeouts? This needs further investigation to improve debugging slow queries.

I think having logs for slow query/writes in the API layer itself, wouldn't be too beneficial. We would need to reference logs of the relevant upstream to figure out _why_ a certain request was slow anyway. 

Maybe we can have a metric to count # of timeouts instead, and then alert on that?

PhilipGough (approved):

douglascamata (commented):
Late to the party, but I love this. Thanks!

PR #505, opened 2023-04-19
Add namespace label filter when fetching alerting rules

aminesnow:
This PR adds a middleware function that adds `namespace` label filter when querying alerting rules. This filter contains the list of namespaces that the user has access to, thus returning only the rules that are in those namespaces.

This PR depends on [grafana/loki/pull/9128](https://github.com/grafana/loki/pull/9128)


periklis:
Replaced by #578

PR #511, opened 2023-05-05
Expose HTTP timeout options as flags

periklis:
Follow-up PR for #493 to expose timeout configuration per CLI flags.

Ref: [LOG-4044](https://issues.redhat.com/browse/LOG-4044)

cc @xperimental


philipgough (commented):
lgtm

PR #512, opened 2023-05-05
NETOBSERV-1020 Fix empty namespace list

jotak:
cf https://github.com/observatorium/opa-openshift/pull/18

We need to prevent attempts to run queries when the matchers list exists but is empty, which corresponds to having 0 allowed namespaces

### DIFF
--- a/api/logs/v1/labels_enforcer.go
+++ b/api/logs/v1/labels_enforcer.go
@@ -46,6 +46,12 @@ func WithEnforceAuthorizationLabels() func(http.Handler) http.Handler {
 				return
 			}
 
+			if len(matchersInfo.Matchers) == 0 {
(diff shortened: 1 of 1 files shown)


api/logs/v1/labels_enforcer.go
```diff
@@ -46,6 +46,12 @@ func WithEnforceAuthorizationLabels() func(http.Handler) http.Handler {
 				return
 			}
 
+			if len(matchersInfo.Matchers) == 0 {
+				httperr.PrometheusAPIError(w, "empty authorization label matchers", http.StatusInternalServerError)
```
jotak, line 50: We might do something different than returning such an error. Ideally, we should not run any loki query but directly return an empty result set. (I need to check how to do that...)

jotak:
closing: amended the `opa-openshift` PR to make it self-contained, so there's no need to change something on the API side.

PR #532, opened 2023-05-30
logs: Per tenant label-based rule filters

periklis:
This PR introduces the capability to filter log-based rules by applying user-selected labels. The administrator can define which label keys are valid via a CLI flag `--logs.rules.label-filters=tenantA:label-a,label-b;tenantB;label-c` and can be used as URL query parameters. The enforcement applies on the prometheus-compatible `/prometheus/api/v1/rules` as well as on the Loki API `/loki/api/v1/rules`.

cc @xperimental @aminesnow @jgbernalp

### DIFF
--- a/api/logs/v1/http.go
+++ b/api/logs/v1/http.go
@@ -54,6 +54,7 @@ type handlerConfiguration struct {
 	registry              *prometheus.Registry
 	instrument            handlerInstrumenter
 	spanRoutePrefix       string
+	rulesLabelFilters     map[string][]string
 	readMiddlewares       []func(http.Handler) http.Handler
 	writeMiddlewares      []func(http.Handler) http.Handler
 	rulesReadMiddlewares  []func(http.Handler) http.Handler
@@ -88,6 +89,13 @@ func WithHandlerInstrumenter(instrumenter handlerInstrumenter) HandlerOption {
 func WithSpanRoutePrefix(spanRoutePrefix string) HandlerOption {
 	return func(h *handlerConfiguration) {
 		h.spanRoutePrefix = spanRoutePrefix
+	}
+}
+
+// WithSpanRoutePrefix adds a prefix before the value of route tag in tracing spans.
+func WithRulesLabelFilters(f map[string][]string) HandlerOption {
+	return func(h *handlerConfiguration) {
+		h.rulesLabelFilters = f
 	}
 }
 
@@ -219,7 +227,7 @@ func NewHandler(read, tail, write, rules *url.URL, rulesReadOnly bool, upstreamC
 	}
 
 	if rules != nil {
-		var proxyRules http.Handler
+		var proxyReadRules, proxyWriteRules http.Handler
 		{
 			middlewares := proxy.Middlewares(
 				proxy.MiddlewareSetUpstream(rules),
@@ -235,7 +243,14 @@ func NewHandler(read, tail, write, rules *url.URL, rulesReadOnly bool, upstreamC
 				TLSClientConfig: tls.NewClientConfig(upstreamCA, upstreamCert),
 			}
 
-			proxyRules = &httputil.ReverseProxy{
+			proxyReadRules = &httputil.ReverseProxy{
+				Director:       middlewares,
+				ErrorLog:       proxy.Logger(c.logger),
+				Transport:      otelhttp.NewTransport(t),
+				ModifyResponse: newModifyResponse(c.logger, c.rulesLabelFilters),
+			}
+
+			proxyWriteRules = &httputil.ReverseProxy{
 				Director:  middlewares,
 				ErrorLog:  proxy.Logger(c.logger),
 				Transport: otelhttp.NewTransport(t),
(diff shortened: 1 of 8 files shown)


api/logs/v1/rules_labels_enforcer.go
```diff
@@ -0,0 +1,179 @@
...
+				Name:     rg.Name,
+				Interval: rg.Interval,
+				Limit:    rg.Limit,
+				Rules:    filteredRules,
+			})
+		}
+	}
+
+	return rulesData{RuleGroups: filtered}
+}
```
aminesnow, line 161: Here's an alternative implementation that's easier to read:
```suggestion
func filterPrometheusRules(rd rulesData, matchers map[string]string) rulesData {
	if len(matchers) == 0 {
		return rd
	}

	var filtered []rulefmt.RuleGroup

	for _, rg := range rd.RuleGroups {
		var filteredRules []rulefmt.RuleNode

		for _, rule := range rg.Rules {
			if ruleLabelsMatcheMatchers(rule, matchers) {
				filteredRules = append(filteredRules, rule)
			}
		}

		if len(filteredRules) > 0 {
			filtered = append(filtered, rulefmt.RuleGroup{
				Name:     rg.Name,
				Interval: rg.Interval,
				Limit:    rg.Limit,
				Rules:    filteredRules,
			})
		}
	}

	return rulesData{RuleGroups: filtered}
}

func ruleLabelsMatcheMatchers(rule rulefmt.RuleNode, matchers map[string]string) bool {
	for key, value := range matchers {
		val, ok := rule.Labels[key]
		if !ok || val != value {
			return false
		}
	}
	return true
}
```

api/logs/v1/rules_labels_enforcer_test.go
```diff
@@ -0,0 +1,369 @@
...
+					{
+						Name: "group-a",
+						Rules: []rulefmt.RuleNode{
+							{Labels: map[string]string{"other": "not"}},
+						},
+					},
+				},
+			},
+			want: rulesData{
+				RuleGroups: []rulefmt.RuleGroup{},
```
aminesnow, line 245: I think this should be nil.

api/logs/v1/rules_labels_enforcer.go
```diff
@@ -0,0 +1,179 @@
...
+			return nil, err
+		}
+
+		res.Data = filterPrometheusRules(res.Data, matchers)
+
+		return json.Marshal(res)
+
+	case "application/yaml":
+		var res lokiRulesResponse
+		err := yaml.Unmarshal(body, &res)
```
aminesnow, line 116: I don't think this works. You are using "github.com/ghodss/yaml" to unmarshal a Go object that is meant for "gopkg.in/yaml.v2". Have you tested this

aminesnow (changes requested):
Looks to be on a good track for the workaround. I have raised some issues in the comments.
Have you tested this with the Loki Operator change in this [PR](https://github.com/grafana/loki/pull/9196)?

api/logs/v1/rules_labels_enforcer.go
```diff
@@ -0,0 +1,179 @@
...
+	body []byte
+}
+
+func (rlf *responseLabelFilter) Write(b []byte) (int, error) {
+	rlf.body = b
+
+	return rlf.ResponseWriter.Write(b)
+}
+
+// WithEnforeFilterLabels returns a middleware that ensure that all returned
```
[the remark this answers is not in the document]
periklis: Replaced by `newModifyResponse`

api/logs/v1/http.go
```diff
@@ -91,6 +92,13 @@ func WithSpanRoutePrefix(spanRoutePrefix string) HandlerOption {
 	}
 }
 
+// WithSpanRoutePrefix adds a prefix before the value of route tag in tracing spans.
```
aminesnow, line 95: Copy pasta

api/logs/v1/rules_labels_enforcer.go
```diff
@@ -0,0 +1,401 @@
...
+	"github.com/prometheus/prometheus/model/labels"
+)
+
+const (
+	contentTypeApplicationJSON = "application/json"
+	contentTypeApplicationYAML = "application/yaml"
+)
+
+var (
+	errUnknownTenantKey        = errors.New("Uknown tenant key")
```
aminesnow, line 27: Typo:
```suggestion
	errUnknownTenantKey        = errors.New("Unknown tenant key")
```

api/logs/v1/rules_labels_enforcer.go
```diff
@@ -0,0 +1,410 @@
...
+				next.ServeHTTP(w, r)
+
+				return
+			}
+
+			// Check that the user provides each label filter as a non-empty URL parameter
+			queryParams := r.URL.Query()
+			for _, key := range keys {
+				if !queryParams.Has(key) {
+					httperr.PrometheusAPIError(w, fmt.Sprintf("missing URL parameter %s", key), http.StatusBadRequest)
```
aminesnow, line 168: If I understand correctly, this makes it mandatory to use all the defined labels. I don't think this is a good idea. 
I think we should be able to define a set of labels to filter on, and if some or none of those labels are provided, then we should still be able to query.

api/logs/v1/rules_labels_enforcer.go
```diff
@@ -0,0 +1,410 @@
...
+		}
+
+		if len(filteredRules) > 0 {
+			group.Rules = filteredRules
+			filtered = append(filtered, group)
+		}
+	}
+
+	return filtered
+}
```
aminesnow, line 354: Here's an alternative way of filtering without repeating the code and using the loop labels.
You will just have to implement the interface for all 3 types.
```suggestion
type labeledRule interface {
	GetLabels() labels.Labels
}

func ruleLabelsMatcheMatchers(rule labeledRule, matchers map[string]string) bool {
	for key, value := range matchers {
		labels := rule.GetLabels().Map()
		val, ok := labels[key]
		if !ok || val != value {
			return false
		}
	}
	return true
}

func filterPrometheusRuleGroups(groups []*ruleGroup, matchers map[string]string) []*ruleGroup {
	var filtered []*ruleGroup

	for _, group := range groups {
		var filteredRules []rule
		for _, rule := range group.Rules {
			if ruleLabelsMatcheMatchers(rule, matchers) {
				filteredRules = append(filteredRules, rule)
			}
		}

		if len(filteredRules) > 0 {
			group.Rules = filteredRules
			filtered = append(filtered, group)
		}
	}

	return filtered
}
```

api/logs/v1/rules_labels_enforcer.go
```diff
@@ -0,0 +1,410 @@
...
+			if !alert.Labels.Has(key) || alert.Labels.Get(key) != value {
+				continue alerts
+			}
+		}
+
+		filtered = append(filtered, alert)
+	}
+
+	return filtered
+}
```
aminesnow, line 371: Following the suggestion above:
```suggestion
func filterPrometheusAlerts(alerts []*alert, matchers map[string]string) []*alert {
	var filtered []*alert
	for _, alert := range alerts {
		if ruleLabelsMatcheMatchers(alert, matchers) {
			filtered = append(filtered, alert)
		}
	}

	return filtered
}
```

api/logs/v1/rules_labels_enforcer.go
```diff
@@ -0,0 +1,410 @@
...
+			}
+		}
+
+		if len(filteredGroups) > 0 {
+			filtered[name] = filteredGroups
+		}
+	}
+
+	return filtered
+}
```
aminesnow, line 410: Following the suggestion above:
```suggestion

func filterLokiRules(res lokiRulesResponse, matchers map[string]string) lokiRulesResponse {
	if len(matchers) == 0 {
		return nil
	}

	filtered := lokiRulesResponse{}

	for name, groups := range res {
		var filteredGroups []lokiRuleGroup

		for _, group := range groups {
			var filteredRules []lokiRule
			for _, rule := range group.Rules {
				if ruleLabelsMatcheMatchers(rule, matchers) {
					filteredRules = append(filteredRules, rule)
				}
			}

			if len(filteredRules) > 0 {
				group.Rules = filteredRules
				filteredGroups = append(filteredGroups, group)
			}
		}

		if len(filteredGroups) > 0 {
			filtered[name] = filteredGroups
		}
	}

	return filtered
}
```

aminesnow (changes requested):
Just a few comments, but looks good overall, great job with the modifyResponse idea!
I did test it and it seems functional. I will be reviewing the unit tests as well as the related LO PR later.

api/logs/v1/rules_labels_enforcer.go
periklis: I like to think this the other way around. They `keys` are provided by the administrator via `--logs.rules.label-filters` to ensure that users only search for logs giving mandatory input to these `keys` so that our middleware can enforce authorization constraints for the given input.

aminesnow: Thing is, they are not really provided by the administrator. This something the Operator defines in our case. 

In terms of API usage, if we set for instance two keys A and B, we no longer have the ability to use the API without these two filters, and we are also forced to use both A and B filters in order to get a response.
This in my opinion makes the API too restrictive, because even as a cluster admin I cannot query without all of these per-defined filters.

periklis: As agreed on our last call, I am writing the following down for the record. The behavior we actually want is twofold:
- For users with authorization matchers returned from the authz endpoint, we should return an empty list if the user is not providing any rule label filters per URL query parameters.
- For users without authorization matchers, we should return the original response if the user is not providing any rule label filters per URL query parameters.

PR #539, opened 2023-06-15
Put back podman compatibility and fix multi-arch builds

douglascamata:
1. Multi-arch image building process only works with `docker` because we rely on `docker buildx`.
2. Put back the single-arch container build make target that's compatible with `podman`.
3. Fix `container-test` make target to be OS-independent.


Makefile
```diff
@@ -171,23 +171,32 @@ proto: ratelimit/gubernator/proto/google ratelimit/gubernator/gubernator.proto $
...
+		--build-arg VCS_BRANCH="$(VCS_BRANCH)" \
+		--build-arg DOCKERFILE_PATH="/Dockerfile.e2e-test" \
+		-t $(DOCKER_REPO):local_e2e_test  \
 		-f Dockerfile.e2e-test \
-		-t $(DOCKER_REPO):local_e2e_test .
-else
-container-test:
-	@docker build \
-		-f Dockerfile \
-		-t $(DOCKER_REPO):local_e2e_test .
```
douglascamata, line 184: This would never have worked in anything that's not Linux.

saswatamcode (approved):
💪🏻

PR #547, opened 2023-07-28
Initialize all handlers metrics

douglascamata:
This PR initializes all the metrics from the **metrics** APIs (v1 and legacy) to ensure they are present on a restart of the application, even without traffic.

We were having issues with the Pyrra SLOs firing the `SLOMetricAbsent` in these scenarios. This should prevent these alerts.

A small refactor was done to stop trying to guess the group and handler based on the URL path at request time. Instead, these are added through handler wrappers that store the group and handler in the context. The list of groups and handlers are used to initialize the metrics.

A bigger refactor can be done to improve this further, making the registration of the group and handler names of an endpoint initialize the metric. This will require many more changes, including changing interfaces in the code, so I didn't do it in this PR.


philipgough (approved):
lgtm

saswatamcode (approved):
LGTM!

PR #549, opened 2023-07-28
Improve jaeger ui path replacement to make it more robust

rubenvp8510:
Altough this solutionn  still feels hacky

Usign a regexp instead of a simple string replacement make the solution a little bit more robust. This make it works even if spaces are injected on the middle of the tags
Eg.
```
<base href="%s/" data-inject-target="BASE_URL"/>
<base href="%s/" data-inject-target="BASE_URL" />
<   base href="%s/"           data-inject-target="BASE_URL"/>
```


squat (commented):
hi @rubenvp8510 what is more robust about this change? is it that you allow for more than one whitespace between attributes?

rubenvp8510:
Have no idea why e2e is failing :(

squat:
it seems like an unrelated metrics test failed. re-trying the failed now :))

squat (approved):

PR #554, opened 2023-08-09
Fix concurrent map access in HTTP instrumentation

aminesnow:
The Labels map that is used for instrumentation is being read and wrote simultaneously by multiple goroutines, resulting in a panic.
This PR removes the need to use the same variable across different routines.

### DIFF
--- a/server/http_instrumentation.go
+++ b/server/http_instrumentation.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"net/http"
 	"strconv"
-	"sync"
 	"time"
 
 	"github.com/go-chi/chi/middleware"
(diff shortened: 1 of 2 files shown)


server/http_instrumentation.go
```diff
@@ -85,15 +83,18 @@ func (m instrumentedHandlerFactory) InitializeMetrics(labels prometheus.Labels)
...
+	// Default group and handler to "unknown" if no extra labels are provided as a parameter.
+	if extraLabels == nil {
+		extraLabels = prometheus.Labels{"group": "unknown", "handler": "unknown"}
+	}
 
-		// Default group and handler to "unknown" if no extra labels are provided as a parameter.
-		if extraLabels == nil {
-			extraLabels = prometheus.Labels{"group": "unknown", "handler": "unknown"}
+	return func(w http.ResponseWriter, r *http.Request) {
+		extraLabelsCopy := make(prometheus.Labels, len(extraLabels))
```
xperimental, line 92: What about a more "speaking name"?

```suggestion
		requestLabels := make(prometheus.Labels, len(extraLabels))
```

xperimental (approved):

PR #555, opened 2023-08-10
Extend logs read endpoint to extract selectors from query enabling more fine-grained access control

shwetaap:
This PR in combination with other changes in opa and loki, makes sure that user requests will now additionally be authorized via namespaced SubjectAccessReviews


authorization/query.go
```diff
@@ -0,0 +1,80 @@
...
+	if query == "" {
+		return []string{}, nil
+	}
+
+	namespaces, err := parseQueryNamespaces(query)
+	if err != nil {
+		return nil, err
+	}
+
+	return namespaces, nil
```
JoaoBraveCoding, line 31:
```suggestion
	return parseQueryNamespaces(query)
```

authorization/query.go
```diff
@@ -0,0 +1,80 @@
...
+							continue
+						}
+
+						namespaces = append(namespaces, v)
+					}
+				}
+
+				if m.Type != labels.MatchEqual {
+					continue
+				}
```
JoaoBraveCoding, line 67: I don't quite get the purpose of this part, since it's at the end of the for loop

authorization/query_test.go
```diff
@@ -0,0 +1,47 @@
...
+	for _, tt := range tests {
+		t.Run(tt.query, func(t *testing.T) {
+			got, err := parseQueryNamespaces(tt.query)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("parseQueryNamespaces() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("parseQueryNamespaces() got = %v, want %v", got, tt.want)
+			}
```
JoaoBraveCoding, line 44: nit: I just like the `require` package too much
```suggestion
			if tt.wantErr {
			   require.Error(t, err)
			   return
			 }
			require.NoError(t, err)
			require.Equal(t, tt.want, got)
```

rbac/rbac_test.go
```diff
@@ -1018,7 +1018,7 @@ func TestNewAuthorizer(t *testing.T) {
 			a := NewAuthorizer(tc.roles, tc.roleBindings, logger.NewLogger("info", logger.LogFormatLogfmt, "observatorium"))
 			for i := range tc.ios {
 				sc, out, data := a.Authorize(tc.ios[i].subject, tc.ios[i].groups, tc.ios[i].permission, tc.ios[i].resource,
```
JoaoBraveCoding, line 1020: Shouldn't we expand the test cases to cover these new variants?

JoaoBraveCoding (commented):
Initial review, no manual testing done yet

authorization/http.go
```diff
@@ -31,6 +46,49 @@ func WithData(ctx context.Context, data string) context.Context {
...
+			}
+
+			selectorsInfo, err := extractQuerySelectors(selectorLabelMap, r.URL.Query())
+			if err != nil {
+				httperr.PrometheusAPIError(w, fmt.Sprintf("error extracting query selectors: %s", err), http.StatusInternalServerError)
+
+				return
+			}
+
+			next.ServeHTTP(w, r.WithContext(WithSelectorsInfo(r.Context(), selectorsInfo)))
```
periklis, line 87: Since this package is signal unaware, I suggest to add another parameter `resource string` to this func (See `authorization.WithAuthorizers`) and use it for a switch case for signals supporting query selector extraction.

main.go
```diff
@@ -1048,6 +1051,8 @@ func parseFlags() (config, error) {
 		"The name of the rules label that should hold the tenant ID in logs upstreams.")
 	flag.StringVar(&rawLogsWriteEndpoint, "logs.write.endpoint", "",
 		"The endpoint against which to make write requests for logs.")
+	flag.StringVar(&rawLogsExtractSelectorLabels, "logs.extract-selector-labels", "",
+		"Comma-separated list of query labels that should be treated as namespaces for authorization.")
```
periklis, line 1055: Similar to comment in `README.md`

rbac/rbac.go
```diff
@@ -49,10 +49,22 @@ type RoleBinding struct {
...
+type ExtraAttributes struct {
+	Logs *LogsExtraAttributes `json:"logs,omitempty"`
+}
+
+// LogsExtraAttributes contains extra attributes about a logs request.
+type LogsExtraAttributes struct {
+	Selectors         map[string][]string `json:"selectors,omitempty"`
+	WildcardSelectors bool                `json:"wildcardSelectors,omitempty"`
+	MetadataOnly      bool                `json:"metadataOnly,omitempty"`
+}
```
periklis, line 62: Similar as mentioned in the opa-openshift repo: https://github.com/observatorium/opa-openshift/pull/20#discussion_r1322528010

authorization/query_test.go
xperimental: No other test in this repo uses testify, so I don't think it's a good idea to add it now.

authorization/http.go
xperimental: I don't get this one. Shouldn't it be "signal agnostic" at this point? That's why we removed the `logs` sub-struct from the input-document. In my mind only the "extraction" is signal-dependent (because it needs to know which query language to parse), but the part sending it to OPA can be signal agnostic.

periklis (commented):
LGTM

periklis (approved):

PR #556, opened 2023-08-15
proxy subset of alertmanager routes

philipgough:
This change proxies a subset of alertmanager routes that can enforce tenancy via labels matchers and filters. The middleware functions are adapted from https://github.com/prometheus-community/prom-label-proxy.

This allows tenants to list their alerts and silences as well as create silences for their alerts. 

As a follow up to this PR, we can consider reading alerts and silences by {id}, deleting and updating silences by {id}. That change will require us to embed an alertmanager client that reads resources on the way out to validate tenancy. I did not include that here since it would be a fundamental change in behaviour for this reverse proxy and wanted to discuss it.


Makefile
```diff
@@ -124,6 +124,7 @@ lint: $(FAILLINT) $(GOLANGCI_LINT) $(MISSPELL) generate format deps build check-
 	@go vet -stdmethods=false ./...
 	@echo ">> linting all of the Go files GOGC=${GOGC}"
 	@$(GOLANGCI_LINT) run
+	@$(GOLANGCI_LINT) run
```
saswatamcode, line 127: Double run?

api/metrics/v1/http.go
```diff
@@ -34,18 +34,28 @@ const (
 	ReceiveRoute     = "/api/v1/receive"
 	RulesRoute       = "/api/v1/rules"
 	RulesRawRoute    = "/api/v1/rules/raw"
+
+	AlertmanagerAlertsRoute   = "/api/v2/alerts"
```
saswatamcode, line 38: I'm not sure about the naming of this endpoint. Thanos Query also has an `/api/v1/alerts` API I believe, which makes such an endpoint kind of confusing for user (v1 to query, v2 to am).

Maybe we can attach an `am` prefix? So `am/api/v2/alerts`?

main.go
```diff
@@ -670,7 +676,20 @@ func main() {
 						metricsv1.WithReadMiddleware(metricsv1.WithEnforceTenancyOnMatchers(cfg.metrics.tenantLabel)),
 						metricsv1.WithReadMiddleware(metricsv1.WithEnforceAuthorizationLabels()),
 						metricsv1.WithUIMiddleware(authorization.WithAuthorizers(authorizers, rbac.Read, "metrics")),
-					))
+						metricsv1.WithAlertmanagerAlertsReadMiddleware(
```
saswatamcode, line 679: Nice! Let's also refactor the other middleware functions to be variadic then, for consistency? 🙂

api/metrics/v1/http.go
```diff
@@ -114,13 +124,34 @@ func WithWriteMiddleware(m func(http.Handler) http.Handler) HandlerOption {
 	}
 }
 
+func WithAlertmanagerAlertsReadMiddleware(m ...func(http.Handler) http.Handler) HandlerOption {
```
saswatamcode, line 127: As we can pass multiple
```suggestion
func WithAlertmanagerAlertsReadMiddlewares(m ...func(http.Handler) http.Handler) HandlerOption {
```

api/metrics/v1/alertmanager_enforcer.go
```diff
@@ -0,0 +1,130 @@
...
+				for _, filter := range filters {
+					m, err := amlabels.ParseMatcher(filter)
+					if err != nil {
+						return
+					}
+					// Keep the original matcher in case of multi label values because
+					// the user might want to filter on a specific value.
+					if m.Name == label {
+						continue
+					}
```
saswatamcode, line 54: What's the behaviour in case we get a request with a `tenant` label, that has a different value? We just don't process it and default to tenant one?

saswatamcode (dismissed):
Thanks! Generally LGTM mod some comments! 🙂

api/metrics/v1/http.go
philipgough: yes I think that makes sense

api/metrics/v1/alertmanager_enforcer.go
philipgough: yup, thats the intention, otherwise they could impersonate right?

main.go
philipgough: yes will do, but I think it makes morse sense in a follow up PR?

api/metrics/v1/http.go
philipgough: Right, I kept the naming convention because Global already accepts multiple. I kind of think it still makes sense even without the plural for variadic arguments, wdyt?

api/metrics/v1/alertmanager_enforcer.go
saswatamcode: Ack!

main.go
saswatamcode: Sure!

api/metrics/v1/http.go
saswatamcode: Ah I see! Fine to skip, this is a tiny nitpick anyway

saswatamcode (approved):
LGTM!

PR #563, opened 2023-09-08
Add e2e test for pushing empty list of rules

douglascamata:
This should match the behavior from rules-objstore, which is being tested in https://github.com/observatorium/rules-objstore/pull/15.

### DIFF
--- a/test/e2e/rules_test.go
+++ b/test/e2e/rules_test.go
@@ -101,6 +101,10 @@ groups:
         annotations: {}
 `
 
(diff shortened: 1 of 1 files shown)


saswatamcode (approved):
LGTM! Maybe adding a fully empty case might also make sense?

douglascamata:
@saswatamcode done, sir.

saswatamcode (approved):
LGTM!

PR #570, opened 2023-09-12
Expose Jaeger metrics API

pavolloffay:


squat:
What's the motivation for this? The APIs exposed by Observatiorium should ideally be a kind of contract that could be fulfilled by any Jaeger-esque backend. What if you put Observatiorium in front of a hosted Jaeger? You won't be able to get metrics from that service, but that should be a valid deployment IMO. In other words, why should Jaeger ever be scraped _through_ Observatiorium and why should we commit to exposing metrics for our backends (imagine we one day swap out single-node Jaeger for multiple replicas or microservices, whose metrics do we expose then? Does Observatiorium need to become an exporter for that backend that knows how to aggregate across multiple replicas?)?

AFAIK we don't expose the metrics for Thanos or Loki components, so it seems funny decide to do it for Jaeger.

pavolloffay:
> In other words, why should Jaeger ever be scraped through Observatiorium and why should we commit to exposing metrics for our backends

@squat let me explain what metrics are actually exposed as your assumptions do not seem correct. 

This PR exposes metrics API (`/api/metrics`) of the Jaeger query service. The API is used by the Jaeger UI Monitor tab (or service performance monitoring tab) https://www.jaegertracing.io/docs/1.49/spm/. These metrics are derived from traces reported by instrumented workloads (applications instrumented with trace APIs). 

The intention here is not not expose Jaeger internal metrics or expose this endpoint to be scraped. These metrics are one of the data points that Jaeger UI uses to help users understand their applications.

squat:
Ah excellent! That is indeed completely different from what I assumed!

frzifus (approved):
So no further concerns? LGTM :1st_place_medal:

test/e2e/services.go
```diff
@@ -95,14 +96,33 @@ func startServicesForLogs(t *testing.T, e e2e.Environment) (
...
+			map[string]int{
+				"http": 9090,
+			}).
+		Init(e2e.StartOptions{Image: prometheusImage})
+	testutil.Ok(t, e2e.StartAndWaitReady(prometheus))
+
+	// The endpoint points to localhost
+	// parse the port and use it with host IP address later
+	promHttpEndpoint := prometheus.Endpoint("http")
+	promHostPort := promHttpEndpoint[strings.Index(promHttpEndpoint, ":")+1:]
```
saswatamcode, line 110: We can probably replace all this with `e2edb.NewPrometheus()` and you can use the port later with `.InternalEndpoint("http")`

saswatamcode (approved):
LGTM! Just one small nit

test/e2e/services.go
pavolloffay: +1 changed

saswatamcode (approved):

PR #575, opened 2023-09-20
Expose tempo trace query API using the gateway

rubenvp8510:
This PR introduces a new flag `-traces.tempo.endpoint` which should point to the **Tempo** _Query-frontend_ component in order to expose the tempo query API. (https://grafana.com/docs/tempo/latest/api_docs/) This will allow us to use the TraceQL API with the gateway.

The endpoints will be mapped according to this design document: https://docs.google.com/document/d/1vQ8F9MI8VhnqSyKjBom87axY7OIhDXeTJ4JGa9aDf9s

@pavolloffay  please review

### DIFF
--- a/api/traces/v1/http.go
+++ b/api/traces/v1/http.go
@@ -126,6 +126,8 @@ func NewV2Handler(read *url.URL, readTemplate string, tempo *url.URL, upstreamCA
 
 	r := chi.NewRouter()
 
+	middlewareMetrics := proxy.MiddlewareMetrics(c.registry, prometheus.Labels{"proxy": "tracesv1-read"})
+
 	if read != nil || readTemplate != "" {
 
 		var proxyRead http.Handler
@@ -143,7 +145,7 @@ func NewV2Handler(read *url.URL, readTemplate string, tempo *url.URL, upstreamCA
 				upstreamMiddleware,
 				proxy.MiddlewareSetPrefixHeader(),
 				proxy.MiddlewareLogger(c.logger),
-				proxy.MiddlewareMetrics(c.registry, prometheus.Labels{"proxy": "tracesv1-read"}),
+				middlewareMetrics,
 			)
 
 			t := &http.Transport{
@@ -191,6 +193,7 @@ func NewV2Handler(read *url.URL, readTemplate string, tempo *url.URL, upstreamCA
 
 	// if tempo upstream is enabled, configure proxy and route
 	if tempo != nil {
+		level.Debug(c.logger).Log("msg", "Configuring upstream Tempo", "queryv2", read)
 
 		t := &http.Transport{
 			DialContext: (&net.Dialer{
@@ -205,7 +208,7 @@ func NewV2Handler(read *url.URL, readTemplate string, tempo *url.URL, upstreamCA
 			proxy.MiddlewareSetUpstream(tempo),
 			proxy.MiddlewareSetTempoPrefixHeader(),
 			proxy.MiddlewareLogger(c.logger),
-			proxy.MiddlewareMetrics(c.registry, prometheus.Labels{"proxy": "tracesv1-read"}),
+			middlewareMetrics,
 		)
 
 		tempoProxyRead := &httputil.ReverseProxy{
--- a/api/traces/v1/http.go
+++ b/api/traces/v1/http.go
@@ -218,7 +218,7 @@ func NewV2Handler(read *url.URL, readTemplate string, tempo *url.URL, upstreamCA
 
 		r.Group(func(r chi.Router) {
 			r.Use(c.tempoMiddlewares...)
-			r.Get("/tempo*", c.instrument.NewHandler(
+			r.Get("/tempo/api*", c.instrument.NewHandler(
 				prometheus.Labels{"group": "tracesv1tempo", "handler": "tempo"},
 				tempoProxyRead))
 		})
--- a/api/traces/v1/http.go
+++ b/api/traces/v1/http.go
@@ -84,7 +84,7 @@ func WithReadMiddleware(m func(http.Handler) http.Handler) HandlerOption {
 // WithTempoMiddleware adds a middleware for all tempo read operations.
 func WithTempoMiddleware(m func(http.Handler) http.Handler) HandlerOption {
 	return func(h *handlerConfiguration) {
-		h.readMiddlewares = append(h.readMiddlewares, m)
+		h.tempoMiddlewares = append(h.tempoMiddlewares, m)
 	}
 }
 
--- a/api/traces/v1/http.go
+++ b/api/traces/v1/http.go
@@ -205,7 +205,8 @@ func NewV2Handler(read *url.URL, readTemplate string, tempo *url.URL, upstreamCA
 		middlewares := proxy.Middlewares(
 			proxy.MiddlewareRemoveURLPrefix("tempo"),
 			proxy.MiddlewareSetUpstream(tempo),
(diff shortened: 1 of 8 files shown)


pavolloffay:
I would like to see some e2e tests in https://github.com/observatorium/api/blob/de7fd756e311dc41adeeee7c18b184f9048b5424/test/e2e/traces_test.go for this feature

pavolloffay:
Did you test this with Grafana?

rubenvp8510:
> I would like to see some e2e tests in https://github.com/observatorium/api/blob/de7fd756e311dc41adeeee7c18b184f9048b5424/test/e2e/traces_test.go for this feature

Got it, will add tests

rubenvp8510:
> Did you test this with Grafana?

Not yet, I just tested that I was able to reach the endpoints, I'll do a quick test today.

api/traces/v1/http.go
```diff
@@ -118,23 +126,74 @@ func NewV2Handler(read *url.URL, readTemplate string, upstreamCA []byte, upstrea
...
+				proxyRead))
+			r.Get("/static/*", c.instrument.NewHandler(
+				prometheus.Labels{"group": "tracesv1static", "handler": "ui"},
+				proxyRead))
+			r.Get("/search*", c.instrument.NewHandler(
+				prometheus.Labels{"group": "tracesv1ui", "handler": "ui"},
+				proxyRead))
+			r.Get("/favicon.ico", c.instrument.NewHandler(
+				prometheus.Labels{"group": "tracesv1ui", "handler": "ui"},
+				proxyRead))
```
saswatamcode, line 190: Probably we can combine the UI handlers into one via `r.Mount` with `/` as the pattern, similar to how we do here: https://github.com/observatorium/api/blob/main/api/metrics/v1/http.go#L324?

proxy/proxy.go
```diff
@@ -55,6 +71,22 @@ func MiddlewareSetPrefixHeader() Middleware {
...
 
+func MiddlewareSetTempoPrefixHeader() Middleware {
+	return func(r *http.Request) {
+		prefix, ok := getPrefix(r.Context())
+		if !ok {
+			return
+		}
+
+		// Do not override the prefix header if it is already set.
+		if r.Header.Get(PrefixHeader) != "" {
```
andreasgerstmayr, line 82: Can this be a security issue?

So the gateway performs authn and authz on the user, and adds the correct `X-Scope-OrgID` header.
And with the above code, any successful authenticated user can impersonate any other user?

proxy/proxy.go
```diff
@@ -55,6 +71,22 @@ func MiddlewareSetPrefixHeader() Middleware {
...
+			return
+		}
+
+		r.Header.Set(PrefixHeader, prefix)
```
andreasgerstmayr, line 86: Should this be `TempoPrefixHeader`? I can't find any usage of the new `TempoPrefixHeader` constant.

proxy/proxy.go
andreasgerstmayr: Maybe I got confused with `PrefixHeader` vs. `TempoPrefixHeader` :confused:

proxy/proxy.go
```diff
@@ -19,6 +21,8 @@ const (
 	prefixKey contextKey = "prefix"
 
 	PrefixHeader string = "X-Forwarded-Prefix"
+
+	TempoPrefixHeader string = "X-Scope-OrgID"
```
andreasgerstmayr, line 25: Should this value come from `cfg.traces.tenantHeader` instead? In tempo-operator, we're starting the gateway with `--traces.tenant-header=x-scope-orgid`

rubenvp8510: My intention was to forward the tenant to tempo in the header that tempo expects. which is defined here: https://grafana.com/docs/tempo/latest/operations/multitenancy/ 


> empo is a multi-tenant distributed tracing backend. It supports multi-tenancy through the use of a header: X-Scope-OrgID.

proxy/proxy.go
rubenvp8510: I fixed it, seems you're right.

api/traces/v1/http.go
rubenvp8510: One question, if I put the r`.Mount("/"..)` here, Will this take precedence over the other above which is `"/tempo"`? because I wanted that "/tempo" will have a different one, forwarding the request to a different endpoint.

rubenvp8510: Did a quick test and was not able to make it work in that way, even if I put the tempo router first.

proxy/proxy.go
andreasgerstmayr: I wonder why `cfg.traces.tenantHeader` is configurable in the first place then, if it's always a constant set to `X-Scope-OrgID`?

rubenvp8510:
@pavolloffay added tests.

test/e2e/services.go
```diff
@@ -28,6 +28,8 @@ const (
 
 	jaegerAllInOneImage = "jaegertracing/all-in-one:1.31"
 	otelCollectorImage  = "otel/opentelemetry-collector:0.45.0"
+	tempoImage          = "grafana/tempo:latest"
```
frzifus, line 31: Maybe we should pin this image too. That way it can not break the tests.

frzifus (commented):
Are we creating `traces.tempo.endpoint` instead of reusing `traces.read.endpoint` to allow the usage of jaeger & grafana at the same time?

saswatamcode (approved):
Thanks! LGTM, as far as I understand this.

saswatamcode (approved):

iblancasa (approved):

test/e2e/configs.go
```diff
@@ -415,3 +417,58 @@ func createLokiYAML(
...
+      path: /tmp/tempo/blocks
+`
+
+// createTempoConfigYAML() creates YAML for Tempo inside the Observatorium API boundary.
+func createTempoConfigYAML(
+	t *testing.T,
+	e e2e.Environment,
+) {
+	// Warn if a YAML change introduced a tab character
+	if strings.ContainsRune(otelConfigTpl, '\t') {
```
pavolloffay, line 462: Why is it checking the `otelConfigTpl`?

pavolloffay (commented):
LGTM just one nit

test/e2e/configs.go
rubenvp8510: Oops, my mistake

saswatamcode (dismissed):

iblancasa (approved):

pavolloffay (approved):

main.go
```diff
@@ -1263,6 +1270,17 @@ func parseFlags() (config, error) {
 		cfg.traces.readEndpoint = tracesReadEndpoint
 	}
 
+	if rawTracesTempoEndpoint != "" {
+		cfg.traces.enabled = true
+
+		tracesTempoEndpoint, err := url.ParseRequestURI(rawTracesTempoEndpoint)
+		if err != nil {
+			return cfg, fmt.Errorf("--traces.read.endpoint %q is invalid: %w", rawTracesReadEndpoint, err)
```
andreasgerstmayr, line 1278:
```suggestion
			return cfg, fmt.Errorf("--traces.tempo.endpoint %q is invalid: %w", rawTracesReadEndpoint, err)
```

andreasgerstmayr (approved):

philipgough (approved):

rubenvp8510:
@philipgough @saswatamcode Is something else required to merge this PR?

PR #595, opened 2023-10-25
Fix enforcement of query matchers for `series`, `label_values`, and `label_names` calls

douglascamata:
While investigating an issue where a tenant was seeing other tenant's label values/names in Grafana I uncovered that it was making a `POST` request to the `series` API with form data. 

Observatorium API was then adding a **query parameter** `match[]` to lock the call to the tenant's metrics. When the request is proxied down to Thanos, it uses `Request.ParseForm()`. This helper merges query params and form data params into a single map, giving priority to form data and ignoring the tenant filter that was added to the query parameter.

I also noticed that we had two helpers that were enforcing matchers:

* A more generic, called `WithEnforceTenancyOnQuery`, which was used to enforce matchers configured via OPA **and** tenancy in calls that use the `query` parameter (`query` and `query_range` calls).
* A more specific, called `WithEnforceTenancyOnMatchers`, which was used to enforce tenancy on queries in the `match[]` parameter (`series`, `label_names`, and `label_values` callls).

So... I got rid of `WithEnforceTenancyOnMatchers` altogether and updated `WithEnforceTenancyOnQuery` to be able to enforce matchers on a configurable param. This way it can be used for both `query` and `path` params where each one of them are needed.


squat:
@douglascamata this is an issue we've been aware of for quite a long time in the prom-label-proxy project. Our solutions to it sometimes cause user complaints (and patches) [0] due to compatibility with some downstreams. Observatorium has it relatively easier since it knows exactly who its downstream is and doesn't need to worry about compatibility. Ideally I would like to see us solve this problem ONCE in the prom-label-proxy project and have Observatorium vendor the enforcement code as a library. WDYT?

[0] https://github.com/prometheus-community/prom-label-proxy/pull/111

douglascamata:
@squat I am open to trying to solve it up there on `prom-label-proxy`, but after I fix this. This is kind of important for me now and I predict the upstream solution to take some time to be merged. I'm finishing to tidy this up to better reuse code and I'm happy to discuss upstream.

squat:
ACK thanks @douglascamata

api/metrics/v1/labels_enforcer.go
```diff
@@ -180,33 +137,35 @@ func enforceRequestQueryLabels(e *injectproxy.Enforcer, w http.ResponseWriter, r
...
-
-	expr, err := parser.ParseExpr(v.Get(queryParam))
-	if err != nil {
-		return "", true, fmt.Errorf("parse expr error: %w", err)
+func enforceQueryValues(e *injectproxy.Enforcer, paramName string, requestParams url.Values) (values string, foundQuery bool, err error) {
+	if len(requestParams[paramName]) == 0 {
+		enforcedMatchers, err := e.EnforceMatchers([]*labels.Matcher{})
+		if err != nil {
+			return "", false, fmt.Errorf("enforce matchers error: %w", err)
+		}
```
douglascamata, line 145: This is a dirty hack to force the introduction of a `match[]` param to add tenancy enforcement even when the param isn't present. This is needed because `match[]` is optional for label names and values, yet we have to add it to avoid leaks.

Would love to find a cleaner way to do it.

api/metrics/v1/labels_enforcer.go
```diff
@@ -180,33 +132,35 @@ func enforceRequestQueryLabels(e *injectproxy.Enforcer, w http.ResponseWriter, r
...
+		if err := e.EnforceNode(expr); err != nil {
+			return "", true, fmt.Errorf("enforce node error: %w", err)
+		}
+		matchers[i] = expr.String()
 	}
-
-	v.Set(queryParam, expr.String())
-
-	return v.Encode(), true, nil
+	requestParams[paramName] = matchers
```
thibaultmg, line 156: Is this needed?

douglascamata: Good catch, @thibaultmg. I pushed a272b8ee5651db5c73814b5d1e7016af2f6e5c36 to get rid of it after verifying in a playground (https://go.dev/play/p/NxPYe__Y7Py) it's not required.

api/metrics/v1/labels_enforcer.go
thibaultmg: Sounds good to me. Maybe might be worth adding this as a comment?

thibaultmg (approved):
LGTM!

api/metrics/v1/labels_enforcer.go
douglascamata: Done it [f10aa57](https://github.com/observatorium/api/pull/595/commits/f10aa57d6bfe1a5aa0562ec88581f33f54e35458).

saswatamcode (approved):
Thanks for the great work! Generally LGTM!

main.go
```diff
@@ -672,9 +672,9 @@ func main() {
 						metricsv1.WithGlobalMiddleware(metricsMiddlewares...),
 						metricsv1.WithWriteMiddleware(authorization.WithAuthorizers(authorizers, rbac.Write, "metrics")),
 						metricsv1.WithQueryMiddleware(authorization.WithAuthorizers(authorizers, rbac.Read, "metrics")),
-						metricsv1.WithQueryMiddleware(metricsv1.WithEnforceTenancyOnQuery(cfg.metrics.tenantLabel)),
+						metricsv1.WithQueryMiddleware(metricsv1.WithEnforceTenancyOnQuery(cfg.metrics.tenantLabel, "query")),
 						metricsv1.WithReadMiddleware(authorization.WithAuthorizers(authorizers, rbac.Read, "metrics")),
-						metricsv1.WithReadMiddleware(metricsv1.WithEnforceTenancyOnMatchers(cfg.metrics.tenantLabel)),
+						metricsv1.WithReadMiddleware(metricsv1.WithEnforceTenancyOnQuery(cfg.metrics.tenantLabel, "match[]")),
```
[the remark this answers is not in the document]
douglascamata: Why though? These strings aren't even duplicated or used anywhere else. 

We have many other duplicated strings that are used in many places and they aren't constants.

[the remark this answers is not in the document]
douglascamata: Ok, `"query"` is duplicated once. But I don't see how it makes things more difficult to reason about. 

Anyway, fixed in b56e222fe563bfd0a20f5b76bdb2e4c8e79390bb.

saswatamcode (approved):

PR #597, opened 2023-11-01
Ensure correct labels values are added during ratelimiting

coleenquadros:
https://issues.redhat.com/browse/RHOBS-914

Currently, Obs API does not have good per-tenant instrumentation to visualize requests that get rejected early (due to rate limits or other cases)

The issue is that when the rate limit middleware or other middlewares return on error, the context does not hold information about the group or handler as the labels are set when the route is handled at the end. So when the control returns back to the instrumentation handler ([line101](https://github.com/observatorium/api/blob/56abe030cbbe3e50a614dacf793fa673317b937b/server/http_instrumentation.go#L101)) the context does not hold the label values and will be reported as `unknown`. 

With this PR we ensure that the labels are set with the group and handler in context before the middlewares run and the instrumentation handler can pick up the values

### DIFF
--- a/api/metrics/legacy/http.go
+++ b/api/metrics/legacy/http.go
@@ -141,23 +141,32 @@ func NewHandler(url *url.URL, upstreamCA []byte, upstreamCert *stdtls.Certificat
 	}
 
 	r.Group(func(r chi.Router) {
+		r.Use(func(handler http.Handler) http.Handler {
+			return server.InjectLabelsCtx(
+				prometheus.Labels{"group": "metricslegacy", "handler": "query"},
+				handler,
+			)
+		})
 		r.Use(c.queryMiddlewares...)
 		r.Handle(QueryRoute,
 			otelhttp.WithRouteTag(
 				c.spanRoutePrefix+QueryRoute,
-				server.InjectLabelsCtx(
-					prometheus.Labels{"group": "metricslegacy", "handler": "query"},
-					legacyProxy,
-				),
+				legacyProxy,
 			),
 		)
+	})
+	r.Group(func(r chi.Router) {
+		r.Use(func(handler http.Handler) http.Handler {
+			return server.InjectLabelsCtx(
+				prometheus.Labels{"group": "metricslegacy", "handler": "query_range"},
+				handler,
+			)
+		})
+		r.Use(c.queryMiddlewares...)
 		r.Handle(QueryRangeRoute,
 			otelhttp.WithRouteTag(
 				c.spanRoutePrefix+QueryRangeRoute,
-				server.InjectLabelsCtx(
-					prometheus.Labels{"group": "metricslegacy", "handler": "query_range"},
-					legacyProxy,
-				),
+				legacyProxy,
 			),
 		)
 	})
--- a/api/metrics/v1/http.go
+++ b/api/metrics/v1/http.go
@@ -213,24 +213,34 @@ func NewHandler(endpoints Endpoints, upstreamCA []byte, upstreamCert *stdtls.Cer
 			}
 		}
 		r.Group(func(r chi.Router) {
+			r.Use(func(handler http.Handler) http.Handler {
+				return server.InjectLabelsCtx(
+					prometheus.Labels{"group": "metricsv1", "handler": "query"},
+					handler,
+				)
+			})
 			r.Use(c.queryMiddlewares...)
 			r.Use(server.StripTenantPrefix("/api/metrics/v1"))
 			r.Handle(QueryRoute,
 				otelhttp.WithRouteTag(
 					c.spanRoutePrefix+QueryRoute,
-					server.InjectLabelsCtx(
-						prometheus.Labels{"group": "metricsv1", "handler": "query"},
-						proxyQuery,
-					),
+					proxyQuery,
 				),
 			)
+		})
+		r.Group(func(r chi.Router) {
+			r.Use(func(handler http.Handler) http.Handler {
+				return server.InjectLabelsCtx(
+					prometheus.Labels{"group": "metricsv1", "handler": "query_range"},
+					handler,
+				)
+			})
+			r.Use(c.queryMiddlewares...)
+			r.Use(server.StripTenantPrefix("/api/metrics/v1"))
 			r.Handle(QueryRangeRoute,
 				otelhttp.WithRouteTag(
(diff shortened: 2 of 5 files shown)


ratelimit/http_test.go
```diff
@@ -107,14 +110,20 @@ func TestWithLocalRateLimiter(t *testing.T) {
 
 	for _, c := range cases {
 		t.Run(c.name, func(t *testing.T) {
-			rlmw := WithLocalRateLimiter(c.configs...)
```
coleenquadros, line 110: I have modified the test file to replicate the order in which the middlewares are executed in metrics/v1/http.go

api/metrics/v1/http.go
```diff
@@ -213,24 +213,34 @@ func NewHandler(endpoints Endpoints, upstreamCA []byte, upstreamCert *stdtls.Cer
 			}
 		}
 		r.Group(func(r chi.Router) {
+			r.Use(func(handler http.Handler) http.Handler {
```
coleenquadros, line 216: The instrumentation middleware is run first (line 189), then the next middleware injects the extra labels `group` and `handler` into context so that when the middleware for ratelimiter runs and when a 429 is returned back to the instrumentation handler(here -> [line101](https://github.com/observatorium/api/blob/56abe030cbbe3e50a614dacf793fa673317b937b/server/http_instrumentation.go#L101)), the ctx already has the labels for `group` and `handler` set

coleenquadros:
Ran it using interactive tests - 

```
18:49:25 observatorium-api: **_Coleen labels and status code map[group:metricsv1 handler:ui] , 404_**
18:49:25 observatorium-api: level=debug name=observatorium ts=2023-11-07T17:49:24.861747017Z caller=instrumentation.go:36 request=observatorium-api/TTNGjUQ7N0-000302 proto=HTTP/2.0 method=POST status=404 content=application/x-www-form-urlencoded path=/api/metrics/v1/test-oidc/api/v1/query/api/v1/query duration=1.111417ms bytes=19
18:49:25 up-metrics-read-write: caller=level.go:63 ts=2023-11-07T17:49:24.86224035Z name=up level=error caller=main.go:153 component=reader msg="failed to query" err="query request failed: client_error: client error: 404"
18:49:25 observatorium-api: **_Coleen labels and status code map[group:metricsv1 handler:receive] , 429_**
18:49:25 observatorium-api: level=debug name=observatorium ts=2023-11-07T17:49:24.862880642Z caller=instrumentation.go:36 request=observatorium-api/TTNGjUQ7N0-000303 proto=HTTP/2.0 method=POST status=429 content= path=/api/metrics/v1/test-oidc/api/v1/receive duration=1.77825ms bytes=79
18:49:25 up-metrics-read-write: caller=level.go:63 ts=2023-11-07T17:49:24.8635241Z name=up level=error caller=main.go:119 component=writer msg="failed to make request" err="non-200 status: 429 Too Many Requests"
18:49:25 observatorium-api: level=debug name=observatorium ts=2023-11-07T17:49:25.03181735Z caller=proxy.go:61 request=observatorium-api/TTNGjUQ7N0-000304 msg="request to upstream" url=http://interactive-loki:3100/loki/api/v1/push
18:49:25 observatorium-api: **_Coleen labels and status code map[group:logsv1 handler:push] , 204_**
18:49:25 observatorium-api: level=debug name=observatorium ts=2023-11-07T17:49:25.0334681Z caller=instrumentation.go:36 request=observatorium-api/TTNGjUQ7N0-000304 proto=HTTP/2.0 method=POST status=204 content=application/json path=/api/logs/v1/test-oidc/loki/api/v1/push duration=2.226209ms bytes=0
18:49:25 observatorium-api: **_Coleen labels and status code map[group:metricsv1 handler:receive] , 429_**
```

api/metrics/legacy/http.go
```diff
@@ -141,23 +141,32 @@ func NewHandler(url *url.URL, upstreamCA []byte, upstreamCert *stdtls.Certificat
 	}
 
 	r.Group(func(r chi.Router) {
+		r.Use(func(handler http.Handler) http.Handler {
```
coleenquadros, line 144: The instrumentation middleware is run first (line 189), then the next middleware injects the extra labels `group` and `handler` into context so that when the middleware for ratelimiter runs and when a 429 is returned back to the instrumentation handler(here -> [line101](https://github.com/observatorium/api/blob/56abe030cbbe3e50a614dacf793fa673317b937b/server/http_instrumentation.go#L101)), the ctx already has the labels for `group` and `handler` set

douglascamata: There is a design problem in injecting these labels here: the handler is not decided yet. 

For example, we could have the following: 

```go
		r.Group(func(r chi.Router) {
			r.Use(func(handler http.Handler) http.Handler {
				return server.InjectLabelsCtx(
					prometheus.Labels{"group": "metricsv1", "handler": "query"},
					handler,
				)
			})
			r.Use(c.queryMiddlewares...)
			r.Use(server.StripTenantPrefix("/api/metrics/v1"))
			r.Handle(QueryRoute,
				otelhttp.WithRouteTag(
					c.spanRoutePrefix+QueryRoute,
					proxyQuery,
				),
			)
            // A different query route -- maybe a different type of query besides range and instant. Or simply a new range query endpoint with a slightly different middleware stack.
			r.Handle(DifferentQueryRoute, 
				otelhttp.WithRouteTag(
					c.spanRoutePrefix+QueryRoute,
					differentQueryProxy,
				),
			)
		})
```

With the changes in this PR both routes would report the same group and handler, but in fact they are different handlers.

douglascamata: The `handler` label can only be injected when the request arrives at the "final" `http.Handler`, often inside a `router.Handle` call. Otherwise we will need a whole new `router.Group` if we add a new route here.

douglascamata: Ah, we already have them separate anyway, so no problem. 😅

thibaultmg (commented):
Looks good!

I was wondering if we could also inject those static labels at startup time instead of dynamically injecting them on each request in the context.
Did you consider the possibility of instantiating new [intrumenters](https://github.com/observatorium/api/blob/81f8fdfd96a3d30c29ca54ca4ec9319d590055c2/main.go#L530) for each endpoint with the relevant set of static labels? I am not sure it is possible though, just a question.

philipgough (approved):
lgtm

PR #605, opened 2023-11-24
NETOBSERV-1324: implement fine-grained access for netobserv

jotak:
Draft PR to:
- show what's needed to have fine-grained access with netobserv (just 1 line change!)
- discuss how we want to implement that

### DIFF
--- a/authorization/http.go
+++ b/authorization/http.go
@@ -138,7 +138,7 @@ func WithAuthorizers(authorizers map[string]rbac.Authorizer, permission rbac.Per
 				selectorsInfo = emptySelectorsInfo
 			}
 
-			metadataOnly := isMetadataRequest(r.URL.Path)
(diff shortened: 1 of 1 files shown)


squat:
I want to make sure that we don't make changes to the way we do authn/authz without submitting proposals to the Observatorium project for community review. I would like to avoid introducing a magic tenant string like `network`, since this seems too tightly coupled to Red Hat / OpenShift logic.

jotak:
@squat that totally makes sense, my PR isn't meant to be merged as is, it's only showing what end result I would like to see.
As I am not deeply involved in observatorium, I am not sure which approach to take. A suggestion I can do is to introduce a new command line arg to explicitly disable namespace extraction from queries, hence falling back to the same path as this `isMetadataRequest` thing. What do you think, would it be a reasonable approach?

jotak:
Closing this PR, as proposed in my previous comment I will submit another one in opa-openshift to achieve the same

PR #615, opened 2023-12-04
Add features to rate limiting configuration to support load shedding

philipgough:
This change adds two features to the rate limiting configuration that enables some extra control over load shedding and traffic management for this proxy. These changes only apply to the remote rate limiting configuration. The local rate limiter already has its own variation of handling the retry-after header. 

The issue at hand is that we have had a number of occurrences now on the metrics ingest path due to the behaviour of the Prometheus remote write client based on the response code from this server. 

In summary, 4xx codes outside of 429s are not retried. 5xx response codes are retried and fall into `queue_config` section of https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write which the person(s)/teams who run this component may have no control over at runtime, or changes might be difficult to make.

We have observed a couple of issues, all leading to a retry storm that can eventually take down this component as well as networking to it (cluster ingress etc). When something on the request path goes slow under heavy load (observatorium-api, gubernator, thanos receive, etc) then we begin to see lots of 5xx responses which leads to a thundering herd, retry storm type problem.

The changes here are as follows:

- Addition of `failOpen` configuration which can allow requests through the stack in case of rate limiting failures.
- Addition of `retryAfterMin` and `retryAfterMax` which essentially put the control of the  `min_backoff` and `max_backoff` configuration on the Prometheus side, back in the hands of the service provider.
     -  https://github.com/prometheus/prometheus/issues/12674
     - https://github.com/prometheus/prometheus/pull/12677

The latter configuration max negatively effect latency to some degree since a lock will be required to track changes but this will only occur in situations where we are actively rate limiting or failing to do so, in which case the latency is a valid tradeoff.

Default behaviour remains the same as before in both cases so no breaking change.


ratelimit/http.go
```diff
@@ -119,15 +138,60 @@ func (l rateLimiter) Handler(next http.Handler) http.Handler {
...
 
 		if err != nil {
-			if err == errOverLimit {
+			// in all cases, where we will return an error
+			// if the rate limiter becomes overloaded or unavailable, we should start to force the client to back off
+			if retryAfter, ok := l.getAndSetNextRetryAfterValue(); ok {
+				w.Header().Set(headerRetryAfter, retryAfter)
+			}
+
+			if errors.Is(err, errOverLimit) {
```
thibaultmg, line 147: In this case shouldn't we set the `headerRetryAfter` header to `resetTime`?

thibaultmg: I mean `resetTime-now()`

thibaultmg (commented):
Awesome, this will help for 500s! Just one remark. Waiting for your reply, otherwise good to me!

ratelimit/http.go
philipgough: yeah looks like I missed the reset. So maybe it makes sense on the happy path to just delete the key, wdyt? That would reset us to min value?

thibaultmg (approved):

saswatamcode (approved):
Thanks for investigating this!

PR #620, opened 2023-12-15
Rate limiter: add an alternative implementation based on Redis.

douglascamata:
## First, why?

I believe Gubernator is not the ideal solution if `"local"` does work for a case. It is a powerful tool, but it has a considerable downside (like all distributed systems out there): it's not easy to operate and understand in general. Particularly in its "global" behavior. 

Because of this, it caused us a few problems. It was very difficult to find documentation/material out there about Gubernator (case studies, etc) that we could consume. 

## Okay, but why Redis?

Redis has been widely adopted as a backend solution. A simple web search will reflect this. Companies like [Stripe](https://stripe.com/blog/rate-limiters) and [Github](https://github.blog/2021-04-05-how-we-scaled-github-api-sharded-replicated-rate-limiter-redis/) scaled great with it. It's a reliable, efficient, and robust solution... yet it is relatively simple and well documented (even its complicated parts).

> But Redis is single threaded!

Yes, and it still scales well in its single thread model (up to hundreds of thousands of requests per second). It's able to scale beyond this with Redis Cluster, which shares many of the positives of Redis itself (reliable, efficient, robust, well known and documented).

## What did you do? How?! 

I implemented a rate limiter backend that uses Redis and the [Generic Cell Rate Algorithm](https://en.wikipedia.org/wiki/Generic_cell_rate_algorithm) (GCRA), which is a leaky bucket implementation.

The GCRA implementation lives inside a Lua script, which is embed in the binary, and has everything it needs to run safely in a single instance or Redis Cluster. The script is evaluated in/by Redis. This ensures it's safe for concurrent use, as Redis runs commands serially (the positives of single thread model, lol), and we don't need to care about thread-safety in Go-land.

Thanks to the awesome https://github.com/redis/rueidis library, the code will first attempt to execute the script in Redis by its SHA, using the `EVALSHA` command, to avoid extra network traffic and other overheads. If the command cannot find the script's SHA in Redis, a normal `EVAL` will be sent with the whole script, creating such cache. More details can be found in the `Script cache` section of [Redis' Lua scripting documentation](https://redis.io/docs/interact/programmability/eval-intro/).

### To the CLI

I added two extra command line options to support this: `middleware.rate-limiter.type` and `middleware.rate-limiter.address`. 

* `type` can be either `"local"` or `"redis"`.
* `address` can be repeated and is collected as a `[]string`. This enables usage of Redis Cluster.

The already existing `middleware.rate-limiter.grpc-address` has precedence over the previously mentioned options.

### To the CI (CircleCI)

I made the unit tests run in a virtual machine instead of Docker container, because I need Docker to spin up a Redis for tests. I considered moving this test to the `e2e` test folder, but it causes lots of problems with un-exported things and lots of extra setup is required. Taking this into consideration, I believe that the build change is okay.


README.md
```diff
@@ -143,8 +143,12 @@ Usage of ./observatorium-api:
     	The number of concurrent requests that can buffered.
   -middleware.concurrent-request-limit int
     	The limit that controls the number of concurrently processed requests across all tenants. (default 10000)
+  -middleware.rate-limiter.address value
+    	The address of the rate limiter. Only used when not using the gRPC nor "local" rate limiters. Can be repeated to specify multiple addresses (i.e. Redis Cluster.
```
thibaultmg, line 147: Closing parenthesis is missing 😉

ratelimit/redis.go
```diff
@@ -0,0 +1,82 @@
...
+		strconv.FormatInt(req.limit, 10),              // limit
+		strconv.FormatInt(req.duration, 10),           // period
+		"1",                                           // cost
+	}
+	result := inspectScript.Exec(ctx, r.client, []string{req.key}, rateLimitParameters)
+	limited, remaining, resetIn, err := r.parseRateLimitResult(&result)
+	if err != nil {
+		return 0, 0, err
+	}
+	resetTime = time.Now().Add(time.Duration(resetIn)).UnixMilli()
```
thibaultmg, line 50: Shouldm't we multiply the resetIn duration with time.Millisecond?

ratelimit/redis.go
```diff
@@ -0,0 +1,82 @@
...
+
+// RedisRateLimiter is a type that represents a rate limiter that uses Redis as its backend.
+// The rate limiting is a leaky bucket implementation using the generic cell rate algorithm.
+// See https://en.wikipedia.org/wiki/Generic_cell_rate_algorithm for details on how this algorithm works.
+type RedisRateLimiter struct {
+	client rueidis.Client
+}
+
+// Ensure RedisRateLimiter implements the SharedRateLimiter interface.
+var _ SharedRateLimiter = (*RedisRateLimiter)(nil)
```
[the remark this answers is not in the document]
douglascamata: `main` assigns it because of implementation details that could change at any point. Here this verification is part of the design and should never be removed or changed.

douglascamata:
@thibaultmg comments addressed. Please have another look whenever you can. 🙇

ratelimit/redis.go
```diff
@@ -0,0 +1,82 @@
...
+	return &RedisRateLimiter{client: client}, nil
+}
+
+// GetRateLimits retrieves the rate limits for a given request using a Redis Rate Limiter.
+// It returns the amount of remaining requests, the reset time in milliseconds, and any error that occurred.
+func (r *RedisRateLimiter) GetRateLimits(ctx context.Context, req *request) (remaining, resetTime int64, err error) {
+	inspectScript := rueidis.NewLuaScript(gcraRateLimitScript)
+	rateLimitParameters := []string{
+		strconv.FormatInt(time.Now().UnixMilli(), 10), // now
+		strconv.FormatInt(req.limit, 10),              // burst
```
thibaultmg, line 40: Is this intended, we want the burst to equal the limit always? Or should it be 0?

douglascamata: For now, this is intended. I didn't want to expose too much configuration yet. Later this can be made configurable too. 👍

thibaultmg (approved):
Awesome! Happy to have reviewed my first LUA script 😅 .
I wonder if taking the current time from the query instead of from Redis could cause some issues if one of the clients is in the future. But looks good nevertheless

thibaultmg (approved):
Nice fix!

ratelimit/redis.go
thibaultmg: I misunderstood how it works. Clarified in latest PR with comments. Nothing to change

thibaultmg (approved):

douglascamata:
> The absence of a remote endpoint already infers that we would use local rate limting, why do we need to add local as a type?

This scenario is still the same. If there isn't a remote endpoint indicated, then the default value of the rate limiter type is "local". So current setups using local rate limiter won't require any change to keep using it. This is not a breaking change.

Keeping the default type to `"local"` and not `""` (simple empty string) helps making the code communicate its intent much more clearly: one doesn't have to know that empty string means local cache.

> It would be a breaking change, but I wonder in a way would it make sense if `middleware.rate-limiter.grpc-address` became bundled into `middleware.rate-limiter.address` and we could specify type of either `gubernator` (default) and `redis`.

I want to avoid the breaking change and I agree with you that the best would be to unify the cache flags.

philipgough:
> I want to avoid the breaking change and I agree with you that the best would be to unify the cache flags.

If we are to merge this, I am actually suggesting a breaking change by unifying the flags across the rate limiters here might make sense. I know this is less than ideal but it also feels weird right now to distinguish the remote rate limiter simply by protocol in the flag.

The secondary issue is that the protocol in the flag for Gubernator will make zero sense if the protocol is referenced is being removed from the project like I linked to earlier.

douglascamata:
@philipgough how can we move forward with this? I prefer to not have to deal with this breaking change in our deployments & projects, as it demands too much attention and care. I don't see a good enough reason for it.

philipgough:
> how can we move forward with this?

Getting the CI green would be a good start :) - i have no objections to this moving forward as is.

philipgough (approved):

PR #633, opened 2024-01-08
Pin k8s deps to semver patches

periklis:
The following PR reverts the following to PRs to the previous state for k8s dependencies. To enable simpler reasoning on which k8s release the api is used along with operators (e.g. Loki Operator, Tempo Operator) this PR also pins dependabot to upgrade only on `semver-patch` candence.

Replaces:
- #625 
- #628


saswatamcode (approved):
Thanks! LGTM!

PR #649, opened 2024-02-09
Print binary version info on gateway startup

rubenvp8510:


main.go
```diff
@@ -81,6 +81,14 @@ const (
 	// Server shutdown grace period.
 	gracePeriod = 2 * time.Minute
 )
+const appName = "observatorium-api"
```
periklis, line 84: We should use the `--debug.name` instead of the constant because this repo is also whitelabeled as `lokistack-gateway` in Loki Operator managed deployments.

periklis (changes requested):

Makefile
```diff
@@ -79,7 +80,7 @@ benchmark.md: $(EMBEDMD) tmp/load_help.txt
 	$(EMBEDMD) -w docs/benchmark.md
 
 $(BIN_NAME): deps main.go rules/rules.go $(wildcard *.go) $(wildcard */*.go)
-	CGO_ENABLED=0 GOOS=$(OS) GOARCH=$(ARCH) GO111MODULE=on GOPROXY=https://proxy.golang.org go build -a -ldflags '-s -w' -o $(BIN_NAME) .
+	CGO_ENABLED=0 GOOS=$(OS) GOARCH=$(ARCH) GO111MODULE=on GOPROXY=https://proxy.golang.org go build -a -ldflags '-s -w  -X main.Version=$(VERSION) -X main.Branch=$(GIT_BRANCH)' -o $(BIN_NAME) .
```
periklis, line 83: Missing `-X main.Revision` here maybe?

periklis:
@rubenvp8510 Since we add this nice way to package the version with the binary, I suggest to drop by default a log line when some starts this container anyway? WDYT?

main.go
rubenvp8510: oh! good to know that. will do the change.

rubenvp8510:
@periklis Thanks for the review, I've already added the changes requested. Please give it another review when you get a chance.

rubenvp8510:
@periklis Please review when you get a chance! Thanks

main.go
```diff
@@ -258,9 +265,26 @@ type tenant struct {
...
+
 	cfg, err := parseFlags()
+
+	stdlog.Println(version.Print(cfg.debug.name))
+
+	if *printVersion {
+		// Only prints the version and exit
+		os.Exit(0)
+	}
+
```
periklis, line 287: Two suggestions below:
1. Use `version.Info()` for the default logging statement. This a good enough one-liner when running the container.
2. Move the flag parsing to `parseFlags`

```suggestion
	cfg, err := parseFlags()
	if err != nil {
		stdlog.Fatalf("parse flag: %v", err)
	}

	if cfg.printVersionOnly {
		stdlog.Println(version.Print(cfg.debug.name))
		os.Exit(0)
	} else {
		stdlog.Println(version.Info())
	}
```

periklis (changes requested):

rubenvp8510:
@periklis Done with the changes! Thanks!

periklis:
@rubenvp8510 Just for clarification, do we need the command at all if we by default log the version?

rubenvp8510:
> @rubenvp8510 Just for clarification, do we need the command at all if we by default log the version?

Is not required, but may be good to have the only thing you want is to know the version. and don't see any weird error like "configuration missing" or something.

periklis:
> > @rubenvp8510 Just for clarification, do we need the command at all if we by default log the version?
> 
> Is not required, but may be good to have if the only thing you want is to know the version. and don't see any weird error like "configuration missing" or something.

For the sanity to keep long-term maintenance low and since we do not have other subcommands in this binary, I would suggest to skip adding the extra subcommand and keep the log line only.

rubenvp8510:
> > > @rubenvp8510 Just for clarification, do we need the command at all if we by default log the version?
> > 
> > 
> > Is not required, but may be good to have if the only thing you want is to know the version. and don't see any weird error like "configuration missing" or something.
> 
> For the sanity to keep long-term maintenance low and since we do not have other subcommands in this binary, I would suggest to skip adding the extra subcommand and keep the log line only.

Ok, I'll remove it then.

rubenvp8510:
@periklis done!

rubenvp8510:
@periklis Can you review this again please? Thank you

saswatamcode (approved):
Thanks! Generally LGTM!
Maybe at some point we could have some subcommands, but this works for now.

periklis (approved):

PR #663, opened 2024-03-01
Bump golang 1.21 and k8s 1.28.7

periklis:


jotak (approved):
sounds good from my side

douglascamata (commented):
Do we need the bump for any specific reason?

periklis:
> Do we need the bump for any specific reason?

We usually bump this to keep up with golang support upstream. However we do not do this so regular as we should because we wait for downstream pipelines providing the golang builders. Usually the downstream builds like in OpenShift Logging/Distributed Tracing benefit from getting CVE fixes faster in golang if they use newer versions.

PR #714, opened 2024-06-10
Add support for logs OTLP route

periklis:
The following PR enables the Loki OTLP route to allow ingesting logs through OTLP.

cc @pavolloffay

### DIFF
--- a/api/logs/v1/http.go
+++ b/api/logs/v1/http.go
@@ -36,6 +36,9 @@ const (
 	volumeRoute            = "/loki/api/v1/index/volume"
 	volumeRangeRoute       = "/loki/api/v1/index/volume_range"
 
+	otlpRoute = "/otlp/v1/logs"
+	pushRoute = "/loki/api/v1/push"
+
 	prometheusRulesRoute  = "/prometheus/api/v1/rules"
 	prometheusAlertsRoute = "/prometheus/api/v1/alerts"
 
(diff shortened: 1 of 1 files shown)


api/logs/v1/http.go
```diff
@@ -34,6 +34,9 @@ const (
 	rulesPerNamespaceRoute = "/loki/api/v1/rules/{namespace}"
 	rulesPerGroupNameRoute = "/loki/api/v1/rules/{namespace}/{groupName}"
 
+	otlpRoute = "/otlp/v1/logs"
```
pavolloffay, line 37: We should think about providing a unified experience for users.

The OTLP HTTP for traces follows this URL `https://host:port/api/traces/v1/{tenant}/v1/traces`

Shall we remote `otlp/` from the endpoint or perhaps add it to the trace endpoint?

pavolloffay (commented):
Just one question about endpoint alignment across signals. 



Shall we add e2e test https://github.com/observatorium/api/blob/576cb11884bb50795035cf37677974609e37307e/test/e2e/traces_test.go#L276 ?

api/logs/v1/http.go
periklis: I am for keeping the `oltp/` prefix for logs and add it to the traces endpoint too. My main concern is that based on our URL based scheme in observatorium-api we configure other collectors e.g. vector when using the Loki/OTLP sink to write to us by using the URL scheme `<host>/api/v1/logs/<tenant>` and let it append the Loki url, e.g. for the old api `<host>/api/v1/logs/<tenant>/loki/api/v1/push` and for the OTLP API in Loki 3.0 `<host>/api/v1/logs/<tenant>/otlp/v1/logs`.

In short I believe to keep compatibility with other collectors sink implementation and our URL-based approach we should keep the `otlp/` prefix

periklis:
> Just one question about endpoint alignment across signals.
> 
> Shall we add e2e test
> 
> https://github.com/observatorium/api/blob/576cb11884bb50795035cf37677974609e37307e/test/e2e/traces_test.go#L276
> ?

I intend to add an OTLP e2e tests for the logs API too. I just have to figure out a small scale Loki 3.0 config for such a test. Working on that.

PR #716, opened 2024-06-15
Add watcher for upstreams TLS certificates

rubenvp8510:
This PR adds a watcher for the certificates so in case of rotation, the certificates will be reloaded.

Added flags to disable/enable this per signal, and in the general server

### DIFF
--- a/tls/ca_watcher.go
+++ b/tls/ca_watcher.go
@@ -1,0 +1,121 @@
+package tls
+
+import (
+	"context"
+	"crypto/sha256"
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"sync"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+)
+
+type CAWatcher struct {
+	caMutex         sync.RWMutex
+	certPool        *x509.CertPool
+	logger          log.Logger
+	fileHashContent string
+	CAPath          string
+	interval        time.Duration
+}
+
+func NewCAWatcher(CAPath string, logger log.Logger, interval time.Duration) (*CAWatcher, error) {
+	w := &CAWatcher{
+		CAPath:   CAPath,
+		logger:   logger,
+		certPool: x509.NewCertPool(),
+		interval: interval,
+	}
+	err := w.loadCA()
+	return w, err
+}
+
+func (w *CAWatcher) Watch(ctx context.Context) error {
+	var timer *time.Timer
+
+	scheduleNext := func() {
+		timer = time.NewTimer(w.interval)
+	}
+	scheduleNext()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-timer.C:
+			err := w.loadCA()
+			if err != nil {
+				return err
+			}
+			scheduleNext()
+		}
+	}
+
+}
+
+func (w *CAWatcher) loadCA() error {
+
+	hash, err := w.hashFile(w.CAPath)
+	if err != nil {
+		level.Error(w.logger).Log("unable to read the file", "error", err.Error())
+		return err
+	}
+
+	// If file changed
+	if w.fileHashContent != hash {
+		// read content
+		caPEM, err := os.ReadFile(filepath.Clean(w.CAPath))
+		if err != nil {
+			level.Error(w.logger).Log("failed to load CA %s: %w", w.CAPath, err)
+			return err
+		}
+		// prevent concurrent updates to the same certPool
+		w.caMutex.Lock()
+		defer w.caMutex.Unlock()
+
+		certPool := x509.NewCertPool()
+		if !certPool.AppendCertsFromPEM(caPEM) {
+			level.Error(w.logger).Log("failed to parse CA %s", w.CAPath)
+			return err
+		}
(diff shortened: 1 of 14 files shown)


periklis:
Why only the traces endpoint? Can't we make it available but opt-in to every signal?

tls/fswatcher.go
```diff
@@ -0,0 +1,135 @@
...
+type fsWatcher struct {
+	watcher            *fsnotify.Watcher
+	fileHashContentMap map[string]string
+	onChange           func()
+	mu                 sync.RWMutex
+	logger             log.Logger
+}
+
+func newWatcher(filepaths []string, onChange func(), logger log.Logger) (*fsWatcher, error) {
+	watcher, err := fsnotify.NewWatcher()
```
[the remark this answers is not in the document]
rubenvp8510: Make sense to me, I think the timer approach can work, I have seen that approach before working good. and we can get rid of all the problems we could have with fsnotify.

tls/options.go
```diff
@@ -0,0 +1,150 @@
...
+
+	// reload enabled
+	if interval != nil {
+		return newWithWatchers(upstreamCertFile, upstreamKeyFile, upstreamCAFile, *interval, logger, ctx, g)
+	}
+
+	return newNoWatchers(upstreamCertFile, upstreamKeyFile, upstreamCAFile)
+}
+
+func newWithWatchers(upstreamCertFile, upstreamKeyFile, upstreamCAFile string,
```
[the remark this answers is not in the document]
rubenvp8510: Not reason, sorry I'll put the argument in the correct order.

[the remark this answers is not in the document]
rubenvp8510: Changed

tls/ca_watcher.go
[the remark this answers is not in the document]
rubenvp8510: Done. I put comments, and transformed some methods to private, because it doesn't make sense to have it public. Anyway I added some comments and descriptions.

tls/options.go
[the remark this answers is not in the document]
rubenvp8510: Added.

rubenvp8510:
@douglascamata Sorry for the late response. this is ready for review again. Thank you very much!

tls/options.go
```diff
@@ -107,23 +110,29 @@ func startCAReloader(ctx context.Context, g run.Group, upstreamCAFile string, in
...
 func (uo *UpstreamOptions) hasUpstreamCerts() bool {
 	return uo.cert != nil || uo.certReloader != nil
 }
 
+// hasCA determine if the CA watcher is enabled.
 func (uo *UpstreamOptions) isCAReloadEnabled() bool {
 	return uo.caReloader != nil
 }
 
+// hasCA determine if the certificate watcher is enabled.
```
douglascamata, line 128: Looks like something went wrong with the doc of these methods here.

tls/ca_watcher.go
```diff
@@ -15,7 +15,8 @@ import (
 	"github.com/go-kit/log/level"
 )
 
-type CAWatcher struct {
+// caWatcher poll for changes on the CA file, if the CA change it will add it to the certificate Pool.
```
douglascamata, line 18: What is the CA file? Is it the CA key, or the CA cert? 🤔 

I believe it's the certificate given what I see, but would be nice to be explicit about it.

douglascamata (changes requested):
Just a few minor things left, imho.

tls/options.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/go-kit/log"
+	"github.com/oklog/run"
+)
+
+// UpstreamOptions represents the options of the upstream TLS configuration
+// this structure contains the certificates and the watchers if the certificate/ca watchers are enabled.
+type UpstreamOptions struct {
+	cert         *stdtls.Certificate
+	ca           []byte
+	certReloader *rbacproxytls.CertReloader
```
douglascamata, line 20: I think we could improve the name here to try to make things easier to understand. `*rbacproxytls.CertReloader` is a certificate reloader for the tenants MTLS certs. So maybe `UpstreamOptions.certReloader` could be named `UpstreamOptions.tenantsCertReloader`.

Without this it makes me wonder: why there is a `certReloader` and a `caReloader` here? Both are certs!

This brings me to wonder about the name of the struct itself. What is "upstream" in `UpstreamOptions`? 🤔

rubenvp8510: why `tenantsCertReloader`? I don't see where the tenants are involved here. Upstream means all upstream services for logs/metrics/traces.  At least is the terminology that is used in all the gateway. for instance:

`    	File containing the TLS client key to authenticate against upstream logs servers. Leave blank to disable mTLS.`

tls/ca_watcher.go
rubenvp8510: Done, I put the clarification

tls/options.go
rubenvp8510: Fixed! Thanks!

tls/options.go
douglascamata: Humm, I see. So I got it wrong. Probably we could clarify this. There are many certs everywhere, and reloaders for them, so things are confusing.

philipgough (approved):
/lgtm

douglascamata (approved):

PR #751, opened 2024-09-06
fix(logql): Add missing support for json parser expr with labels

periklis:
The logQL parser expression returns an syntax error when used with a label selector list, i.e.:
```
{log_type="application",kubernetes_namespace_name="demo-1"}|json pipeline_run_uid="kubernetes.labels.tekton_dev_pipelineRunUID" msg=".message" 
```
returns
```
failed parsing LogQL expression: syntax error: unexpected IDENTIFIER: 1:66
```

The fix enables this missing feature in the logQL parser syntax.

### DIFF
--- a/logql/v2/parser_test.go
+++ b/logql/v2/parser_test.go
@@ -217,7 +217,7 @@ func TestParseExpr(t *testing.T) {
 				filter: LogPipelineExpr{
 					&LogLabelExpr{
 						parser: "logfmt",
-						labels: []LogLabel{
+						labels: LogLabelList{
 							{
 								identifier: "addr",
 								matcher:    nil,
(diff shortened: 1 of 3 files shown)


logql/v2/parser_test.go
```diff
@@ -815,6 +815,49 @@ func TestParseExpr(t *testing.T) {
 				},
 			},
 		},
+		{
+			input: `{first="value"} | json userid="spec.user.id", group="spec.user.group" | level = "info"`,
```
xperimental, line 819: Do we need a test case for "just one parameter"?

xperimental (commented):
Looks fine for me, just had one idea.

PR #762, opened 2024-10-07
Bump go1.22 and k8s 1.30.5

periklis:
cc @JoaoBraveCoding @xperimental


xperimental:
Already talked to Peri about this, just leaving this here so that others are aware as well:

The k8s 1.30 libraries require Go 1.22, which we currently can not update to due to a bug in relation to FIPS. It looks as if a fix might be happening soon, but as long as the fix is not available, this can not be merged. An option would be to just upgrade to 1.29 as an intermediary step. And we're currently discussing if the issue is actually relevant to us (mostly because I don't understand the mechanism by which it is affecting us).

periklis:
> Already talked to Peri about this, just leaving this here so that others are aware as well:
> 
> The k8s 1.30 libraries require Go 1.22, which we currently can not update to due to a bug in relation to FIPS. It looks as if a fix might be happening soon, but as long as the fix is not available, this can not be merged. An option would be to just upgrade to 1.29 as an intermediary step. And we're currently discussing if the issue is actually relevant to us (mostly because I don't understand the mechanism by which it is affecting us).

Re-tested go 1.22 binaries on a OCP 4.17 FIPS cluster with an up2date runc and all works as expected. I cannot disclose the exact details here but it seems an older version of runc exec was unable to run binaries which used `dlopen` for openssl.

.bingo/faillint.mod
```diff
@@ -2,4 +2,4 @@ module _ // Auto generated by https://github.com/bwplotka/bingo. DO NOT EDIT
 
 go 1.19
```
JoaoBraveCoding, line 3: Why did we update the go version on `.bingo/golangci-lint.mod` but not in here?

authentication/grpc.go
```diff
@@ -33,10 +32,10 @@ func WithGRPCTenantHeader(header string, tenantIDs map[string]string, logger log
 
 		headerTenants := md[header]
 		if len(headerTenants) == 0 {
-			return status.Errorf(codes.InvalidArgument, fmt.Sprintf("header %q not set", header))
+			return status.Errorf(codes.InvalidArgument, "header %q not setf", header)
```
JoaoBraveCoding, line 35: looks to be a typo
```suggestion
			return status.Errorf(codes.InvalidArgument, "header %q not set", header)
```

.bingo/faillint.mod
periklis: I believe this is what bingo is doing internally. golangci-lint requires go 1.22 but faillint not?!?

JoaoBraveCoding (approved):
/lgtm

PR #764, opened 2024-10-10
Fix logQL metric with grouping losing opMetric

jotak:
Previously, parsing
```
avg_over_time({app="foo"}[15s]) by (app)
```
would result in:
```
({app="foo"}[15s]) by (app)
```

Now it correctly keeps the metric function.

Another issue fixed is allowing enclosing parenthesis for range/metric expressions, for instance:
```
(count_over_time({first="value"}[10h]))
```

is now allowed

Ref: [NETOBSERV-1921](https://issues.redhat.com/browse/NETOBSERV-1921)


logql/v2/parser_test.go
```diff
@@ -1214,28 +1214,6 @@ func TestParseExpr(t *testing.T) {
 				},
 			},
 		},
-		{
```
jotak, line 1217: hey @periklis 
I'm removing this test because now it fails, as the enclosing parenthesis are removed (so the parsed expression is not strictly equal to the input expression)
But I would say this test intention is covered via the new test I'm adding in ast_test:

```go
		{
			input:  `(count_over_time({first="value"}[10h]))`,
			output: `count_over_time({first="value", second="next"}[10h])`,
		},
```

what do you think?

jotak: test restored

periklis (approved):

PR #768, opened 2024-10-16
Remove jaeger exporter and use OTLP/http

pavolloffay:
The jaeger exporter was removed in October 2023 https://github.com/open-telemetry/opentelemetry-go/blob/main/CHANGELOG.md#11800410006-2023-09-12 

This PR introduces a breaking change on flags, however I think it's better than making the existing flag noop - users might not notice the change and tracing will stop working. 

OTLP/http is well supported e.g. in OTEL collector, Jaeger and other agents/collectors.


andreasgerstmayr (approved):

saswatamcode (approved):

philipgough (approved):

moadz (commented):
Feels like changing it in-place might be confusing for users as the otlp/http endpoint is not backwards compatible with the jaeger endpoint (unless i misunderstand this change). Probably better to deprecate the flag over rebinding it.

PR #788, opened 2025-01-08
Adjust log level of 'fallback to read cookie, ...' log message

andreasgerstmayr:
When accessing the Jaeger UI, cookies are used for authentication.
At the moment, every HTTP request to the Jaeger UI prints this message
in the logs:

    fallback to read cookie, no serviceaccount bearer token or mTLS certs provided

which may concern some users. Reading cookies is part of the normal operation,
therefore let's print this message only when the log level is set to debug.


philipgough (approved):
/lgtm

PR #789, opened 2025-01-22
Add trace query RBAC

pavolloffay:
Flag `--traces.query-rbac=true` enables query RBAC. The user will be able to see span/resource/scope attributes and events only from namespaces it can access (the list of namespaces is returned form opa-openshift).


api/traces/v1/trace_rbac.go
```diff
@@ -0,0 +1,182 @@
...
+	tracev1 "github.com/grafana/tempo/pkg/tempopb/trace/v1"
+
+	apilogsv1 "github.com/observatorium/api/api/logs/v1"
+)
+
+func WithTraceQLNamespaceSelect() func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			query := r.URL.Query()
+			q := query.Get("q")
```
andreasgerstmayr, line 28: This modification should only happen for calls to `/api/search`

api/traces/v1/trace_rbac.go
```diff
@@ -0,0 +1,182 @@
...
+					reader = response.Body
+				}
+
+				b, err := io.ReadAll(reader)
+				if err != nil {
+					return err
+				}
+				strResponse := string(b)
+				buf := bytes.NewBufferString(strResponse)
+				response.Header["Content-Length"] = []string{fmt.Sprint(buf.Len())}
```
andreasgerstmayr, line 81: The new, modified response (after `traceRBAC()`) can have a different length than the original response

api/traces/v1/trace_rbac.go
```diff
@@ -0,0 +1,182 @@
...
+		return nil
+	}
+}
+
+func traceRBAC(allowedNamespaces map[string]bool, trace *tempopb.Trace) *tempopb.Trace {
+	for _, rs := range trace.ResourceSpans {
+		notAllowedNamespace := ""
+		if rs.Resource != nil && rs.Resource.Attributes != nil {
+			for _, resAttr := range rs.Resource.Attributes {
+				if resAttr.Key == "k8s.namespace.name" && !allowedNamespaces[resAttr.Value.GetStringValue()] {
```
andreasgerstmayr, line 129: Traces without the `k8s.namespace.name` attribute should also be filtered, they could leak sensitive data.

api/traces/v1/trace_rbac.go
```diff
@@ -0,0 +1,182 @@
...
+		traceSearchMetadata.SpanSet = spanSetRBAC(allowedNamespaces, traceSearchMetadata.GetSpanSet())
+	}
+	return searchResponse
+}
+
+func spanSetRBAC(allowedNamespaces map[string]bool, spanSet *tempopb.SpanSet) *tempopb.SpanSet {
+	for _, span := range spanSet.GetSpans() {
+		notAllowedNamespace := ""
+		for _, attribute := range span.GetAttributes() {
+			if attribute.GetKey() == "k8s.namespace.name" && !allowedNamespaces[attribute.GetValue().GetStringValue()] {
```
andreasgerstmayr, line 168: What happens when you add `| select(resource.k8s.namespace.name)` to the search query, but the trace doesn't have a `k8s.namespace.name` attribute? I think to be safe I'd filter in case this attribute is (for any reason) not present in the search response.

pavolloffay: yes I have added a test case for it

api/logs/v1/labels_enforcer.go
```diff
@@ -178,3 +182,18 @@ func initAuthzMatchers(lm []*labels.Matcher) ([]*labels.Matcher, error) {
...
+// AllowedNamespaces returns the list of namespaces that the user is allowed to list.
+func AllowedNamespaces(ctx context.Context) []string {
+	matchers := ctx.Value(matchersContextKey{})
+	if matchers == nil {
+		return nil
+	}
+	matchersTyped := matchers.(AuthzResponseData)
+
+	var namespaces []string
+	for _, m := range matchersTyped.Matchers {
```
andreasgerstmayr, line 195: Does `matchersTyped.Matchers` always contain a list of `|`-separated namespace names?
It's not directly related to this PR, but I wonder why the attribute is called `Matchers` and not `Namespaces`.

api/traces/v1/trace_rbac.go
```diff
@@ -0,0 +1,227 @@
...
+	return func(response *http.Response) error {
+
+		allowedNamespaces := map[string]bool{}
+		namespaces := apilogsv1.AllowedNamespaces(response.Request.Context())
+		for _, ns := range namespaces {
+			allowedNamespaces[ns] = true
+		}
+		level.Debug(log).Log("AllowedNamespaces", allowedNamespaces)
+
+		if strings.Contains(response.Request.URL.Path, "/api/traces/") || strings.Contains(response.Request.URL.Path, "/api/search") {
```
andreasgerstmayr, line 66: Should we send a 403 error for other API endpoints?
https://grafana.com/docs/tempo/latest/api_docs/#endpoints

https://grafana.com/docs/tempo/latest/api_docs/#search-tag-values-v2 could leak private data (but this will also disable the auto-complete in the tracing UI :/)

api/logs/v1/labels_enforcer.go
pavolloffay: It comes from the Prometheus API. I think this is what the OPA openshift returns

pavolloffay: https://github.com/observatorium/opa-openshift?tab=readme-ov-file#post-v1datapackagerule

api/traces/v1/trace_rbac.go
pavolloffay: That is a great point. 

What endpoints is using the UI? 

Another iteration on this feature will be supporting a white listing of attributes. Then we could enable it and return only allowed ones.

api/traces/v1/trace_rbac.go
```diff
@@ -0,0 +1,227 @@
...
+			}
+			traceQL = traceQL + " | select(resource.k8s.namespace.name)"
+			query.Set("q", traceQL)
+			r.URL.RawQuery = query.Encode()
+			next.ServeHTTP(w, r)
+		})
+	}
+}
+
+func responseRBACModifier(log log.Logger) func(response *http.Response) error {
```
JoaoBraveCoding, line 56: General curiosity question, I'm certainly missing context but AFAIU, in the current implementation we are using OPA to extract the namespaces and then we are doing the authorization bit on the response in obs/api instead of OPA doing authorization on the query itself. Why did we go with such an approach?

api/traces/v1/trace_rbac.go
andreasgerstmayr: The UI uses
```
/api/traces
/api/search
/api/v2/search/tags
/api/v2/search/tag
```

The last 2 are for auto-completion of attribute names and values.

api/traces/v1/trace_rbac.go
pavolloffay: This predates my involvement in the project. I think they wanted to separate that functionality into an external component.

api/traces/v1/trace_rbac.go
pavolloffay: Added the 403 check.

andreasgerstmayr (approved):

philipgough (approved):
lgtm

PR #800, opened 2025-02-17
 logs: extend LokiStack authorization to support OTel Semantics 

JoaoBraveCoding:
In general this PR:

- Introduces a new middleware (`WithNamespaceLabelEnforcer`) to ensure that log queries do no allow for users to set both `kubernetes_namespace_name` and `k8s_namespace_name`
- Modifies the `WithEnforceRulesLabelFilters` middleware to also enforce that user do not query with both `kubernetes_namespace_name` and `k8s_namespace_name` but only a single one.


api/logs/v1/labels_enforcer.go
```diff
@@ -197,3 +197,49 @@ func AllowedNamespaces(ctx context.Context) []string {
 	}
 	return namespaces
 }
+
+// WithNamespaceLabelEnforcer returns a middleware that ensures queries do not use both
+// kubernetes_namespace_name and k8s_namespace_name labels.
+func WithNamespaceLabelEnforcer() func(http.Handler) http.Handler {
```
periklis, line 203: Why a new one and not adapting `WithEnforceAuthorizationLabels`? What is the old one not handling well?

JoaoBraveCoding: I opted for creating a new middleware because:
1. these changes are only required for the migration period from ViaQ to OTEL. Then, once we would consider the migration over we could just be able to delete the middleware.
2. minimize the impact on other signals mainly tracing. Since they also use our version of `WithEnforceAuthorizationLabels` 

That being said I'm also happy to port them to `WithEnforceAuthorizationLabels` I would just need to move the check inside both `enforceValuesOnLabelValues` and `enforceValuesOnQuery`

periklis: I guess it is ok for a migration-period-only middleware. Suggest to name it like this to make this visible for the code reader in 3/5 mo time.

xperimental:
For other reviewers: @JoaoBraveCoding and I just discussed a different approach for what we want to achieve that will move some of these changes into `opa-openshift` instead, so this PR is currently "work-in-progress" again.

PR #578, opened 2023-09-27
Use Rules API to filter alerts/rules for application tenant

JoaoBraveCoding:
Refactors some of the code introduced in https://github.com/observatorium/api/pull/532 thanks to https://github.com/grafana/loki/pull/9128

Issue: https://issues.redhat.com/browse/LOG-4209

Testing:
General set up
```shell
# Set up cluster
# Install cluster-logging-operator
# In log-tool-box 
# Install bootstrap
kc apply -k  manifests/bootstrap
# Install CLF
kc apply -k manifests/cluster-logging/loki-default-audit/
# In obs/api 
# Build & Push container
export DOCKER_REPO=$MY_REPO/api
make container
podman push $CONTAINER_IMAGE
# In grafana/loki/operator 
# Patch the overlay in operator/config to use your obs/api image $CONTAINER_IMAGE
./hack/deploy-aws-storage-secret.sh $MY_BUCKET
# Deploy loki olm
make olm-deploy ....
oc apply -f hack/lokistack_gateway_ocp.yaml 
```
Feature testing
```shell
# Create RecordingRule & AlertingRule for user testuser-0 and testuser-1
# Grant necessary permissions to user
oc adm policy add-cluster-role-to-user monitoring-rules-view testuser-0
oc adm policy add-cluster-role-to-user monitoring-rules-edit testuser-0
# Login as user testuser-0 and get API TOKEN
# Use token to CURL rules API with parameters 
# Should work
curl -k -H "Authorization: Bearer $LOKI_BEARER_TOKEN" https://lokistack-dev-openshift-logging.apps.MY_CLUSTER/api/logs/v1/application/loki/api/v1/rules?kubernetes_namespace_name=log-test-0
#  Should not return any rules
curl -k -H "Authorization: Bearer $LOKI_BEARER_TOKEN" https://lokistack-dev-openshift-logging.apps.MY_CLUSTER/api/logs/v1/application/loki/api/v1/rules?kubernetes_namespace_name=log-test-1
```

PR #805, opened 2025-02-28
logs: re-introduce rules filtering for Prometheus API as this is still not supported by Loki

JoaoBraveCoding:
In https://github.com/observatorium/api/pull/578 we completely removed the filtering code and relied only on the query parameter for Loki to filter the rules. However, the filtering only works for the Loki API and not for the Prometheus API

Fixes https://issues.redhat.com/browse/LOG-6148

### DIFF
--- a/api/logs/v1/rules_prometheus_labels_enforcer.go
+++ b/api/logs/v1/rules_prometheus_labels_enforcer.go
@@ -20,8 +20,8 @@ import (
 const contentTypeApplicationJSON = "application/json"
 
 var (
-	errUnknownTenantKey        = errors.New("Unknown tenant key")
-	errUnknownRulesContentType = errors.New("Unknown rules response content type")
+	errUnknownTenantKey        = errors.New("unknown tenant key")
+	errUnknownRulesContentType = errors.New("unknown rules response content type")
 )
 
 type alert struct {
@@ -253,9 +253,9 @@ func filterPrometheusRuleGroups(groups []*ruleGroup, matchers map[string]string)
 
 	for _, group := range groups {
 		var filteredRules []rule
-		for _, rule := range group.Rules {
-			if hasMatchingLabels(&rule, matchers) {
-				filteredRules = append(filteredRules, rule)
+		for _, r := range group.Rules {
+			if hasMatchingLabels(&r, matchers) {
+				filteredRules = append(filteredRules, r)
 			}
 		}
 
@@ -270,9 +270,9 @@ func filterPrometheusRuleGroups(groups []*ruleGroup, matchers map[string]string)
 
 func filterPrometheusAlerts(alerts []*alert, matchers map[string]string) []*alert {
 	var filtered []*alert
-	for _, alert := range alerts {
-		if hasMatchingLabels(alert, matchers) {
-			filtered = append(filtered, alert)
+	for _, a := range alerts {
+		if hasMatchingLabels(a, matchers) {
+			filtered = append(filtered, a)
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


api/logs/v1/rules_prometheus_labels_enforcer.go
```diff
@@ -0,0 +1,280 @@
...
+	"github.com/observatorium/api/authentication"
+	"github.com/observatorium/api/authorization"
+	"github.com/prometheus/prometheus/model/labels"
+)
+
+const contentTypeApplicationJSON = "application/json"
+
+var (
+	errUnknownTenantKey        = errors.New("Unknown tenant key")
+	errUnknownRulesContentType = errors.New("Unknown rules response content type")
```
xperimental, line 24: Make the error message start lower-case?

```suggestion
	errUnknownTenantKey        = errors.New("unknown tenant key")
	errUnknownRulesContentType = errors.New("unknown rules response content type")
```

api/logs/v1/rules_prometheus_labels_enforcer.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+func filterPrometheusRuleGroups(groups []*ruleGroup, matchers map[string]string) []*ruleGroup {
+	var filtered []*ruleGroup
+
+	for _, group := range groups {
+		var filteredRules []rule
+		for _, rule := range group.Rules {
+			if hasMatchingLabels(&rule, matchers) {
+				filteredRules = append(filteredRules, rule)
```
xperimental, line 258: The variable `rule` shadows the `rule` type, which is not an error but can be confusing (feel free to pick a more descriptive name than `r`, but I think for this short block it works fine):

```suggestion
		for _, r := range group.Rules {
			if hasMatchingLabels(&r, matchers) {
				filteredRules = append(filteredRules, r)
```

api/logs/v1/rules_prometheus_labels_enforcer.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	return filtered
+}
+
+func filterPrometheusAlerts(alerts []*alert, matchers map[string]string) []*alert {
+	var filtered []*alert
+	for _, alert := range alerts {
+		if hasMatchingLabels(alert, matchers) {
+			filtered = append(filtered, alert)
```
xperimental, line 275: The variable `alert` shadows the `alert` type, which is not an error but can be confusing:

```suggestion
	for _, a := range alerts {
		if hasMatchingLabels(a, matchers) {
			filtered = append(filtered, a)
```

api/logs/v1/rules_prometheus_labels_enforcer.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+type prometheusRulesResponse struct {
+	Status    string    `json:"status"`
+	Data      rulesData `json:"data"`
+	Error     string    `json:"error"`
+	ErrorType string    `json:"errorType"`
+}
+
+func newModifyResponseProm(logger log.Logger, labelKeys map[string][]string) func(*http.Response) error {
```
xperimental, line 134: Some of the error code-paths are not covered by the current set of tests. Should we add one or more tests to cover some of them? Most of them seem to be simple cases, so it's more a question of "completeness" and if it's worth the effort (not talking about all the paths where it's an IO operation or marshalling).

xperimental (commented):
Seems to work fine in tests :+1: 

I have added some comments, but none of them would be a show-stopper.

api/logs/v1/rules_prometheus_labels_enforcer.go
JoaoBraveCoding: Added tests for alerts let me know if you were also referring to anything else

PR #874, opened 2026-02-25
fix: OIDC issuerCA to load all certificates from PEM bundle

JoaoBraveCoding:
Solves https://issues.redhat.com/browse/LOG-8727

### DIFF
--- a/authentication/oidc.go
+++ b/authentication/oidc.go
@@ -45,7 +45,7 @@ type oidcConfig struct {
 	GroupClaim    string `json:"groupClaim"`
 	IssuerRawCA   []byte `json:"issuerCA"`
 	IssuerCAPath  string `json:"issuerCAPath"`
-	issuerCA      *x509.Certificate
+	issuerCAs     []*x509.Certificate
 	IssuerURL     string `json:"issuerURL"`
 	RedirectURL   string `json:"redirectURL"`
 	UsernameClaim string `json:"usernameClaim"`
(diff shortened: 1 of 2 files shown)


philipgough (dismissed):
/lgtm

@squat any concerns?

JoaoBraveCoding:
Is there anything else I can help with to get this one merged?

squat (commented):
I'm find with this functionality. I just have one style comment to improve clarity and prevent future mistakes with validation.

JoaoBraveCoding:
@squat thanks for the reply! However, I don't think your comment went through

authentication/oidc.go
```diff
@@ -96,17 +96,28 @@ func newOIDCAuthenticator(c map[string]interface{}, tenant string,
...
-		}
+		var (
+			block *pem.Block
+			rest  = config.IssuerRawCA
+		)
+
+		for {
+			block, rest = pem.Decode(rest)
+			if block == nil {
+				if len(config.issuerCAs) == 0 {
```
squat: For style and clarity, let's instead break on line 107, ie, whenever `block == nil` since that's the exit condition and let's move this check outside the block, since that's the validation. I think that would be clearer IMO.

squat:
@JoaoBraveCoding oops. I just reposted!

squat (dismissed):
Lgtm!

squat:
Please remove that last commit. We can make it a separate PR and merge before this of necessary but it's unrelated

JoaoBraveCoding:
@squat thank you for that, you are right:
- Changes to e2e to reduce flakiness that came with re-using test names: https://github.com/observatorium/api/pull/894
- Commit that I removed that reduces flakiness in unit tests: https://github.com/observatorium/api/pull/895

philipgough (approved):

PR #882, opened 2026-02-27
Add multi-variant expressions to LogQL parser

btaani:
Add support for LogQL features included in Loki between v3.1.0 and v3.6.5

```
$ git log --oneline v3.1.0..v3.6.5 -- pkg/logql/syntax/expr.y pkg/logql/syntax/syntax.y
417d0a5252 chore: add multi-variant query support to LogQL (#16196)
1187a73667 chore: LogQL syntax and AST simplification and cosistency (#16028)
7b53f20f70 feat: Introduce shardable probabilistic topk for instant queries. (#14243)
```

- https://github.com/grafana/loki/pull/14765

- https://github.com/grafana/loki/pull/16196

This PR only adds support for `variants` queries. `approx_topk` is not supported for setups without the `query-scheduler`

Ref: [LOG-8712](https://issues.redhat.com/browse/LOG-8712)


btaani:
/hold
fixing & adding more tests

JoaoBraveCoding (commented):
`approx_topk` seems to still be in experimental feature according to the docs. It works when sent directly to the querier, but fails with `parse error at line 1, col 1: syntax error: unexpected IDENTIFIER` when sent through query-frontend.

I verified this by calling both components directly over mTLS with the same query: the querier returned 200 and valid results, while the query-frontend returned 400. This points to a query-frontend bug... likely related to approx_topk being rewritten into internal non-LogQL expressions during frontend processing and then reparsed on a code path that expects normal LogQL.

/cc @xperimental with this shall we add support for `approx_topk`? I'm on the fence because I'm not sure if grafana will eventually support it or drop it... at the same time if we don't add support we would need to remember that

multi-variant worked perfectly 👌

JoaoBraveCoding:
As discussed during the daily we will move ahead with backporting both `approx_topk` and `multi-variant` to continue allow us to use git to track which aspects if grafana/loki LogQL need to be ported to obs/api LogQL

@btaani feel free to correct the tests so we can approve and merge this one

JoaoBraveCoding (approved):
/lgtm

JoaoBraveCoding (approved):
/lgtm

JoaoBraveCoding:
We can close this one as `variants` were deprecated today https://github.com/grafana/loki/pull/23823

PR #888, opened 2026-03-25
Support BYO/external authentication in OpenShift authenticator

btaani:
When OpenShift is configured with external OIDC providers (Keycloak, Red Hat SSO, etc.), the OAuth discovery endpoint returns 404 since OAuth server gets disabled. The current code retries indefinitely causing the LokiStack gateway to hang during startup.
This PR:
- adds graceful handling for OpenShift clusters using BYO (Bring Your Own) external authentication where built-in OAuth server is disabled.
- Detects 404 responses from OAuth endpoint and disables OAuth flow.
- Generates a warning message
- Bearer token authentication continues to work.
- Transient errors still retry as before.

### References
- Setting up Keycloak through the operator: https://docs.redhat.com/en/documentation/red_hat_build_of_keycloak/24.0/html-single/operator_guide/index
- Enabling direct authentication with an external OIDC identity provider: https://docs.redhat.com/en/documentation/openshift_container_platform/4.21/html-single/authentication_and_authorization/index#external-auth

### DIFF
--- a/authentication/openshift.go
+++ b/authentication/openshift.go
@@ -145,19 +145,20 @@ func newOpenshiftAuthenticator(c map[string]interface{}, tenant string,
 	var clientID string
 	var clientSecret string
 
-	for b.Reset(); b.Ongoing(); {
-		clientID, clientSecret, err = openshift.DiscoverCredentials(config.ServiceAccount)
-		if err != nil {
-			level.Error(logger).Log(
-				"tenant", tenant,
-				"msg", errors.Wrap(err, "unable to read serviceaccount credentials"))
-			registrationRetryCount.WithLabelValues(tenant, OpenShiftAuthenticatorType).Inc()
-			b.Wait()
+	if oauthEnabled {
+		for b.Reset(); b.Ongoing(); {
+			clientID, clientSecret, err = openshift.DiscoverCredentials(config.ServiceAccount)
+			if err != nil {
+				level.Error(logger).Log(
+					"tenant", tenant,
+					"msg", errors.Wrap(err, "unable to read serviceaccount credentials"))
+				registrationRetryCount.WithLabelValues(tenant, OpenShiftAuthenticatorType).Inc()
+				b.Wait()
 
-			continue
+				continue
+			}
+			break
 		}
-
-		break
 	}
 
 	authOpts := openshift.DelegatingAuthenticationOptions{
@@ -207,6 +208,8 @@ func newOpenshiftAuthenticator(c map[string]interface{}, tenant string,
 		oauthEnabled:  oauthEnabled,
 	}
 
+	r := chi.NewRouter()
+	r.Use(tracing.WithChiRoutePattern)
 	if oauthEnabled {
 		osAuthenticator.oauth2Config = oauth2.Config{
 			ClientID:     clientID,
@@ -223,15 +226,12 @@ func newOpenshiftAuthenticator(c map[string]interface{}, tenant string,
 			},
 			RedirectURL: config.RedirectURL,
 		}
-		r := chi.NewRouter()
-		r.Use(tracing.WithChiRoutePattern)
+
 		r.Handle(loginRoute, osAuthenticator.openshiftLoginHandler())
 		r.Handle(callbackRoute, osAuthenticator.openshiftCallbackHandler())
-		osAuthenticator.handler = r
-
-	} else {
-		osAuthenticator.handler = chi.NewRouter()
 	}
+
+	osAuthenticator.handler = r
 
 	return osAuthenticator, nil
 }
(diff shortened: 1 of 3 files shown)


authentication/openshift.go
```diff
@@ -541,3 +537,30 @@ func (a OpenShiftAuthenticator) GRPCMiddleware() grpc.StreamServerInterceptor {
...
+					level.Error(logger).Log(
+						"tenant", tenant,
+						"msg", errors.Wrap(err, "unable to auto discover OpenShift OAuth endpoints"))
+					registrationRetryCount.WithLabelValues(tenant, OpenShiftAuthenticatorType).Inc()
+					b.Wait()
+					continue
+				}
+				break
+			}
+		}
```
JoaoBraveCoding: Two things:
1. We don't need to check for the disabled oauth outside the for loop
2. Instead of relying in string comparison we can just use type comparisons

```suggestion
	var authURL, tokenURL *url.URL
	var err error
	for b.Reset(); b.Ongoing(); {
		authURL, tokenURL, err = openshift.DiscoverOAuth(client)
		if err != nil {
			if errors.Is(err, openshift.ErrOAuthServerNotFound) {
				return nil, nil, false
			}
			level.Error(logger).Log(
				"tenant", tenant,
				"msg", errors.Wrap(err, "unable to auto discover OpenShift OAuth endpoints"))
			registrationRetryCount.WithLabelValues(tenant, OpenShiftAuthenticatorType).Inc()
			b.Wait()
			continue
		}
		break
```

authentication/openshift.go
```diff
@@ -139,27 +140,9 @@ func newOpenshiftAuthenticator(c map[string]interface{}, tenant string,
...
-				"tenant", tenant,
-				"msg", errors.Wrap(err, "unable to auto discover OpenShift OAuth endpoints"))
-			registrationRetryCount.WithLabelValues(tenant, OpenShiftAuthenticatorType).Inc()
-			b.Wait()
-
-			continue
-		}
-
-		break
-	}
```
JoaoBraveCoding, line 159: If oauthEnabled is disabled we should also no try to discover the credentials (for loop after)

authentication/openshift.go
```diff
@@ -235,15 +222,17 @@ func newOpenshiftAuthenticator(c map[string]interface{}, tenant string,
...
-		},
+		}
+		r := chi.NewRouter()
+		r.Use(tracing.WithChiRoutePattern)
+		r.Handle(loginRoute, osAuthenticator.openshiftLoginHandler())
+		r.Handle(callbackRoute, osAuthenticator.openshiftCallbackHandler())
+		osAuthenticator.handler = r
+
+	} else {
+		osAuthenticator.handler = chi.NewRouter()
```
JoaoBraveCoding: We should not drop the tracing `WithChiRoutePattern` if oauth is disabled

authentication/openshift_test.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	// Split host and port for KUBERNETES env vars
+	host, port, err := net.SplitHostPort(mockURL.Host)
+	if err != nil {
+		t.Fatalf("failed to parse mock server address: %v", err)
+	}
+
+	t.Setenv("KUBERNETES_SERVICE_HOST", host)
+	t.Setenv("KUBERNETES_SERVICE_PORT", port)
```
JoaoBraveCoding: `toKubeAPIURLWithPath` never consumes `KUBERNETES_SERVICE_PORT` so this can be dropped

JoaoBraveCoding (commented):
Overall looks good, however we should also open a Jira issue (not a bug) to eventually support external OpenShift authenticators in the operator. The operator should see how the cluster is configured and configure obs/api correctly to continue supporting the Oauth flow, otherwise we will break customers that use Loki with Grafana if they have an external oauth configured. 

cc @xperimental

authentication/openshift.go
```diff
@@ -221,7 +205,13 @@ func newOpenshiftAuthenticator(c map[string]interface{}, tenant string,
...
 		config:        config,
 		cookieName:    fmt.Sprintf("observatorium_%s", tenant),
-		oauth2Config: oauth2.Config{
+		oauthEnabled:  oauthEnabled,
+	}
+
+	r := chi.NewRouter()
+	r.Use(tracing.WithChiRoutePattern)
+	if oauthEnabled {
+		osAuthenticator.oauth2Config = oauth2.Config{
```
JoaoBraveCoding, line 214: Suggestion to move all the code regarding oauth together
```suggestion
	// Discover OAuth endpoints and credentials if OAuth is enabled
	if authURL, tokenURL, oauthEnabled := discoverOAuthEndpoints(client, logger, tenant, registrationRetryCount, b); oauthEnabled {
		osAuthenticator.oauthEnabled = true

		var clientID string
		var clientSecret string

		for b.Reset(); b.Ongoing(); {
			clientID, clientSecret, err = openshift.DiscoverCredentials(config.ServiceAccount)
			if err != nil {
				level.Error(logger).Log(
					"tenant", tenant,
					"msg", errors.Wrap(err, "unable to read serviceaccount credentials"))
				registrationRetryCount.WithLabelValues(tenant, OpenShiftAuthenticatorType).Inc()
				b.Wait()

				continue
			}
			break
		}

```

JoaoBraveCoding (commented):
Just a small nit but overall lgtm

JoaoBraveCoding (approved):
LGTM

philipgough (approved):
/lgtm

PR #893, opened 2026-03-30
Avoid empty label values for HTTP metrics

simonpasquier:
This commit prevents the initialization of HTTP metrics with empty values for the `code`, `method` and `tenant` labels. These metrics were set to zero but never incremented.

### DIFF
--- a/server/http_instrumentation.go
+++ b/server/http_instrumentation.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"context"
+	"maps"
 	"net/http"
 	"strconv"
 	"time"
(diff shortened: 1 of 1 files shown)


simonpasquier:
cc @saswatamcode @philipgough

philipgough:
@simonpasquier just to add context, these change was added in https://github.com/observatorium/api/pull/547 and IIUC it was to provide a solution to SLO alerts that were being generated by https://github.com/pyrra-dev/pyrra in low traffic environments.

Just to clarify, I am not against the change in principle, just wanted to shine some light on it :)

PR #921, opened 2026-08-13
Avoid full rebuild when building binary

xperimental:
I don't think we need this precaution and it just increases wait time when doing a build.


philipgough (approved):
/lgtm

PR #923, opened 2026-08-26
feat: Introduce curve support to tls config

JoaoBraveCoding:
Issue https://redhat.atlassian.net/browse/LOG-8797


philipgough (approved):
/lgtm

PR #927, opened 2026-09-16
tls: Add client CA configuration for mTLS verification

saswatamcode:
Add --tls.client-ca-file flag and configure tls.Config.ClientCAs to enable proper verification of client certificates. Without this, the gateway cannot verify client certs against a specific CA.

Note, this is for mTLS between sender and Observatorium API


saswatamcode:
@philipgough pointed out that each tenant passes it through already https://github.com/observatorium/api/blob/main/main.go#L259-L261 via their own config, so this is probably not required in such a setup

PR #933, opened 2026-09-22
Fix e2e tests by using mirrored minio image

xperimental:


moadz (approved):
blimey that's an old image

PR #935, opened 2026-09-23
traces/rbac: use gogo/protobuf to match Tempo marshalling lib

andreasgerstmayr:
The response RBAC modifier unmarshals a Tempo API response, filters it, and re-marshals it. To stay compliant with Tempo's API, that round-trip must produce output identical to what Tempo sends, which means it has to (un)marshal with the same library Tempo uses: gogo/protobuf.

This code used golang/protobuf's jsonpb instead, which disagrees with gogo on empty repeated fields (gogo emits "[]", golang/protobuf omits them). A zero-result /api/search therefore came back without the "traces" field instead of Tempo's {"traces":[],...}.

Switch to gogo's jsonpb/proto so the modifier's output matches Tempo marshalling and "traces" field is always present.


andreasgerstmayr:
Oh, this is basically the same as #934 (I worked on it yesterday but forgot to push). Main difference is that it uses `gogo` for both `protobuf/jsonpb` and `protobuf/proto`, to mirror Tempo.

